PR #11, opened 2024-02-28
feat: allow inviting to scope via user id

crowlKats:


api/src/api/types.rs
```diff
@@ -285,8 +285,9 @@ impl From<(ScopeMember, UserPublic)> for ApiScopeMember {
 
 #[derive(Debug, Deserialize)]
 #[serde(rename_all = "camelCase")]
-pub struct ApiAddScopeMemberRequest {
-  pub github_login: String,
+pub enum ApiAddScopeMemberRequest {
+  GithubLogin(String),
+  Uuid(Uuid),
```
lucacasonato, line 290:
```suggestion
  Id(Uuid),
```

We don't call this the "user uuid" but the "user id"

PR #56, opened 2024-02-29
fix: npm package missing non-script files

marvinhagemeister:
This PR addresses the issue that npm tarballs only contain the files discovered by `deno_graph`, instead of all files that were part of the jsr package tarball. So we'll keep track of which files we already processed from the tarball and insert the remaining ones.


lucacasonato:
Closing in favor of #236

PR #103, opened 2024-03-01
Fix website horizontal scroll

MattheousDT:
Website has a horizontal scrollbar on operating systems (Windows etc.) that have permanent scrollbars.

### DIFF
--- a/frontend/static/styles.css
+++ b/frontend/static/styles.css
@@ -61,6 +61,7 @@
     font-size: 1.125rem;
     font-weight: 400;
     color: theme("colors.jsr-gray.700");
+    overflow-x: hidden;
     @apply opsize-normal;
   }
 


lucacasonato:
@MattheousDT Where are you seeing this? Can you post a screenshot?

MattheousDT:
Firefox
image

Chromium
image

Note the horizontal scrollbar at the bottom. Happens regardless of window size

lucacasonato:
Thanks. This seems to be happening because of the `width: 100vw` on the canvas for particle.js.

@josh-collinsworth PTAL

josh-collinsworth:
> Thanks. This seems to be happening because of the `width: 100vw` on the canvas for particle.js.
> 
> @josh-collinsworth PTAL

Will do. This looks good at a glance but I want to take some time to thoroughly test to make sure there aren't other side effects.

josh-collinsworth:
I have what I think is a more preferable implementation here: https://github.com/jsr-io/jsr/pull/134

It's the same fix/CSS, but just applied _only_ to the homepage, just so we avoid any unforeseen side effects, or situations where we might _want_ horizontal overflow.

Thanks for reporting and opening this!

PR #104, opened 2024-03-01
fix: 100% score is slightly bigger than others

ryuapp:
If the score is 100%, the size is slightly bigger than other scores.
The layout was a little broken, so I fixed it.


donjo (approved):
thank you!

PR #107, opened 2024-03-01
Make Logo pixel perfect

mxdvl:
# What does this change?

Make the `jsr` logo pixel-perfect.

Use a much smaller canvas size, where each unit result in one square and ensure the view box has a 2:1 ratio to ensure perfect scaling in relation to font-size and UI elements

See [interactive demo](https://svelte.dev/repl/60728637967647bbb589913d6312d2ad?version=4.2.12) for how this was generated

| Before | After |
|--------|--------|
| image | image |


frontend/components/Logo.tsx
```diff
@@ -3,21 +3,20 @@ export function Logo(props: { class?: string }) {
...
-          fill="#F7DF1E"
-          d="M100.101 196h47.171V49h49v196H51.102v-98H100.1v49ZM588.272 98v98h-49v-49h-49v147h-49V98h147ZM294.272 98v49h98v147h-147v-49h98v-49h-98V49h147v49h-98Z"
-        />
-      </g>
+      <path
+        d="M0,2h2v-2h7v1h4v4h-2v2h-7v-1h-4"
+        fill="#083344"
+      />
+      <path
+        d="M1.5,3.5v1h2v-3m4,0h-2v2h2v2h-2m4,0v-3h2v1"
```
mxdvl, line 15: A 1px stroke must be aligned to a half extra pixel to ensure that it lines up perfectly.

demonstration of pixel perfect stroke

frontend/components/Logo.tsx
```diff
@@ -3,21 +3,20 @@ export function Logo(props: { class?: string }) {
   return (
     <svg
       xmlns="http://www.w3.org/2000/svg"
-      viewBox="0 0 638 343"
-      fill="none"
+      viewBox="-1.5 -0.5 16 8"
```
mxdvl, line 6: Nudge the `viewBox` as the artwork fits in 13×7, but a 2:1 aspect ratio that lines up with `rem` is more flexible

mxdvl (commented):
Some pointers on changes…

The path was written by hand using simple horizontal and vertical movements based on the original asset.

josh-collinsworth (changes requested):
@mxdvl The original version of the animated SVG included two "squares" worth of white space all around the logo inside the viewbox, in order to allow space for the squares to be up to 200% out of position before sliding into place with the animation. (You'll notice with this version, some of the squares initially render out of bounds of the SVG and are invisible until they slide into the viewbox.)

Would it be possible to refactor this PR to add that extra whitespace, so that all squares will be visible on the initial render, pre-animation?

mxdvl:
> @mxdvl The original version of the animated SVG included two "squares" worth of white space all around the logo inside the viewbox, in order to allow space for the squares to be up to 200% out of position before sliding into place with the animation. (You'll notice with this version, some of the squares initially render out of bounds of the SVG and are invisible until they slide into the viewbox.)

Are you talking about this specific PR? I'm only modifying the static logo here and not touching the animated one. I think you might be [referring to a follow-up branch I've got in my fork](https://github.com/mxdvl/jsr/compare/mxdvl/crisp-logo...mxdvl/animated-logo), that's not ready yet for PR.

I do have a question about static SVG assets… are they used anywhere? I could not find any reference in the code.

josh-collinsworth:
> Are you talking about this specific PR? I'm only modifying the static logo here and not touching the animated one. I think you might be [referring to a follow-up branch I've got in my fork](https://github.com/mxdvl/jsr/compare/mxdvl/crisp-logo...mxdvl/animated-logo), that's not ready yet for PR.
> 
> I do have a question about static SVG assets… are they used anywhere? I could not find any reference in the code.

Yeah, I must have seen the follow-up. The homepage animated logo was definitely affected in the branch I looked at.

The `logo-square.svg` file is used in `_app.tsx`. The main `logo.svg` file is used on the homepage and package score page, in addition to a couple other places as images.

frontend/components/Logo.tsx
josh-collinsworth: I have concerns about the viewbox extending beyond the artwork bounds, just because it will mean any time the SVG is used anywhere, there will be negative margins or positioning necessary to align it. Is a way to accomplish this without adding the space?

mxdvl:
> The `logo-square.svg` file is used in `_app.tsx`. The main `logo.svg` file is used on the homepage and package score page, in addition to a couple other places as images.

Would you like me to update these assets as part of this PR?

frontend/components/Logo.tsx
mxdvl: Yes, but for the height, there are not many multiple of 7 within the Tailwind dimensions, hence why 8 units high was chosen. One possibility would be to keep the height at 8 with one empty unit at the bottom rather than 0.5 at the top and bottom.

For the width, it’s entirely possible to keep it at 13 units if you prefer, and I’ll make the change.

mxdvl: I've now made the change. How do you suggest we adjust the height? Adding the `h-8` class will not work as that would result in 32px high, which is not divisible by 7.

Here are some options I thought of:

- add a `height` prop with some JSDoc suggesting to use a multiple of 7
- add a `size` prop that takes integers that are multiplied by 7 to produce de final height in pixels
- add a `size` prop that takes a series of values like S M L XL that uses CSS calc to get a valid size in `rem`: `calc(2 * 0.875rem)` for `h-8` equivalent
- do not allow size variations for this component

frontend/components/Logo.tsx
```diff
@@ -3,21 +3,20 @@ export function Logo(props: { class?: string }) {
   return (
     <svg
       xmlns="http://www.w3.org/2000/svg"
-      viewBox="0 0 638 343"
-      fill="none"
+      viewBox="0 13 7"
```
josh-collinsworth, line 6:
```suggestion
      viewBox="0 0 13 7"
```

frontend/components/Logo.tsx
josh-collinsworth: I like the idea of taking a size prop, and making sure all the sizes match up to exact pixel alignments.

mxdvl:
@josh-collinsworth does this still require changes?

josh-collinsworth:
@mxdvl Sorry for letting this sit for so long.

My main hesitation here is in making the logo use a stroke to render, and relying on the stroke as a part of the design. Maybe I'm being overly cautious, but I've come across some edge cases where strokes can vary when SVGs are ported between platforms or softwares (the thickness, the cap, etc.)

Again, maybe that's excessive caution. But could we not achieve this desired outcome with the original SVG (or, at least, an SVG that is fully expanded and which doesn't use strokes?)

Here's an example of what I mean; it's essentially applying this PR's ideas to the original SVG. https://github.com/jsr-io/jsr/pull/402

mxdvl:
> I just like to keep SVGs as "expanded" as possible, to borrow the Illustrator terminology.

I think that's probably overly cautious but definitely not my call to make in this case. I'll expand the SVG while keeping the smaller viewBox where each unit maps to one “block” for slightly reduced filesize.

josh-collinsworth (approved):

PR #115, opened 2024-03-01
Minor UI improvements

roj1512:
- Made more areas of the homepage hero interactive (e.g., the header, the motto).
- Disabled selectablity in more areas that don't make sense.

### DIFF
--- a/frontend/islands/HomepageHero.tsx
+++ b/frontend/islands/HomepageHero.tsx
@@ -186,7 +186,10 @@ export function HomepageHero(
           />
         </div>
         <div class="flex flex-col items-center gap-4">
-          <a class="button-primary relative z-10 pointer-events-auto" href="/new">
+          <a
+            class="button-primary relative z-10 pointer-events-auto"
+            href="/new"
+          >
             <Plus /> Publish a package
           </a>
         </div>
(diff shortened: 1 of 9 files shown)


roj1512:
The PR is ready, but I would like to add some more improvements. Where is the source code for this located? I mean the installation commands' UI.

image

roj1512:
Seems like that is in deno_doc. So marking this PR as ready for review.

frontend/islands/HomepageHero.tsx
```diff
@@ -152,7 +152,7 @@ export function HomepageHero(
         <script src={asset("/scripts/particles.js")} defer></script>
       </Head>
       <style>{canvasStyle}</style>
-      <div class="section-x-inset-xl flex flex-col items-center justify-center gap-12 relative">
+      <div class="section-x-inset-xl flex flex-col items-center justify-center gap-12 relative pointer-events-none">
```
josh-collinsworth, line 155: I'm not so sure about this one. I understand disabling select for interactive controls (and it is nice to have the background interactivity here), but this is just a heading, and a user might legitimately want to select the text. I'd prefer to err on the side of trusting the user here.

josh-collinsworth (changes requested):
Just one location where I feel these changes might be a little bit of an overreach. Otherwise, a solid addition.

frontend/islands/HomepageHero.tsx
roj1512: This `pointer-events-none` is not for disabling selectability. Users can still select the text in the header. This lets the user interact with the animation (or whatever it is called) from empty parts of the header.

josh-collinsworth: Ah yes I see that; my mistake.

josh-collinsworth (approved):

PR #116, opened 2024-03-01
fix: case typo

altfoxie:


marvinhagemeister (approved):
Good catch, thanks for making a PR!

PR #79, opened 2024-02-29
[WIP] chore: Remove waitlist from frontend

bartlomieju:
## DO NOT MERGE

PR #123, opened 2024-03-02
chore: remove waitlist UI, emails, and admin APIs

ry:


bartlomieju (commented):
I did most of this work in https://github.com/jsr-io/jsr/pull/79 but we can land your PR. Can you remove handlebars templates for waitlist accept emails too?

lucacasonato (approved):
LGTM!

PR #134, opened 2024-03-02
Fix horizontal scroll on homepage

josh-collinsworth:
As reported in https://github.com/jsr-io/jsr/pull/103

This is the same fix, but applied in a more targeted way to avoid any possible side effects.


lucacasonato (approved):

lucacasonato:
Thanks for the report and suggested fix @MattheousDT!

PR #148, opened 2024-03-03
Fix spinning logo in header. (Good idea? ¯\_(ツ)_/¯)

josh-collinsworth:
I noticed this little spinny effect was broken so I reintroduced it in a new way. I don't now if it's a good idea or a good implementation of the idea, but it's kinda fun. ¯\\\_(ツ)_/¯

### DIFF
--- a/frontend/components/Header.tsx
+++ b/frontend/components/Header.tsx
@@ -3,9 +3,9 @@
 import { FullUser } from "../utils/api_types.ts";
 import { GlobalSearch } from "../islands/GlobalSearch.tsx";
 import { UserMenu } from "../islands/UserMenu.tsx";
-import { Logo } from "./Logo.tsx";
 import { GitHub } from "./icons/GitHub.tsx";
 import { SearchKind } from "../util.ts";
+import { HeaderLogo } from "../islands/HeaderLogo.tsx";
 
 export function Header({
   user,
@@ -61,6 +61,17 @@ export function Header({
               <Logo size="medium" class="flex-none hover:animate-flip-rotate" />
             </a>
           )}
+          <div>
+            {!isHomepage && (
+              <a
+                href="/"
+                class="outline-none focus-visible:ring-2 ring-cyan-700"
+              >
+                <span className="sr-only">JSR home</span>
+                <HeaderLogo class="h-8 flex-none" />
+              </a>
+            )}
+          </div>
           <div class="hidden sm:block grow-1 flex-1">
             {!isHomepage && (
               <GlobalSearch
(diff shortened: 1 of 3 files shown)


mxdvl (approved):
I love it, but I think that making it CSS only would be superior.

```css
a[href="/"] {
  transform: rotate(0deg);
  transition: transform 360ms;
}
a[href="/"]:hover {
  transform: rotate(180deg);
}
```

lucacasonato:
I agree - can we do it with CSS? :)

josh-collinsworth:
I had a reason I wasn't doing it in CSS before; there was something janky about it. But let me see if it can be reworked.

josh-collinsworth:
Ok, I'm remembering now: if you do it purely in CSS (whether via animation or transition), there's a bit of jank if the user's cursor leaves before the animation is finished.

But comparing the two, it seems like that's just a possibility either way, so this is probably the superior option.

lucacasonato (approved):
Love it

mxdvl:
> if you do it purely in CSS (whether via animation or transition), there's a bit of jank if the user's cursor leaves before the animation is finished.

Transition are free from these artefacts, provided that you do not conditionally apply the `transition` property on hover. I've [updated the demo to illustrate the behaviour](https://svelte.dev/repl/60728637967647bbb589913d6312d2ad?version=4.2.12).

josh-collinsworth:
> > if you do it purely in CSS (whether via animation or transition), there's a bit of jank if the user's cursor leaves before the animation is finished.
> 
> Transition should free from such artefacts, provided that you do not conditionally apply the `transition` property on hover. I've [updated the demo to illustrate the behaviour](https://svelte.dev/repl/60728637967647bbb589913d6312d2ad?version=4.2.12).

The issue is: I think we want the logo to _stay_ in its rotated position at the end.

@lucacasonato do you have any opinions on the better implementation?

a) Logo rotates 180 degrees and remains in rotated state until next hover, but could jerk if cursor leaves during animation;
b) Logo rotates 180 degrees on hover, and un-rotates when hover ends. No jerk, but always un-rotates.

mxdvl:
> > > if you do it purely in CSS (whether via animation or transition), there's a bit of jank if the user's cursor leaves before the animation is finished.
> 
> > Transition should free from such artefacts, provided that you do not conditionally apply the `transition` property on hover. I've [updated the demo to illustrate the behaviour](https://svelte.dev/repl/60728637967647bbb589913d6312d2ad?version=4.2.12).
> 
> The issue is: I think we want the logo to _stay_ in its rotated position at the end.

I see! If my input is welcome, I think the best way to achieve this effect would be to set the desired rotation as a CSS Custom Property that is incremented by `180deg` every time a cursor enters with a transition: no jank, no rotation back, minimal JS state handling. Feel free to disregard this suggestion 😊.

Crafting little delights is hard, but oh so rewarding when you get there!

josh-collinsworth:
Ok I'm back and @mxdvl I think you're right; a sprinkling of JS with signals creates the ideal effect here. @lucacasonato you agree?

PR #161, opened 2024-03-04
chore: reserve apache scope for the ASF

tisonkun:
I'm one of The ASF Members. Perhaps I'd start a discussion in the foundation to register a foundation-wise scope and publish Apache projects to JSR.

This file seems to reserve some well-known names to avoid branding issues. I suggest to add "apache" on the list.

### DIFF
--- a/api/src/reserved_scopes.json
+++ b/api/src/reserved_scopes.json
@@ -18,6 +18,7 @@
   "antd",
   "anymatch",
   "anypromise",
+  "apache",
   "apollo",
   "apple",
   "appwrite",


lucacasonato (approved):

tisonkun:
@lucacasonato Thanks for reviewing this patch. May I ask what I should do if the foundation reaches a consensus on registering the scope on behalf of it?

lucacasonato:
Then please send an email to help@jsr.io from a foundation affiliated email address - we can then proceed with registering the scope.

PR #163, opened 2024-03-04
feat: use provenance in JSR score

littledivy:
Closes https://github.com/jsr-io/jsr/issues/84


lucacasonato (commented):
@littledivy Do you want this to be an "additional point" or update the maximum total by one?

littledivy:
I think we update max total by one. 

oh i see, i have to bump the `MAX_SCORE` for that

lucacasonato:
You also have to bump it on the JS side

littledivy:
Can you point me to where that is?

frontend/routes/package/score.tsx
```diff
@@ -164,6 +164,14 @@ export default function Score(
             This package is compatible with more than one runtime, and is marked
             as such in the package settings.
           </ScoreItem>
+          <ScoreItem
+            value={data.score.hasProvenance}
+            scoreValue={1}
+            title="Has provenance"
+          >
+            This package is published from a verifiable CI/CD workflow, and has
+            a public transparency log.
```
lucacasonato, line 173: Add links to docs

lucacasonato (approved):
Oh my bad - that does not exist anymore. Then, LGTM modulo the docs nit!

frontend/routes/package/score.tsx
lucacasonato: Added a link, and updated the max down to 17. Let's see if this is widely adoptable, and once it is, we can tweak the max score :)

Issue #139, opened 2024-03-02
Something rewrote my import to `from 'npm:lit@^2.2.7';` and now I can't build with `esbuild`


simonw:
My original source code here: https://github.com/simonw/datasette-table/blob/0.1.0/datasette-table.js

```javascript
import {LitElement, html, css} from 'lit';
```
After I published it to `jsr.io` it was rewritten like this: https://jsr.io/@datasette/table/0.1.0/datasette-table.js

```javascript
import {LitElement, html, css} from 'npm:lit@^2.2.7';
```
But now when I try to install it and run it through `esbuild` I get an error:
```bash
mkdir /tmp/datasette-demo
cd /tmp/datasette-demo
npx jsr add @datasette/table
echo 'import * as mod from "jsr:@datasette/table";' > index.js
npx esbuild index.js --bundle --outfile=bundle.js
```
Error:
```
✘ [ERROR] Could not resolve "npm:lit@^2.2.7"

    node_modules/@datasette/table/datasette-table.js:1:36:
      1 │ import {LitElement, html, css} from 'npm:lit@^2.2.7';
        ╵                                     ~~~~~~~~~~~~~~~~

  You can mark the path "npm:lit@^2.2.7" as external to exclude it from the bundle, which will
  remove this error and leave the unresolved path in the bundle.

1 error
```
I asked in Discord and `marvinh` said: https://discord.com/channels/684898665143206084/1203185670508515399/1213595537505521664

> Esbuild doesn't understand `npm:` or `jsr:` . That's a Deno only thing. So your source code should not have import specifiers with these prefixes anywhere. That said the lit error looks like a bug. That should've been transpiled away

More of my notes here: https://til.simonwillison.net/javascript/jsr-esbuild

simonw:
Bill Mill found a workaround using [esbuild_deno_loader](https://github.com/lucacasonato/esbuild_deno_loader):

https://notes.billmill.org/programming/javascript/build_tools/using_esbuild_to_package_a_deno_package_for_the_browser.html

dsherret:
The `jsr` tool should not be pulling in npm specifiers. That's a bug. The specifier at https://jsr.io/@datasette/table/0.1.0/datasette-table.js is correct, but the specifier in node_modules/@datasette/table/datasette-table.js is wrong. It should be "lit"

lucacasonato:
The fix is in https://github.com/jsr-io/jsr/pull/172

simonw:
[...]

PR #172, opened 2024-03-04
fix: rewrite js import sources

marvinhagemeister:
We previously only transpiled ts sources, but we need to transpile js sources as well to be able to rewrite imports like turning `import ... from "npm:lit@^2.2.7"` into `import ... from "lit";`.

Fixes https://github.com/jsr-io/jsr/issues/139

### DIFF
--- a/api/src/analysis.rs
+++ b/api/src/analysis.rs
@@ -759,7 +759,7 @@ impl<'a> deno_graph::source::Loader for GcsLoader<'a> {
 }
 
 #[derive(Default)]
-struct ModuleAnalyzer {
+pub struct ModuleAnalyzer {
   pub analyzer: CapturingModuleAnalyzer,
   pub module_info: RefCell<HashMap<Url, ModuleInfo>>,
 }
--- a/api/src/npm/tarball.rs
+++ b/api/src/npm/tarball.rs
@@ -435,15 +435,6 @@ mod tests {
     Ok(transpiled_files)
   }
 
-  fn get_content_without_source_map(data: &[u8]) -> String {
-    let s = String::from_utf8_lossy(data);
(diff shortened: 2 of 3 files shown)


api/src/npm/tarball.rs
```diff
@@ -101,10 +101,23 @@ pub fn create_npm_tarball<'a>(
...
+        deno_ast::MediaType::JavaScript | deno_ast::MediaType::Mjs => {
+          let source = sources
+            .get_parsed_source(module.specifier())
+            .expect("parsed source should be here");
+          let source_url = Url::options()
+            .base_url(Some(registry_url))
+            .parse(&format!("./@{scope}/{package}/{version}{path}",))
+            .unwrap();
+          let transpiled = transpile_to_js(source, source_url)
+            .with_context(|| format!("failed to transpile {}", path))?;
```
lucacasonato, line 115: This is kind of annoying. Means source maps also become necessary for JS.

I'll do a follow up on this to make this use text changes, like we do in CLI, rather than going via the AST. That way we can avoid having to re-emit the JS.

api/src/analysis.rs
```diff
@@ -759,7 +759,7 @@ impl<'a> deno_graph::source::Loader for GcsLoader<'a> {
 }
 
 #[derive(Default)]
-struct ModuleAnalyzer {
+pub struct ModuleAnalyzer {
```
lucacasonato, line 762: Not needed?

lucacasonato (approved):
Great!

Please don't land without bumping revision

api/src/analysis.rs
marvinhagemeister: It's needed for the tests https://github.com/jsr-io/jsr/blob/js-imports/api/src/npm/tarball.rs#L382

PR #192, opened 2024-03-05
fix: respect img height in docs

### LINKED ISSUE #120: Image not being rendered correctly in readme
Someone uses pics in readme to add a bit style. Like:
image
Github handle html tag with size adjustment:
```html
<img height="18" src="https://raw.githubusercontent.com/PKief/vscode-material-icon-theme/main/icons/deno.svg">
```
But JSR front end shows this pic streached to fullsize:
image

Is it ok?

unbyte:
Tailwindcss applies `height: auto` for img and video tags by default in its preflight.css, so imgs and videos with a height attribute, the most common practice to resize media in markdown, are broken because height property in CSS has a higher priority than the attribute on the DOM.

This PR reimplements the preflight core plugin for tailwindcss and modifies the preflight.css to address this issue.

fixes #120

### DIFF
--- a/frontend/deno.json
+++ b/frontend/deno.json
@@ -25,7 +25,8 @@
     "@orama/orama": "npm:@orama/orama@^2",
     "@oramacloud/client": "npm:@oramacloud/client@^1",
     "tailwindcss": "npm:tailwindcss@3.4",
-    "tailwindcss/": "npm:/tailwindcss@3.4/"
+    "tailwindcss/": "npm:/tailwindcss@3.4/",
+    "postcss": "npm:postcss@8.4"
   },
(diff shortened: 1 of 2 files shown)


lucacasonato (commented):
Would a simpler option be to add `.markdown img { height: initial }` to the `gfm.css` file?

unbyte:
> Would a simpler option be to add `.markdown img { height: initial }` to the `gfm.css` file?

Sadly no, because `height: initial` equals to `height: auto`.
Also, once height property has been set (to any value) in CSS, it can no longer behave as if it had not been set.

lucacasonato (approved):
Ok, thanks for the explanation. I have added another selector to target `img` that are not inside of a `.markdown`, so they still get reset to `height: auto`.

LGTM!

PR #196, opened 2024-03-05
chore(frontend): remove search placeholder flash

sno2:
Going to a package's page and pressing refresh caused the search bar's placeholder to show the mac symbol then flash to the windows version. This fixes that problem.


lucacasonato:
Thanks for the report - I fixed this in a different way in c27b992fba4a27334144b51488d488a9c42961af.

PR #209, opened 2024-03-06
Nitpick docs

roj1512:
- Consistently refer to Node.js as Node.js, not Node.
- Other minor improvements.


ry (commented):
I mean, if that helps readers. But the name of project is "Node".

roj1512:
@ry It is referred to Node.js in their docs, and you already have a number of mentions of it as _Node.js_. Here’s an example:

https://github.com/jsr-io/jsr/blob/9fd9987661488134989623578d69c8955ee1dfe2/frontend/docs/npm-compatibility.md?plain=1#L7

Regardless of which one is right, my goal is to make it consistent. Do you want to switch to _Node_?

frontend/docs/badges.md
```diff
@@ -1,63 +1,62 @@
 ---
-description: JSR provides a badges for each package that can be used to showcase the package version or JSR score for a package.
+description: JSR provides badges for each package that can be used to showcase its latest version or its score.
 ---
 
-For every package, JSR provides an SVG badge that can be included in READMEs,
-documentation, or other places to showcase the package version or JSR score for
-a package.
+JSR provides badges for each package that can be used to showcase its latest
+version or its score.
```
lucacasonato, line 6:
```suggestion
version or its JSR score.
```

frontend/docs/badges.md
```diff
@@ -1,63 +1,62 @@
 ---
-description: JSR provides a badges for each package that can be used to showcase the package version or JSR score for a package.
+description: JSR provides badges for each package that can be used to showcase its latest version or its score.
```
lucacasonato, line 2:
```suggestion
description: JSR provides badges for each package that can be used to showcase its latest version or its JSR score.
```

frontend/docs/badges.md
```diff
@@ -1,63 +1,62 @@
...
+[![JSR Score](https://jsr.io/badges/@<scope>/<package>/score)](https://jsr.io/@<scope>/<package>)
 ```
 
-In a HTML document, use the following code:
+In HTML documents, the following can be used:
 
 ```html
 <a href="https://jsr.io/@<scope>/<package>">
-  <img src="https://jsr.io/badges/@<scope>/<package>/score" alt="">
+  <img src="https://jsr.io/badges/@<scope>/<package>/score" alt="JSR Score" />
```
lucacasonato, line 60: The lack of `alt` was intentional - we can provide a better alt text in the svg itself. Please revert.

roj1512: Done, but I don’t understand. How can the alt in the SVG be effective if the SVG itself fails to load?

lucacasonato: That is not what alts are for. Alt is the alternate text for screen readers.

roj1512: Alright.

lucacasonato (approved):
Thank you

PR #228, opened 2024-03-07
feat: revamp the publishing interstitial

### LINKED ISSUE #200: On attempting to publish a duplicate version from the CLI, browser is opened to a page that never closes
I tried to publish a package version that already existed, but was brought to this browser page which never closed. I went back to the terminal eventually, but it would be nice to not sit on this page as watch it do nothing if `deno publish` fails

image

lucacasonato:
image

Fixes #200


frontend/routes/publishing/(_islands)/publishing.tsx
```diff
@@ -0,0 +1,189 @@
...
+        props.status.value = {
+          loading: false,
+          task,
+        };
+        if (task?.status === "success" || task?.status === "failure") break;
+        await new Promise((resolve) => setTimeout(resolve, 4000));
+      }
+    })().catch((e) => console.error("polling error", scope, name, version, e));
+
+    (() => cancel = true);
```
marvinhagemeister, line 51: Nit: Missing return statement. I think this should be:

```suggestion
    return () => (cancel = true);
```

marvinhagemeister (approved):

PR #235, opened 2024-03-07
fix: clean up redirects around package pages

lucacasonato:
Previously we redirected you to the publish page, even if you had a prerelease version published.

This handles the "only prereleases" scenario better in a bunch of places. Additionally it fixes an access control check on the `/publish` page - previously it was only visible to admins. Now it's also visible to package members. The content of the page is slightly different for admins and regular members.

Fixes ##222


crowlKats (approved):
LGTM

PR #236, opened 2024-03-07
fix: include other non-graph files in NPM tarball

lucacasonato:
Reincarnation of #56

Co-authored-by: Marvin Hagemeister <marvin@deno.com>


marvinhagemeister (approved):
Nice! Let's update the tarball revision too 👍

PR #238, opened 2024-03-07
fix: update include/exclude docs

dsherret:


ghostdevv:
The schema would need updating too, it's at `frontend/static/schema/config-file.v1.json`

lucacasonato (approved):
LGTM

PR #252, opened 2024-03-08
fix: update name and email from GitHub every sign-in

### LINKED ISSUE #173: Account names aren’t updated when a user changes their GitHub account’s name
Perhaps it should update on each sign in?

lucacasonato:
Previously we'd only sync name and email from GitHub on signup. We now do this in both signup, and signin.

Fixes #173

### DIFF
--- a/api/src/auth.rs
+++ b/api/src/auth.rs
@@ -145,6 +145,7 @@ async fn generate_access_token(
 
   db.upsert_github_identity(github_identity).await?;
 
+  let name = gh_user.name.unwrap_or(gh_user.login);
   let gh_email = match gh_user.email.as_ref() {
     Some(email) => Some(email.clone()), // Email address from public profile.
     None => gh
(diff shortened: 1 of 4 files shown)


api/src/auth.rs
```diff
@@ -156,28 +157,22 @@ async fn generate_access_token(
...
-
-      db.insert_user(user).await?.id
-    }
+  let new_user = NewUser {
+    name: name.as_str(),
+    email: gh_email.as_deref(),
+    avatar_url: gh_user.avatar_url.as_str(),
+    github_id: Some(gh_user.id),
+    is_blocked: false,
+    is_staff: false,
```
marvinhagemeister, line 166: I might be reading this wrong, wouldn't that mean we'd always set these two values to `false` regardless of whether the user is staff or is blocked upon login?

marvinhagemeister: oh good point!

marvinhagemeister (approved):

PR #255, opened 2024-03-09
chore: add table design

ryuapp:
The current table design seems to be default design.
I add table styles to match other designs.

Before
image

After
image
https://jsr.io/@ryu/enogu


lucacasonato (approved):
Great, thank you!

PR #257, opened 2024-03-09
fix(frontend): code block style of file

kecrily:
## Summary

<table>
  <tr>
    <th>Before</th>
    <th>After</th>
  </tr>

  <tr>
    <td>

image

</td>
    <td>

image

</td>
  </tr>
</table>

Test link: <https://jsr.io/@oak/oak/14.2.0/deno.json>

## Solution

The `flex` of the code block is overridden by the `display: block` in `.ddoc .markdown pre>code:first-child`. 

image

[The html in the code block section comes from outside](https://github.com/jsr-io/jsr/blob/acb39921e96fb2b436d352650a1f7675d0982e98/frontend/routes/package/source.tsx#L117-L124) and it is not suitable to change `flex` to `!flex`. So choose `style.css` to add styles to override the `.ddoc .markdown pre>code:first-child` default.

### DIFF
--- a/frontend/static/styles.css
+++ b/frontend/static/styles.css
@@ -197,6 +197,10 @@ body .ddoc .markdown pre {
   
   & > code:first-child {
     @apply p-6 flex;
+  
+    &.flex {
+      display: flex;
+    }
   }
 }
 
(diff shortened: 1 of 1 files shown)


crowlKats:
this should be fixed upstream

lucacasonato (approved):
Thanks.

Your fix broke styling on some other inline codeblocks, but I have fixed this by targeting your fix a bit more to the specific relevant element.

image

@crowlKats can you upstream this fix into `deno_doc` for the next upgrade?

Issue #41, opened 2024-02-28
Sidebar on package page is full screen height, even if this doesn't make sense


lucacasonato:
![image](https://github.com/jsr-io/jsr/assets/7829205/f11327d5-1d9b-4563-a39e-d531460b4aad)

Here to view the bottom of the sidebar, I need to scroll the viewport (the page), and the sidebar all the way to the bottom. This makes no sense, because the page content is small enough to fit on the page without the viewport having to scroll.

The current behaviour only made sense if the content was larger than the viewport (which it is not).

jollytoad:
This sidebar really needs to scroll independently of the main content, so far this is probably my number one bugbear with the UI atm. I tried to take a look at fixing this, but I've no idea what's going on with all the tailwind gubbins!

yoavbls:
My PR here solves this: https://github.com/jsr-io/jsr/pull/260
Without limiting the height at all, without scrollbars and it remains sticky

jollytoad:
The sidebar still doesn't scroll all the way to the bottom, unless you scroll the main page. You can see this behaviour here... https://jsr.io/@http/fns

PR #258, opened 2024-03-09
fix(frontend): made sidebar on package page scroll independently

furkankly:
Fixes https://github.com/jsr-io/jsr/issues/41

Since the layout is designed in a way that the viewport always scrolls (even when the page content is smaller) as it has full width top elements with a sticky sidebar and a large whitespace before the footer (with `class="mb-20"`) on every page, it made sense to make sidebar scrollable independently.

If we don't want viewport to scroll when the page content is smaller, we might want to revisit the layout. (I really like the approach of https://docs.rs on this) But for now independent scrolling should be sufficient imo.

### DIFF
--- a/frontend/routes/package/(_components)/Docs.tsx
+++ b/frontend/routes/package/(_components)/Docs.tsx
@@ -83,7 +83,7 @@ export function DocsView({ docs, params, selectedVersion }: DocsProps) {
           version={selectedVersion.version}
         />
         <div
-          class="ddoc w-full lg:min-h-0 lg:*:!h-full"
+          class="ddoc w-full lg:*:max-h-[calc(100vh-75px)]"
           dangerouslySetInnerHTML={{ __html: docs.sidepanel }}
         />
       </div>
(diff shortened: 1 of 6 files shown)


frontend/routes/package/(_components)/Docs.tsx
```diff
@@ -83,7 +83,7 @@ export function DocsView({ docs, params, selectedVersion }: DocsProps) {
           version={selectedVersion.version}
         />
         <div
-          class="ddoc w-full lg:min-h-0 lg:*:!h-full"
+          class="ddoc w-full lg:*:max-h-[calc(100vh-75px)]"
```
furkankly, line 86: `75px` is not a random number. `100vh-75px` is the correct height of scrollable area so when sidebar sticks to the top it fills the whole page. The [container](https://github.com/jsr-io/jsr/blob/73bd3a20d647485663c25a15c03c65a389a9f7ba/frontend/routes/package/(_components)/Docs.tsx#L79) has `py-4` and the [search input](https://github.com/jsr-io/jsr/blob/73bd3a20d647485663c25a15c03c65a389a9f7ba/frontend/routes/package/(_islands)/LocalSymbolSearch.tsx#L165) has `35px` of height and `m-2` which all sum up to `75px`.

lucacasonato (approved):
Great, thank you. I spent some more time cleaning up various other paddings that resulted in difficult to understand scroll behaviour on pages where the content is shorter than 100vh, and the sidebar is longer than 100vh.

This means for example that we do not show the footer on docs pages anymore.

Let's land it!

PR #260, opened 2024-03-09
Fix: symbols column trimmed when it's longer than the screen

yoavbls:
Now the left pane in a package page is cutting some of the symbols when it's bigger than the screen:  
(which is very common)
  
image
  

And this is after the fix:  
image
  
Of course, it's still sticky.  
I removed the `lg:max-h-screen` because it's limiting the height of the column,  
and I wrapped the column content in another div so it won't be longer than the container,  
which disables the `position: sticky`

It also solves this https://github.com/jsr-io/jsr/issues/41 
Without height limitation at all (like here: https://github.com/jsr-io/jsr/pull/258)


lucacasonato:
Thanks, this was now fixed by #258

PR #266, opened 2024-03-11
fix(frontend): unable to scroll the sidebar of overview tab

kecrily:
## Summary

Previously the sidebar could not be scrolled independently. The sidebar could only be scrolled if the entire page was scrolled to the bottom

And even then there were trim issues like the ones reported at https://github.com/jsr-io/jsr/pull/260.

image

<table>
  <tr>
    <th>Before</th>
    <th>After</th>
  </tr>

  <tr>
    <td>

截屏2024-03-11 23 24 03

</td>
    <td>

截屏2024-03-11 23 26 00

</td>
  </tr>
</table>

Test link: <https://jsr.io/@oak/oak>

## Solution

Add `flex`, `flex-column` class to sidebar so it doesn't overflow.

截屏2024-03-11 23 45 25

By the way, it makes the invalid `flex-none` class of the search box work.

image

image

---

close https://github.com/jsr-io/jsr/issues/41, close https://github.com/jsr-io/jsr/pull/258, close https://github.com/jsr-io/jsr/pull/260

### DIFF
--- a/frontend/routes/package/(_components)/Docs.tsx
+++ b/frontend/routes/package/(_components)/Docs.tsx
@@ -76,7 +76,7 @@ export function DocsView({ docs, params, selectedVersion }: DocsProps) {
 
   return (
     <div class="grid grid-cols-1 lg:grid-cols-4 py-2">
-      <div class="col-span-1 top-0 md:pl-0 md:pr-2 py-4 lg:sticky lg:max-h-screen box-border">
+      <div class="flex flex-col col-span-1 top-0 md:pl-0 md:pr-2 py-4 lg:sticky lg:max-h-screen box-border">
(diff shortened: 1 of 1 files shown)


yoavbls:
Hey, you mentioned my PR here (https://github.com/jsr-io/jsr/pull/260)
But I think that an independent scroll area with permanent scrollbars isn't the right solution and does not look so good, especially when there is also a table of contents like [here](http://localhost:8000/@david/dax). There is also the search up there, so it's unnecessary.

https://github.com/jsr-io/jsr/pull/260 is fixing the issues and maintaining the sticky, which I think is the desired behavior,  
but maybe @lucacasonato could put more light into that

PR #268, opened 2024-03-12
A data-driven & random `AnimatedLogo`

mxdvl:
## Why?

By extracting the logo as an ASCII drawing, and mapping over the values, we get more control
over the shape, colours, animation randomness and positioning of the internal elements.

```
  ███████
  █░█░░░█████
███░█░███░░░█
█░█░█░░░█░█░█
█░░░███░█░███
█████░░░█░█
    ███████
``` 

## What?

Iterate over the ASCII representation of the logo, transforming `█` and `░` blocks into their respective colours and discarding spaces to generate the individual SVG `rect`.

The SVG asset is a snapshot of the rendered HTML on a given page. Manual adjustments of individual blocks is easier with inline styles.

### DIFF
--- a/frontend/components/AnimatedLogo.tsx
+++ b/frontend/components/AnimatedLogo.tsx
@@ -1,3 +1,37 @@
+const getRandomStyles = () =>
+  [
+    `animation-duration: ${Math.ceil(Math.random() * 1400) + 600}ms`,
+    `animation-name: ${
+      [
+        "jsr_logo_square_slide_in_down",
+        "jsr_logo_square_slide_in_left",
+        "jsr_logo_square_slide_in_right",
+        "jsr_logo_square_slide_in_up",
+      ][Math.floor(Math.random() * 4)]
+    }`,
+  ].join(";");
+
+const colors = {
+  "░": "#f7df1e",
+  "█": "#083344",
+} as const;
+
+const squares = `
+  ███████
+  █░█░░░█████
+███░█░███░░░█
+█░█░█░░░█░█░█
+█░░░███░█░███
+█████░░░█░█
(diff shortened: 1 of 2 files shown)


josh-collinsworth (approved):
This is really, really cool. Awesome work! I love it. @lucacasonato any issues with merging this?

frontend/components/AnimatedLogo.tsx
```diff
@@ -1,396 +1,93 @@
...
       width="100%"
       height="100%"
       id="jsr-logo"
-      viewBox="0 0 833 539"
-      version="1.1"
-      xmlns="http://www.w3.org/2000/svg"
-      xmlnsXlink="http://www.w3.org/1999/xlink"
-      xml:space="preserve"
-      style="fill-rule:evenodd;clip-rule:evenodd;stroke-linejoin:round;stroke-miterlimit:2;"
+      viewBox="-2 -2 17 11"
```
mxdvl, line 42: The actual drawing is 13 by 7, but accounting for a padding of 2 all around.

We could write this thusly with less magic numbers:

```js
const padding = 2;
const width = 13;
const height = 7;
const viewBox = [
  -padding,
  -padding,
  padding + width + padding,
  padding + height + padding,
].join(' ');
```

lucacasonato:
Cool. I pushed a change that fixes a little glitch I saw where the server render and the client rendered random positions would not be in sync, causing a little jump of all the pixels immediately after page load (as soon as the logo re-rendered on the client).

Fixed this by not re-rendering the logo on the client.

Also updated the CSS to respect the `prefers-reduced-motion` user preference.

PR #272, opened 2024-03-12
fix(UI/UX): truncate versions instead of package names on Recent updates

### LINKED ISSUE #145: Long versions occlude package name in "Recent updates"
This package has a very long version.
https://jsr.io/@fedify/fedify@0.1.0-dev.20+c6eda9ec

Seems like it would be better to cut off the end of the version rather than package name.

Screenshot 2024-03-03 at 8 29 27 AM

furkankly:
Fixes #145.

This PR truncates package versions while ensuring to show full package names in a single line as:
Screenshot 2024-03-12 at 16 43 33

There is also a suggestion on excluding pre-release versions entirely [here](https://github.com/jsr-io/jsr/issues/145#issuecomment-1976846714) but we would prefer to handle that in backend imo if we decide to do it.

### DIFF
--- a/frontend/components/ListPanel.tsx
+++ b/frontend/components/ListPanel.tsx
@@ -37,11 +37,11 @@ export function ListPanel(
                 }`}
                 href={entry.href}
               >
-                <span class="block w-full truncate group-hover:text-jsr-cyan-800 pr-4 group-hover:underline">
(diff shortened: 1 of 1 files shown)


lucacasonato (approved):
LGTM!

PR #273, opened 2024-03-12
fix: punch through googlebot without breaking everything

lucacasonato:
I broke everything in my last PR (reverted now).


crowlKats (approved):
RSLGTM

marvinhagemeister (approved):

lucacasonato:
Tested this on staging - it does not break things anymore.

PR #276, opened 2024-03-13
docs: improved score description consistency

### LINKED ISSUE #244: Improve score description consistency
Some score item description is written in terms of "The package should", while others, like "No Slow Types", are written as "The package uses no slow types". This is confusing as it gives the impression the package does not use slow types, when it does. 

image

IMO the "The package should" phrasing is less confusing and should be adopted for all items.

furkankly:
Resolves #244.
...and renamed a misnamed component.
Left titles as they are since short, _positive_ sentences make more sense with the progress indicator and the checkmarks imo.

### DIFF
--- a/frontend/routes/docs/[...id].tsx
+++ b/frontend/routes/docs/[...id].tsx
@@ -25,7 +25,7 @@ interface Data {
   content: string;
 }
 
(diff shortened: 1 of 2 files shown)


lucacasonato (approved):
LGTM, thanks!

PR #277, opened 2024-03-13
feat: add github repo to api documentation

### LINKED ISSUE #267: API docs missing `githubRepository` entry
Recently I have implemented a basic support for JSR in [my dependency updater](https://github.com/hasundue/molt).
I'm thinking of adding a feature to show "what's changed" commit logs next, but find it tricky because the package details endpoint of the JSR API does not contain information on the original repository of source code, while the front-end has a link to a repository if available.

I respect that JSR doesn't even require a package to be hosted online or managed with git, but it might be cool to have such a field in the endpoint in some use cases.

Jayanth-Parthsarathy:
# Add Github Repository property to Packages
Fixes #267 

- [x] Added Github Repository property to Package and made it optional
- [x] Checked locally with `@redocly/cli`

## Screenshot

image

### DIFF
--- a/api/src/api.yml
+++ b/api/src/api.yml
@@ -1659,6 +1659,18 @@ components:
       required:
         - isAdmin
 
+    GitHubRepository:
+      type: object
+      properties:
+        owner:
+          type: string
+          description: The GitHub user / organization of the repository.
+          example: denoland
+        name:
+          type: string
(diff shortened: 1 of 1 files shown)


hasundue:
Looks good to an user :)

lucacasonato (approved):
Great, thank you!

PR #279, opened 2024-03-13
docs: Update troubleshooting.md with actorNotAuthorized

nzakas:
I just ran into the `actorNotAuthorized` error and noticed it's not listed on the Troubleshooting page, so I added it.


lucacasonato (approved):
Thank you @nzakas. I expanded it a bit more to also include another possible auth failure on GitHub Actions.

PR #282, opened 2024-03-14
feat: add option to require publishing from CI in scope settings

### LINKED ISSUE #237: Feature: disable direct publish setting
It feels like there should be an option to disable other mechanisms of publishing if you have publishing via GitHub actions enabled. I can see a situation where you might invite people to the org to administer packages but still ensure that any publishes on specific packages are always attested to and come via the "official" channel.

This could also prevent individuals from "forgetting" and attempting to run publish locally.

lucacasonato:
Allow users to require that all packages are published from CI, to prevent
accidental / malicious publishing of packages from local machines.

Fixes #237


crowlKats (approved):
LGTM

PR #284, opened 2024-03-14
fix: add `bin` field to npm compat tarballs

### LINKED ISSUE #157: Add npx-like behavior
### Context
I was wondering if we could run packages  "_on the fly_" similar to how `npx`, `yarn dlx`, `pnpm dlx` and `bunx` does.

### Behavior
Basically, the expected behavior would be the same or similar as explained in the docs of the previously mentioned commands ([npx for reference](https://docs.npmjs.com/cli/v10/commands/npx#synopsis)).

In other words, running the command will first search for the package locally. If it's found, it will run it directly. Otherwise, the command will temporarily fetch the package, run it, and then delete it after use.

### Examples
I guess this will be discussed either here, in the pr or by the team but here are a few _examples_ from the top of my head.

- `bunx jsrx @<namespace>/<package>`
- `npx jsr:@<namespace>/<package>`
- And honestly that would be it. Feel free to comment more examples.

---

##### NOTE: I hope the idea is well conveyed and there wasn't a misunderstanding with the feature, feel free to comment what you think about it or ping me if want to add something to the post.

lucacasonato:
Fixes #157

Still need to add tests.

### DIFF
--- a/frontend/docs/publishing-packages.md
+++ b/frontend/docs/publishing-packages.md
@@ -447,3 +447,39 @@ case, you can un-ignore the `dist/` directory by using a negation in the
 
 In this case, the `dist/` directory will be included when publishing, even
 though it is listed in the `.gitignore` file.
+
+## Publishing command line interfaces (CLIs)
+
+JSR supports publishing of command-line interfaces (CLIs) as part of a package.
+This allows package users to execute your CLI from the command line using
+`deno run`, `npx`, `yarn dlx`, and similar tools.
+
+Unlike npm, JSR does not have a special field in the `jsr.json` / `deno.json` to
+specify bin entrypoints. Instead, include the bin entrypoints in the `exports`
+field. If your package only has a single bin entrypoint and no other
+entrypoints, you can specify it as the default entrypoint in the `exports`
+field.
+
+```json
+// jsr.json
+{
+  "name": "@luca/greet",
+  "version": "1.0.0",
+  "exports": {
+    ".": "./cli.ts"
+  }
+}
+```
+
+Deno users will be able to use this CLI with `deno run`, as they can with any
+remote package: `deno run jsr:@luca/greet`. If the bin is instead specified in a
+non-default exports, they can still call it: `deno run jsr:@luca/greet/bin`.
+
+For users using [JSR's npm compatibility layer](/docs/npm-compatibility), a
+`bin` field is added to the `package.json` in the generated npm compatible
+tarball. JSR will put all files specified in the `exports` in this field that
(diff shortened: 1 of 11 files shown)


frontend/docs/publishing-packages.md
```diff
@@ -447,3 +447,39 @@ case, you can un-ignore the `dist/` directory by using a negation in the
...
 though it is listed in the `.gitignore` file.
+
+## Publishing command line interfaces (CLIs)
+
+JSR supports publishing of command-line interfaces (CLIs) as part of a package.
+This allows package users to execute your CLI from the command line using
+`deno run`, `npx`, `yarn dlx`, and similar tools.
+
+Unlike npm, JSR does not have a special field in the `jsr.json` / `deno.json` to
+specify bin entrypoints. Instead, include the bin entrypoints in the `exports`
```
wojpawlik, line 458: That might make it difficult to publish existing packages to JSR, unless https://github.com/nodejs/modules/issues/274

lucacasonato: Why do you say so? Your bin entrypoint does not have to be in the `main`.

wojpawlik: I'm not exactly thrilled about having to resort to [hacks](https://github.com/nodejs/node/pull/32223#issuecomment-792909161), or rethink my `package.json` to match JSR.

wojpawlik: I'd be fine with this if JSR transpiled `import.meta.main` to something Node can understand.

frontend/docs/publishing-packages.md
```diff
@@ -447,3 +447,39 @@ case, you can un-ignore the `dist/` directory by using a negation in the
...
+```
+
+Deno users will be able to use this CLI with `deno run`, as they can with any
+remote package: `deno run jsr:@luca/greet`. If the bin is instead specified in a
+non-default exports, they can still call it: `deno run jsr:@luca/greet/bin`.
+
+For users using [JSR's npm compatibility layer](/docs/npm-compatibility), a
+`bin` field is added to the `package.json` in the generated npm compatible
+tarball. JSR will put all files specified in the `exports` in this field that
+contain a [shebang](https://en.wikipedia.org/wiki/Shebang_(Unix)#Examples). JSR
```
wojpawlik, line 481: Docs should specify what shebang should universal code use. I believe it should be a Node shebang, since Deno will ignore it.

pawk3k:
@wojpawlik @lucacasonato , do you need some help to push it forward ?

frontend/docs/publishing-packages.md
AugustinMauroy: it's should be resolved in newest node.js version !

PR #299, opened 2024-03-18
chore: update reserved scope names

marvinhagemeister:


ry (approved):

api/src/reserved_scopes.json
```diff
@@ -493,6 +499,7 @@
   "serializejavascript",
   "servefavicon",
   "servestatic",
+  "shadcdn",
```
roj1512, line 502: Did you mean shadcn?

marvinhagemeister: Oh you're right 😅

Issue #307, opened 2024-03-19
Search shortcut on main page (meta + /) doesn't work anymore


bartlomieju:
<img width="1049" alt="Screenshot 2024-03-19 at 18 27 32" src="https://github.com/jsr-io/jsr/assets/13602871/0b62f37f-4666-4ba1-9d8c-eac9519774e0">

Using the keyboard shortcut suggested in the searchbox doesn't work anymore.

bartlomieju:
This might be because the same key combo is supposed to work for symbol search in package page (which works correctly):
<img width="867" alt="Screenshot 2024-03-19 at 18 30 26" src="https://github.com/jsr-io/jsr/assets/13602871/bd039c7e-1563-4bdb-822b-4c3a0bc762f7">

PR #317, opened 2024-03-21
fix: wrong key combo shown in package search

lucacasonato:
Fixes https://github.com/jsr-io/jsr/issues/307


crowlKats (approved):

bartlomieju:
Fixes https://github.com/jsr-io/jsr/issues/307?

PR #318, opened 2024-03-21
docs: add excluded-module publish error

dsherret:


frontend/docs/publishing-packages.md
```diff
@@ -447,3 +447,17 @@ case, you can un-ignore the `dist/` directory by using a negation in the
 
 In this case, the `dist/` directory will be included when publishing, even
 though it is listed in the `.gitignore` file.
+
+### Excluded module publish error
```
lucacasonato, line 451: Can you add this to `troubleshooting.md` instead of `publishing-packages.md`?

dsherret: I wondered about that, but all those errors seemed to be server side? I can move it

lucacasonato: I think it makes sense to put all the troubleshooting instructions in one place.

lucacasonato (approved):

PR #340, opened 2024-03-30
docs(native-imports): updated anchor tag in native imports

niklasmtj:
I found that the anchor tag seems to be updated in the past. Therefore the anchor is not working as expected. This PR should fix it


marvinhagemeister (approved):
Good catch, thanks for making a PR 👍

lucacasonato (approved):

PR #345, opened 2024-03-31
Set pointer-events-auto class on HomepageHero text

AdamAnSubtractM:
Set pointer-events to none for the HomepageHero canvas so that text can be selected.

### DIFF
--- a/frontend/components/HomepageHero.tsx
+++ b/frontend/components/HomepageHero.tsx
@@ -39,7 +39,7 @@ export function HomepageHero(
             <span className="sr-only">JSR</span>
             <AnimatedLogo />
             <div
-              class="text-2xl text-balance leading-[1.1] sm:text-3xl md:text-3xl lg:text-4xl opsize-normal md:opsize-sm text-center -mt-5 md:-mt-6 max-w-[20em]"
+              class="pointer-events-auto text-2xl text-balance leading-[1.1] sm:text-3xl md:text-3xl lg:text-4xl opsize-normal md:opsize-sm text-center -mt-5 md:-mt-6 max-w-[20em]"
               style="text-shadow: 0 0 2em white, 0 0 1em white, 0 0 0.5em white, 0 0 0.25em white, 0 0 3em white, 0 0 0.5em white;"
             >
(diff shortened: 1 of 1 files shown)


lucacasonato:
Not allowing to select the hero text was intentional: https://github.com/jsr-io/jsr/pull/115. Is there a specific issue you are running into?

AdamAnSubtractM:
> Not allowing to select the hero text was intentional: #115. Is there a specific issue you are running into?

Did not realize this was intentional. I was trying to copy and paste the text to send to someone and noticed I could not. Seemed like a usability issue to me and opened a PR. If the JSR team feels users shouldn't be able to select text on the page, then I don't think Its my place to say otherwise. 

Let me know if you want me to close this, otherwise, feel free to decline. Thanks for your time.

josh-collinsworth:
@AdamAnSubtractM Thanks for the contribution! I agree; this is an issue for usability.

However, this approach prevents some interactions with the background. I believe the better way to achieve the desired outcome would be to simply add the `pointer-events-auto` class to the `<div>` that contains the text in question. That will allow its text to be selectable, while keeping everything around it as-is.

AdamAnSubtractM:
> @AdamAnSubtractM Thanks for the contribution! I agree; this is an issue for usability.
> 
> However, this approach prevents some interactions with the background. I believe the better way to achieve the desired outcome would be to simply add the `pointer-events-auto` class to the `<div>` that contains the text in question. That will allow its text to be selectable, while keeping everything around it as-is.

Thanks @josh-collinsworth . I'll update the PR with your suggestions.

AdamAnSubtractM:
> @AdamAnSubtractM Thanks for the contribution! I agree; this is an issue for usability.
> 
> However, this approach prevents some interactions with the background. I believe the better way to achieve the desired outcome would be to simply add the `pointer-events-auto` class to the `<div>` that contains the text in question. That will allow its text to be selectable, while keeping everything around it as-is.

@josh-collinsworth  - PR updated, thanks.

josh-collinsworth (approved):
Perfect, thanks!

PR #348, opened 2024-04-02
fix: allow leading digits in scope and package names

crowlKats:


marvinhagemeister (approved):

lucacasonato (commented):
Does deno_doc already understand it can't use leading digits in identifiers in usage blocks?

crowlKats:
> Does deno_doc already understand it can't use leading digits in identifiers in usage blocks?

I am not sure I understand the question

marvinhagemeister:
@crowlKats It's about the usage block:

```ts
// 123lib is not a valid JS identifier. JS identifiers must not start with a number
import 123lib from "@foo/123lib";
```

Might be that this is already handled in `deno_doc`

crowlKats:
oh i see, its about auto detection of default import names. hm, that is not handled currently

api/src/npm/tarball.rs
```diff
@@ -133,7 +133,13 @@ pub async fn create_npm_tarball<'a>(
...
             .expect("parsed source should be here");
 
-          let module_info = DefaultModuleAnalyzer::module_info(&source);
+          let analyzer = DefaultModuleAnalyzer;
+
+          let module_info = analyzer.analyze(
+            module.specifier(),
+            source.text_info().text(),
+            js.media_type,
+          )?;
```
dsherret, line 142: This will reparse. I think it's now: `ParsedModuleAnalyzer::modulr_info(&source)`

dsherret (approved):
LGTM

crowlKats:
@lucacasonato i just looked through the docs and didnt find anything that talks about the first character/requiring a starting letter

lucacasonato:
`scopes.md` and `packages.md` both mention it.

crowlKats:
@lucacasonato i guess i really cannot read; fixed it up now

marvinhagemeister (approved):

lucacasonato:
Great, TY!

PR #355, opened 2024-04-04
fix: serve empty json at registry root

lucacasonato:
This allows adding JSR's npm endpoint as an upstream to private registries like Azure DevOps Artifacts.


ry (approved):
LGTM

marvinhagemeister (approved):

PR #360, opened 2024-04-05
fix: give better error on SAML auth issue

lucacasonato:


api/src/github.rs
```diff
@@ -108,25 +108,6 @@ impl GitHubUserClient {
     let repo: Repository = res.json().await?;
     Ok(Some(repo))
   }
-
-  #[instrument(name = "GitHubUserClient::get_repo_by_id", skip(self), err)]
-  pub async fn get_repo_by_id(
```
marvinhagemeister, line 113: Is this dead code that isn't used anywhere?

marvinhagemeister (approved):

api/src/github.rs
lucacasonato: no :)

PR #363, opened 2024-04-06
add scorecard pipeline

AugustinMauroy:


lucacasonato:
Can you elaborate on why this is useful? Do other repos do this?

AugustinMauroy:
Hey !
it's create a score base on some criteria. It's allow user (those who use or those who have their instance) to have a security overview.

[openssf scorecard](https://securityscorecards.dev)
[openssf](https://openssf.org)


> Do other repos do this?

nodejs/node
nodejs/security-wg
nodejs/nodejs.org

There are others but I don't know them. You have to bear in mind that the tools are new, so you need to give them time to be adopted.

AugustinMauroy:
@lucacasonato any news ??

PR #377, opened 2024-04-11
feat: redo how we transpile JSR packages for NPM

### LINKED ISSUE #180: [Feature Request] Add typings to npm package
When installing as an npm package, if the file is a mod.ts, it will not automatically detect typings, so it should be added to package.json, like:

```json
{
  "types": "./mod.d.ts"
}
```

### LINKED ISSUE #216: Typescript can't find type definitions of published package
In my JSR package I published, I specified `./mod.ts` as the export. This generates a file called `mod.js` and `mod.d.ts`. However, `mod.d.ts` is never actually used, so I get this error when I try to use the package in a Node/Yarn/Vite project:

Screenshot from 2024-03-06 15-39-04

Importing it with `/mod` at the end works fine:

image

The code also works at runtime. It's just the types.

I think the `package.json` generated by JSR is missing a "types" property. It should point to `mod.d.ts` in this case.

image

### LINKED ISSUE #313: Should include declaration maps and original ts source in npm compat
Would be helpful if the NPM compat layer still included the original ts and declaration maps so that "go to definition" in vscode worked properly.

I'm open to trying to implement. But this is here to track discussion now 😄

### LINKED ISSUE #326: Typescript Interfaces are getting Dropped
While publishing @lucsoft/web_bson we found out that some critical interfaces are getting dropped

Source:
https://github.com/lucsoft/web_bson/pull/72

### LINKED ISSUE #364: Bug in the *.d.ts transform for package
JSR Package: https://jsr.io/@nikiv/utils
Code for package: https://github.com/nikitavoloboev/ts-utils
Consumer of package: https://github.com/nikitavoloboev/python

image

As @marvinhagemeister said on Discord:

image

### LINKED ISSUE #367: JSR is not transpiling `npm:` specifiers in existing `.d.ts` files for Node compat
Failing NPM tarball test spec (PR that adds this spec format still WIP).

```
# foo.js
import { html } from "npm:lit@^2.2.7";
await import("npm:lit@^2.2.7")

# bar.mjs
import { html } from "npm:lit@^2.2.7";
await import("npm:lit@^2.2.7")

# baz.ts
import { html } from "npm:lit@^2.2.7";
await import("npm:lit@^2.2.7")

# fizz.d.ts
import { html } from "npm:lit@^2.2.7";
await import("npm:lit@^2.2.7")

# jsr.json
{
  "name": "@scope/foo",
  "version": "0.0.1",
  "exports": {
    "./foo": "./foo.js",
    "./bar": "./bar.mjs",
    "./baz": "./baz.ts",
    "./fizz": "./fizz.d.ts"
  }
}

# npm:lit@^2.2.7

# output
== /bar.mjs ==
import { html } from "lit";
await import("lit")

== /baz.js ==
await import("lit");
//# sourceMappingURL=data:application/json;base64,eyJ2ZXJzaW9uIjozLCJzb3VyY2VzIjpbImh0dHA6Ly9qc3IudGVzdC9Ac2NvcGUvZm9vLzAuMC4xL2Jhei50cyJdLCJzb3VyY2VzQ29udGVudCI6WyJpbXBvcnQgeyBodG1sIH0gZnJvbSBcIm5wbTpsaXRAXjIuMi43XCI7XG5hd2FpdCBpbXBvcnQoXCJucG06bGl0QF4yLjIuN1wiKVxuIl0sIm5hbWVzIjpbXSwibWFwcGluZ3MiOiJBQUNBLE1BQU0sTUFBTSxDQUFDIn0=

== /fizz.d.ts ==
import { html } from "npm:lit@^2.2.7";
await import("npm:lit@^2.2.7")

== /foo.js ==
import { html } from "lit";
await import("lit")

== /jsr.json ==
{
  "name": "@scope/foo",
  "version": "0.0.1",
  "exports": {
    "./foo": "./foo.js",
    "./bar": "./bar.mjs",
    "./baz": "./baz.ts",
    "./fizz": "./fizz.d.ts"
  }
}

== /package.json ==
{
  "name": "@jsr/scope__foo",
  "version": "0.0.1",
  "homepage": "http://jsr.test/@scope/foo",
  "type": "module",
  "dependencies": {},
  "exports": {
    "./foo": "./foo.js",
    "./bar": "./bar.mjs",
    "./baz": "./baz.js",
    "./fizz": "./fizz.d.ts"
  },
  "_jsr_revision": 7
}
```

### LINKED ISSUE #372: Handwritten mod.d.ts file is overridden when using the JSR npm package.
I have a handwritten mod.d.ts TypeScript definition file. However, when my package is installed via npm, the mod.d.ts file is not my custom types but a generated file based on mod.ts.

My package is [jsr/wangb__vibrato-deno](https://jsr.io/@wangb/vibrato-deno)

lucacasonato:
This fixes various bugs, and introduces some features. Specifically:

- Source code is now included in the bundle to enable better "Go to source definition"
- More usecases using `/// <reference types="..." />` work correctly now, especially when d.ts file is not a sibling of the JS file.
- We strip `/// <reference types="..." />` out of JS files correctly now.
- We transpile `npm:` and `jsr:` specifiers in user provided `.d.ts` files now.

Fixes #367
Fixes #372
Fixes #364
Fixes #313
Fixes #180
Fixes #216
Fixes #326

~~Blocked on landing https://github.com/denoland/deno_ast/pull/230 and https://github.com/denoland/deno_graph/pull/429~~ Landed


.env.example
lucacasonato, line 1: This was duplicated with the `/api/.env.example`.

api/testdata/specs/npm_tarballs/additional_files.txt
```diff
@@ -52,9 +56,14 @@ export const foo = 'bar';
...
+    "./foo": {
+      "types": "./foo.d.ts",
+      "default": "./foo.js"
+    },
+    "./bar": {
+      "default": "./bar.json"
+    }
   },
-  "_jsr_revision": 7
+  "_jsr_revision": 0
```
ry, line 67: why 0 instead of 8?

api/src/npm/specifiers.rs
```diff
@@ -1,12 +1,105 @@
...
+#[derive(Clone, Copy)]
+pub struct SpecifierRewriter<'a> {
+  pub base_specifier: &'a ModuleSpecifier,
+  pub source_rewrites: &'a HashMap<&'a ModuleSpecifier, ModuleSpecifier>,
+  pub declaration_rewrites: &'a HashMap<&'a ModuleSpecifier, ModuleSpecifier>,
+  pub dependencies: &'a IndexMap<String, Dependency>,
+}
+
+impl<'a> SpecifierRewriter<'a> {
+  pub fn rewrite(&self, specifier: &str, kind: RewriteKind) -> Option<String> {
```
ry, line 30: would be nice to have a unit test

api/src/npm/specifiers.rs
```diff
@@ -1,12 +1,105 @@
...
+      relative_import_specifier(self.base_specifier, resolved_specifier)
+    } else {
+      resolved_specifier.to_string()
+    };
+
+    Some(new_specifier)
+  }
+}
+
+pub fn relative_import_specifier(
```
ry, line 67: ditto

api/testdata/specs/npm_tarballs/additional_files.txt
lucacasonato: This is just for the fixture files, so we don't have to update all the fixture files when we update the revision.

api/src/npm/specifiers.rs
lucacasonato: This is very difficult to unit test at a level lesser than fully integrated into the publishing flow where we have a proper `deno_graph::ModuleGraph`. This is because the rewriter needs to have a valid list of resolved dependencies for the given module in the form of `Dependency` structs. These are very complicated to manually write, as they have source ranges attached.

The spec test with the .txt files is already a unit test that exercises this code, so I don't want to spend significant additional work to test this function individually. I don't think it adds much.

api/src/npm/specifiers.rs
lucacasonato: Done!

api/src/npm/import_transform.rs
```diff
@@ -0,0 +1,98 @@
...
+
+pub struct ImportRewriteTransformer<'a> {
+  pub specifier_rewriter: SpecifierRewriter<'a>,
+  pub kind: RewriteKind,
+}
+
+impl<'a> VisitMut for ImportRewriteTransformer<'a> {
+  fn visit_mut_module(&mut self, module: &mut Module) {
+    module.visit_mut_children_with(self);
+  }
```
dsherret, line 27: I think this block can be removed since I believe this is the default implementation.

api/src/npm/tarball.rs
```diff
@@ -106,155 +108,167 @@ pub async fn create_npm_tarball<'a>(
...
+          .unwrap();
+        let rewritten = rewrite_specifiers(
+          parsed_source.text_info(),
+          &module_info,
+          specifier_rewriter,
+          RewriteKind::Declaration,
+        );
+        package_files.insert(
+          js.specifier.path().to_owned(),
+          rewritten.as_bytes().to_vec(),
```
dsherret, line 216: Use `rewritten.into_bytes()` to avoid cloning and replace other occurences of `as_bytes().to_vec()`?

dsherret (approved):
LGTM!

marvinhagemeister (approved):

PR #379, opened 2024-04-12
fix: open external links in new tabs

ydcjeff:
It is a good practice to open external links in new tabs.


lucacasonato (approved):
LGTM, thanks.

@crowlKats should we do the same in rendered markdown?

PR #410, opened 2024-04-19
fix: don't ban triple slash references inside of jsdoc comments

### LINKED ISSUE #409: "Triple slash directives that modify globals" incorrectly showing up when using triple slash directives in jsdocs
I think the title is pretty clear, this error is still holding back Astral from being published on the JSR. [Here's a failed run to look at](https://github.com/lino-levan/astral/actions/runs/8745910661/job/24001796000).

Here's the code it's incorrectly tripping over:
image

Reported over email and discord, but figured I'd open an issue.

lucacasonato:
Fixes #409

### DIFF
--- a/api/src/tarball.rs
+++ b/api/src/tarball.rs
@@ -300,7 +300,7 @@ pub async fn process_tarball(
               ".".to_owned()
             } else {
               format!("./{}", sub_path)
-            } 
+            }
           } else {
             ".".to_owned()
           };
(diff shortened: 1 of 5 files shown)


api/src/tarball.rs
```diff
@@ -300,7 +300,7 @@ pub async fn process_tarball(
               ".".to_owned()
             } else {
               format!("./{}", sub_path)
-            }
+            }
```
marvinhagemeister, line 303: nit: space https://www.youtube.com/watch?v=bZUpggRKOms

marvinhagemeister (approved):

PR #411, opened 2024-04-19
fix: don't do case insensitive fs ops during graph building

lucacasonato:
Previously users could import files relatively with the wrong file casing. So far only two package versions have run into this, both from the same package.


dsherret (approved):
LGTM

PR #427, opened 2024-04-22
fix: ban import assertions during publishing

### LINKED ISSUE #418: Prevent publishing packages that use import assertions
This is related to https://github.com/denoland/deno/issues/17944.

The old import assertions proposal (`import ... assert { ... }`) has been replaced by import attributes (`import ... with { ... }`). There are three platforms that currently support the old syntax: Deno, Node.js and Chrome.  They all also support the new syntax (which is also supported by Bun and Safari).

As far as I know, Deno is the only platform that doesn't have a concrete plan yet to remove `assert`:
- Node.js will drop it in v22 (next week)
- Chrome will drop it in v126 (in June)

A good step to prevent more users from depending on the old syntax would be to ban it on JSR, so that new packages are forced to use the new syntax.

lucacasonato:
Only import attributes are allowed now.

Technically this also bans `import "./deno.json" /* assert */ with { type: "json" };` but hey, who cares?! :D

Fixes #418


marvinhagemeister:
What about dynamic assertions like:

 ```ts
await import('./foo.json', { assert: { type: 'json' }});
 ```
 
 Looking at the PR I think these are still allowed.

marvinhagemeister:
Makes sense, didn't know that the dynamic variant shared wider support in Node. In that case I'm fine with landing this.

marvinhagemeister (approved):

PR #434, opened 2024-04-23
feat: add personal access tokens

### LINKED ISSUE #393: Implementing fine grained personal access tokens
We'd like to support long lived personal access tokens to let users publish packages from CI systems other than GitHub Actions.

These personal access tokens should be fine grained: ie no "allow everything tokens". A user should be able to set one or more packages that may be published with the token.

The user should be able to set an expiry date for the token.

When a token is created, the user should be sent an email informing them of the token creation.

Personal access tokens may only be created when authenticated in an interactive session (ie on the web).

lucacasonato:
This commit adds API endpoints for managing personal access tokens. Users
can create and revoke tokens, and tokens can be used to authenticate API
requests.

Tokens are created with a description, and an optional expiry date. Tokens
can be revoked at any time.

There is a UI for managing tokens in the account settings - users can create
and revoke tokens from there. Tokens are displayed with their description,
expiry date, the granted permissions, and the date they were created.

An email is sent to users when a new token is created to notify them of the
token's creation.

The token creation UI contains a questionnaire flow to help users understand
that what they are doing is potentially dangerous, and to guide users to use
alternative authentication mechanisms such as OIDC or interactive flow where
possible.

Fixes #393

### DIFF
--- a/api/src/emails/templates/personal_access_token.html.hbs
+++ b/api/src/emails/templates/personal_access_token.html.hbs
@@ -1,0 +1,37 @@
+{{#*inline "html_inner"}}
+<h1 style="margin-top: 0; text-align: left; font-size: 24px; font-weight: 700; color: #333333">
+  Hey {{ name }},
+</h1>
+<p style="margin-top: 15px; font-size: 16px; line-height: 24px; color: #52525b">
+  We're notifying you that a new personal access token was just created for your {{ registry_name }} account.
+</p>
+<p style="margin-bottom: 15px; font-size: 16px; line-height: 24px; color: #52525b">
+  You can now use this token to authenticate with JSR. Never share this token with anyone,
+  including JSR support. You can revoke the token at any time in your account settings.
+</p>
+<table align="center" style="margin: 30px auto; width: 100%; text-align: center" cellpadding="0" cellspacing="0" role="presentation">
+  <tr>
+    <td align="center">
+      <table style="width: 100%;" cellpadding="0" cellspacing="0" role="presentation">
+        <tr>
+          <td align="center" style="font-size: 16px;">
+            <a href="{{ registry_url }}account/tokens" class="button" style="display: inline-block; color: #fff; text-decoration-line: none; line-height: 1.25; background-color: #2563eb; border-radius: 0.375rem; font-weight: 500; padding-left: 1.125rem; padding-right: 1.125rem; padding-top: 0.625rem; padding-bottom: 0.625rem">Review Tokens</a>
+          </td>
+        </tr>
+      </table>
+    </td>
+  </tr>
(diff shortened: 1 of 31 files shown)


frontend/routes/account/tokens/index.tsx
```diff
@@ -0,0 +1,207 @@
...
+                You have no personal access tokens.
+              </p>
+              <p class="mt-2">
+                <a
+                  href="/account/tokens/create"
+                  class="flex items-center gap-2 text-jsr-cyan-700 hover:text-jsr-cyan-600 hover:underline outline-none focus-visible:ring-2 ring-jsr-cyan-700 ring-offset-2 rounded-sm"
+                >
+                  <Plus />
+                  Create new token
+                </a>
```
crowlKats, line 64: maybe use the usual button rather than text? doesnt work too well in the case there is a token, so the button should be outside the list

frontend/routes/account/tokens/(_islands)/CreateToken.tsx
```diff
@@ -0,0 +1,542 @@
...
+  return (
+    <div class="mt-8 max-w-2xl">
+      <div class="bg-blue-50 border border-blue-500 p-4 rounded-lg">
+        <p class="text-gray-600">
+          Your personal access token has been created. Copy it now, as you will
+          not be able to see it again.
+        </p>
+        <code class="mt-4 p-2 pr-4 bg-blue-100 rounded-md text-sm w-full relative flex items-center justify-between">
+          <div>{token}</div>
+          <CopyButton text={token} title="Copy token" />
```
crowlKats, line 534: no hover state

crowlKats (changes requested):
odd view with 2 back buttons
Screenshot 2024-05-10 at 13 03 29

frontend/routes/account/tokens/index.tsx
lucacasonato: I tried a button first - but it was very intrusive, especially on mobile. Can we leave it as is?

api/src/emails/templates/personal_access_token.html.hbs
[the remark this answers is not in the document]
lucacasonato: Fixed: 
image

frontend/routes/account/tokens/(_islands)/CreateToken.tsx
lucacasonato: fixed

frontend/routes/account/tokens/index.tsx
crowlKats: sure

crowlKats (approved):
LGTM

PR #436, opened 2024-04-23
Add navigation menu for mobile

### LINKED ISSUE #40: Render smaller truncated menu in mobile view
Currently the package navigation takes up a lot of space in mobile view and users have to scroll quite a bit to see any of the content.

image

I believe we should explore something like this where the navigation tabs get a bit smaller, hide or shrink any chips, and limit the displayed nav items to 3-4 and put the rest in a clickable menu. This mirrors the experience users have on github so it should be a pretty familiar mobile pattern to this audience.

image

Rough example of suggested change

image

ryuapp:
close #40
If horizontal width of a navigation list exceeds a window size, it will store them in a small menu.

Preview
image


donjo (changes requested):
thank you!  so happy to see this get cleaned up.

One thing I'd like to see fixed before we merge is how this handles using tab to navigate. Right now I can't seem to navigate to the items available in the dropdown by hitting tab unless the dropdown is already open. I would expect to be able to tab to the dropdown and open it to view the contents inside right after tab is focusing on Versions.

Mobile users won't necessarily see this but a lot of people use mobile viewport widths on their desktop to have reference docs alongside their editor so it would be nice to have this working for keyboard-only users in that setting.

iuioiua:
> This being the case, I would personally champion one of the following alternatives:
> 
> 1. The items could sit together on one line that scrolls horizontally (given obvious enough scroll indicators); or
> 2. _All_ of the items could be in a single dropdown/select menu.

My $0.02 (not that anyone asked) is I like option 1.

ryuapp:
> One thing I'd like to see fixed before we merge is how this handles using tab to navigate. Right now I can't seem to navigate to the items available in the dropdown by hitting tab unless the dropdown is already open. I would expect to be able to tab to the dropdown and open it to view the contents inside right after tab is focusing on Versions.

I fixed the accessibility issue.

> This being the case, I would personally champion one of the following alternatives:
> 
> 1. The items could sit together on one line that scrolls horizontally (given obvious enough scroll indicators); or
> 2. All of the items could be in a single dropdown/select menu.

Option 1 is user-friendly on mobile, but it is very painful on PC because there is no mouse wheel for horizontal scrolling. 
So I prefer option 2 to 1. However, it is easier to navigate between screens if items of the menu is partially displayed.
If it change from `...` to `More...`, the problem may be slightly improved.

lucacasonato:
One further bit of data:

image

GitHub also does the Tab bar with overflow into a `...` menu.

donjo (approved):

PR #463, opened 2024-04-29
fix: emit .d.ts and .js files into _dist folder and other fixes

### LINKED ISSUE #462: Co-locating `.d.ts` files and `.ts` source files results in TypeScript loading the `.ts` file instead of `.d.ts` in some cases
In `https://npm.jsr.io/~/8/@jsr/std__fs/0.223.0.tgz`, a relative import to `./_create_walk_entry.d.ts` is resolved to `_create_walk_entry.ts` instead by `tsc`: https://github.com/microsoft/TypeScript/issues/58353

To work around this, we should emit all `.js` and `d.ts` files into a separate folder.

lucacasonato:
Fixes #462 

To make this work, also bans publishing packages containing a top level folder called `_dist`.

Also we now do not reference `.d.ts` files from other `.d.ts` files using the `.d.ts` extension, because otherwise tsc gets upset. So we instead use `.js` (even if the file does not exist!)

Also we emit source maps into separate files, because TSC doesn't understand embedded source maps.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -561,6 +561,9 @@ name = "bumpalo"
 version = "3.16.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "79296716171880943b8470b5f8d03aa55eb2e645a4874bdbb28adb49162e012c"
+dependencies = [
+ "allocator-api2",
+]
 
 [[package]]
 name = "byteorder"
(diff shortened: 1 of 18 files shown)


marvinhagemeister (approved):

dsherret (approved):
LGTM

marvinhagemeister (approved):

donovanglover:
Looks like this change broke my package [base16-tailwind](https://jsr.io/@donovanglover/base16-tailwind@0.1.6) since it depends on the `schemes` directory being relative to the `src` directory (`path.join(__dirname, '../schemes')`).

marvinhagemeister:
@donovanglover Can you file a new issue for that? It's much easier for us to track issues in issues, rather then comments on closed PRs

PR #468, opened 2024-04-30
fix: design update to Docs tab

### LINKED ISSUE #451: Links in descriptions on the docs overview page are un-styled
Screenshot 2024-04-26 at 10 32 41 AM
Screenshot 2024-04-26 at 10 34 28 AM

Some descriptions contain links to external content but the links are un-styled. They should either be styled like links or stripped of their anchor tag if it doesn't make sense to allow external links in that setting.

donjo:
Quite a long list of style changes affecting the Docs tab. I did them all in the jsr repo instead of ddoc because it was an easier dev workflow but it may be better to use this as a reference for what needs to change in deno_doc with the goal of eliminating nearly every `body .ddoc` style out of the stylesheet with the exception of any jsr theme colors. 

If we want to review and merge this we can but if we want to leave it as a TODO list and then close it out when deno_doc gets the updates, that's fine too. 

also fixes #451 

jsr test_@oak_oak_doc

### DIFF
--- a/frontend/static/styles.css
+++ b/frontend/static/styles.css
@@ -134,7 +134,7 @@
   }
 
   .link {
-    @apply text-jsr-cyan-700 hover:text-jsr-cyan-900 underline
+    @apply text-jsr-cyan-700 hover:text-jsr-cyan-900 underline decoration-jsr-cyan-700/50 underline-offset-2
       outline-none focus-visible:ring-2 ring-jsr-cyan-700 ring-offset-2 rounded-sm;
   }
 
--- a/frontend/static/styles.css
+++ b/frontend/static/styles.css
@@ -231,7 +231,8 @@ body .ddoc .usage nav details > summary {
 }
 
 body .ddoc .markdown .highlight .context_button {
-  @apply hover:bg-jsr-cyan-50 border-inherit;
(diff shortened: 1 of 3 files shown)


crowlKats (approved):
LGTM

Lets merge this and I'll look into backporting these changes to deno_doc in the coming days

frontend/static/styles.css
```diff
@@ -218,7 +218,71 @@ body .ddoc .markdown a:not(.no_color) {
...
+
+body .ddoc .namespaceSection .namespaceItem {
+  @apply lg:pr-4;
+  @apply md:min-h-[4rem];
+}
+
+.ddoc .markdown_summary :not(pre) > code {
+  @apply bg-jsr-cyan-100/70 !important;
+  @apply px-1.5 !important;
+}
```
crowlKats, line 266: actually this is problematic: inline code can have links, which are also blue

crowlKats (changes requested):
blocking due to issues

frontend/static/styles.css
donjo: updated to be a slightly lighter background so links pass contrast checks and wrote a style to apply link styles to links within markdown_summary (fix #451)

crowlKats (approved):

PR #476, opened 2024-05-02
fix: update styles on symbol detail pages

donjo:
First attempt at updating some of the styles on the symbol detail pages. I tried to leave some comments on bits where the thing I was changing could actually be fixed by applying a different tailwind class in deno_doc (like changing margin-top when I really want a `space-y-3` to be a `space-y-2`

Same caveat as the last one, this could be used as a TODO list for changes that should happen back in deno_doc or merged in and used as a reference for refactoring later.

Example page of these styles on Oak/oak since that contains a large amount of varied text to format. (sidebar omitted from screenshot since that's changing in a different PR and I didn't want it to distract)
jsr test_@oak_oak_doc_~_Context

### DIFF
--- a/frontend/static/styles.css
+++ b/frontend/static/styles.css
@@ -230,9 +230,19 @@ body .ddoc .anchorable.docEntry > .markdown_border {
   @apply border-none pl-0 ml-0 max-w-prose;
 }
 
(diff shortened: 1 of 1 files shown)


crowlKats (approved):
LGTM

PR #483, opened 2024-05-05
chore: update deno_doc

crowlKats:


lucacasonato (approved):

deno.json
```diff
@@ -32,6 +32,7 @@
     "target/",
     "api/testdata/",
     ".gcs/",
+    "e2e",
```
lucacasonato, line 35: This is why CI is failing.

PR #485, opened 2024-05-06
fix: show search and usage on mobile

crowlKats:
image

### DIFF
--- a/frontend/routes/package/(_components)/Docs.tsx
+++ b/frontend/routes/package/(_components)/Docs.tsx
@@ -38,8 +38,8 @@ export function DocsView(
           </div>
         )}
 
-      <div class="grid grid-cols-1 lg:grid-cols-4 gap-12">
-        <div class={`min-w-0 ${docs.toc ? "col-span-3" : "col-span-full"}`}>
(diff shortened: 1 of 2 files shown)


frontend/routes/package/(_components)/Docs.tsx
```diff
@@ -83,9 +87,8 @@ export function DocsView(
             </div>
           )}
         </div>
-
         {docs.toc && (
-          <div class="max-lg:hidden col-span-1 top-0 lg:sticky lg:max-h-screen box-border space-y-4 -mt-4 pt-4">
+          <div class="max-lg:row-start-1 lg:col-[span_1_/_-1] lg:top-0 lg:sticky lg:max-h-screen box-border space-y-4 -mt-4 pt-4">
```
josh-collinsworth, line 91: FWIW, `col-span-full` can be used instead of that JIT class

josh-collinsworth (approved):

frontend/routes/package/(_components)/Docs.tsx
crowlKats: nope, that would be `1 / -1`, not `span 1 / -1`.
I could switch it to `col-start-4`, but find that a bit less clear since it doesnt state how many cols it spans for

PR #487, opened 2024-05-07
refactor: use shields.io built-in JSR logo for badges

LitoMore:
The JSR logo has been added to Simple Icons since https://github.com/simple-icons/simple-icons/pull/10794.

I used the `logoSize` parameter to make the logo fit the badge. The feature was introduced at https://github.com/badges/shields/pull/9191.

Currently, we can't use `logoSize` in the response due to https://github.com/badges/shields/pull/10132. So I passed these parameters with search parameters instead.


lucacasonato (approved):
LGTM, thanks!

PR #493, opened 2024-05-08
fix: improve npm tarball emit

### LINKED ISSUE #477: Breaking change for file paths in recent npm compatibility update
The recent PR https://github.com/jsr-io/jsr/pull/463 causes `__dirname` to include `./_dist`, breaking packages that depend on files relative to `__dirname`.

For example, [`join(__dirname, '../schemes')`](https://jsr.io/@donovanglover/base16-tailwind/0.1.6/src/Base16Path.ts#L10) will resolve to `./schemes` in `_jsr_revision` 8 and `./_dist/schemes` in 10.

Related: https://github.com/jsr-io/jsr/issues/462

### LINKED ISSUE #491: bug: npm.jsr.io compiles incorrectly with mixed .ts and .js(+.d.ts) repositories
## tl;dr

- When resolving modules with npm.jsr.io as backend, when `.ts` and `.js` are mixed, only `.ts` is expanded to `_dist/*`, but this is out of relative path with `.js`.

Can be reproduced with the package `@mizchi/repro-jsr-npm`.

```
$ pnpm init
$ pnpm add @jsr/mizchi__repro-jsr-npm@0.0.2
```

## Reproduce

```
$ tree .
.
README.md
├── deno.json
├── mod.ts
└── sub
    ├─ lib.d.ts
    └─ lib.js
```

Published by this command.

```
deno publish --allow-slow-types
```

Error on `_dist` relative path

```bash
$ mkdir xxx
$ cd xxx
$ echo "registry=https://npm.jsr.io" > .npmrc
$ pnpm init
$ pnpm add @jsr/mizchi__repro-jsr-npm
$ tree node_modules/@jsr/mizchi__repro-jsr-npm
node_modules/@jsr/mizchi__repro-jsr-npm
├── _dist
│   ├── mod.js
│   └── mod.js.map
├── deno.json
├── mod.ts
├── package.json
└── sub
    ├── lib.d.ts
    └── lib.js
$ cat node_modules/@jsr/mizchi__repro-jsr-npm/_dist/mod.js
export * from "./sub/lib.js"; // import x from "./sub/lib.js";
//# sourceMappingURL=mod.js.map%

```

`_dist/mod.js` can not resolve uncompiled `. /sub/lib.js`

## Context

When building rust code with wasm-pack, it generates `pkg/*.js` and `pkg/*.d.ts`. wasm-pack cannot compile directly to TypeScript.

I ran into this bug when I tried to re-export it from the top-level mod.ts and check if it is importable by node. This seems to affect all relative path resolution when mixing .js, not just wasm-pack products.

(Not related to this case, but in the end node.js could not resolve the wasm binary because it did not support local fetch. This is another problem with node.js)

lucacasonato:
Some relative specifiers were being incorrectly emitted when referencing a `.js`
file from a `.d.ts` file in the dist folder.

Additionally, `.js` files were previously emitted into a `_dist` folder, which
broke users that need to relatively access files in the NPM package using
`import.meta.url` or similar. Now only `.d.ts` files are emitted into the
`_dist` folder.

Fixes #491
Fixes #477

### DIFF
--- a/api/src/npm/specifiers.rs
+++ b/api/src/npm/specifiers.rs
@@ -29,6 +29,7 @@ pub struct SpecifierRewriter<'a> {
 
 impl<'a> SpecifierRewriter<'a> {
   pub fn rewrite(&self, specifier: &str, kind: RewriteKind) -> Option<String> {
+    let source_text_specifier = specifier;
     let dep = self.dependencies.get(specifier)?;
 
     let specifier = match kind {
@@ -68,17 +69,17 @@ impl<'a> SpecifierRewriter<'a> {
       }
     }
 
-    if *resolved_specifier == *specifier {
-      // No need to rewrite if the specifier is the same as the resolved
-      // specifier.
-      return None;
-    }
-
     let new_specifier = if resolved_specifier.scheme() == "file" {
       relative_import_specifier(self.base_specifier, &resolved_specifier)
     } else {
       resolved_specifier.to_string()
     };
+
+    if &new_specifier == source_text_specifier {
+      // No need to rewrite if the specifier is the same as the resolved
+      // specifier.
+      return None;
+    }
 
     Some(new_specifier)
   }
--- a/api/src/npm/specifiers.rs
+++ b/api/src/npm/specifiers.rs
@@ -75,7 +75,7 @@ impl<'a> SpecifierRewriter<'a> {
       resolved_specifier.to_string()
     };
 
-    if &new_specifier == source_text_specifier {
+    if new_specifier == source_text_specifier {
       // No need to rewrite if the specifier is the same as the resolved
       // specifier.
       return None;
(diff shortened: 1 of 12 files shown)


api/src/npm/specifiers.rs
```diff
@@ -68,18 +69,18 @@ impl<'a> SpecifierRewriter<'a> {
...
-      return None;
-    }
-
     let new_specifier = if resolved_specifier.scheme() == "file" {
       relative_import_specifier(self.base_specifier, &resolved_specifier)
     } else {
       resolved_specifier.to_string()
     };
 
+    if &new_specifier == source_text_specifier {
```
lucacasonato, line 78: This was happening in the wrong place. If we need to update the specifier because the base specifier is different, we still need to write the updated specifier.

marvinhagemeister (approved):

PR #514, opened 2024-05-14
docs: add section about isolated declarations

### LINKED ISSUE #495: add to faq: difference between slow types and isolatedDeclarations

lucacasonato:
Fixes #495

### DIFF
--- a/frontend/docs/about-slow-types.md
+++ b/frontend/docs/about-slow-types.md
@@ -391,11 +391,11 @@ export function foo() {
 However, there are some differences between the two:
 
 - Isolated declarations require that all symbols that are exported from a module
-  follow the "no inference" rules. JSR only requires that symbos that are
+  follow the "no inference" rules. JSR only requires that symbols that are
   actually part of the public API of a package follow the "no inference" rules.
 - Isolated declarations is sometimes more strict than JSR. For example, isolated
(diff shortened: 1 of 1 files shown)


frontend/docs/about-slow-types.md
```diff
@@ -367,6 +367,37 @@ adding the following to your `deno.json` configuration file:
...
+// This is not allowed with `isolatedDeclarations`.
+export function foo() {
+  return Math.random().toString();
+}
+```
+
+However, there are some differences between the two:
+
+- Isolated declarations require that all symbols that are exported from a module
+  follow the "no inference" rules. JSR only requires that symbos that are
```
marvinhagemeister, line 394:
```suggestion
  follow the "no inference" rules. JSR only requires that symbols that are
```

frontend/docs/about-slow-types.md
```diff
@@ -367,6 +367,37 @@ adding the following to your `deno.json` configuration file:
...
+  actually part of the public API of a package follow the "no inference" rules.
+- Isolated declarations is sometimes more strict than JSR. For example, isolated
```
marvinhagemeister, line 396: Nit: Should we show an example of what that looks like? Not sure if JS devs are familiar with the term `function expression body`.

marvinhagemeister (approved):

Issue #473, opened 2024-05-02
Mobile breadcrumbs wrapping looks bad


ry:
See for example
https://jsr.io/@gameplay/games/doc/poker/~/CheckAction

![IMG_3087](https://github.com/jsr-io/jsr/assets/80/8b73be09-97dd-403f-a0af-2b33090ea78d)

PR #518, opened 2024-05-15
Fix: allow module breadcrumbs to wrap on mobile

josh-collinsworth:
Closes https://github.com/jsr-io/jsr/issues/473

Also includes some CSS auto-formatting, and the removal of a wayward server-side `console.log` that somehow made it to prod.

### DIFF
--- a/frontend/static/styles.css
+++ b/frontend/static/styles.css
@@ -127,7 +127,6 @@
     @apply active:shadow-accent-sm-close active:translate-x-[3px] active:translate-y-[4px];
   }
 
-
   .chip {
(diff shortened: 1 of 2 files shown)


frontend/static/styles.css
```diff
@@ -300,6 +305,10 @@ body .ddoc .contextLink {
   color: theme("colors.jsr-cyan.700");
 }
 
+body .ddoc .breadcrumbs {
+  @apply flex-wrap;
+}
+
```
josh-collinsworth, line 311: This is the only new CSS; the rest of the changes in this file are just auto-formatting fixes.

lucacasonato (approved):
LGTM, thx

PR #524, opened 2024-05-16
 feat: sticky breadcrumbs 

crowlKats:


donjo (approved):
works as expected on local dev, design 👍

PR #528, opened 2024-05-17
feat: package version in scope and search result lists

### LINKED ISSUE #421: suggestion: show package versions in scope package list
I see this being useful for, say, the Standard Library, allowing users to quickly see the latest version of each package at a glance. Maybe this could appear immediately to the right of the package name.

iuioiua:
Closes #421

localhost_8000_packages_search=x (1)


lucacasonato:
I think this needs some different styling. Right now it is not clear that this is the latest version.

@josh-collinsworth any ideas?

jollytoad:
For what it's worth, I had a little hack at this myself before I realised this PR was open. I think it's pretty obvious that it refers to the latest version, what else would it be? adding a latest pill would just be noise. I think that the version could be a little bit more subtle though, here was my hack at it, based on the styling of on the main package page...

```js
        <div class="grow-1">
          <div class="flex items-baseline gap-x-3">
            <span class="text-cyan-700 font-semibold">
              {`@${pkg.scope}/${pkg.name}`}
            </span>
            {'latestVersion' in pkg && (
              <span class="text-[0.75em] font-bold text-gray-500 truncate">
                <span class="relative text-[0.80em] -top-[0.175em] font-[800]">
                  @
                </span>
                {pkg.latestVersion}
              </span>
            )}
          </div>
```

Screenshot 2024-05-23 at 11 07 03 am

iuioiua:
It seems clear to me that this is the latest version. Perhaps we can merge this and then re-evaluate if we get complaints.

iuioiua:
How would you like to proceed, @lucacasonato?

iuioiua:
Gentle ping on this one once more. Having this would be nice, specifically for `std`, as we would then be able to remove the packages table in the [README](https://github.com/denoland/std#packages).

jollytoad:
Any progress on this? I'd really like to see this feature in jsr, really useful for @std and other libraries that consist of multiple packages.

josh-collinsworth (approved):
Love this. Went ahead and included some styling and layout updates (some relevant, some just unrelated things I noticed poking around).

Sorry for letting it lag for so long.

Issue #539, opened 2024-05-20
Jsr documentation badges are broken


Hexagon:
"Please use https"

![Screenshot_20240520_213844_Chrome](https://github.com/jsr-io/jsr/assets/419737/fd9820aa-5254-4dff-a5c3-bbd377c43417)

I am using the markdown provided by the official docs. It seems to affect all (checked) repos.

o-az:
same issue shows up in jsr docs: https://jsr.io/docs/badges#version-badge
![CleanShot 2024-05-20 at 17 08 25](https://github.com/jsr-io/jsr/assets/23618431/938b13e3-6189-49d4-8d2b-b4f3d530b39b)

PR #510, opened 2024-05-14
chore: remove repeated URL parsing

iuioiua:
Replaces #270

PR #540, opened 2024-05-21
fix: broken badge urls

marvinhagemeister:
We passed an incorrect url to shields.io for generating the badges. 

This regression was introduced in https://github.com/jsr-io/jsr/pull/510/files#diff-8caf5f85764d74fed531a1043245722b7f0fb15b3354df1ff3468742cd450ed7L35

Fixes https://github.com/jsr-io/jsr/issues/539


iuioiua:
Are we able to create a regression test?

lucacasonato (approved):
LGTM

PR #548, opened 2024-05-22
fix: remember the users package manager preference

### LINKED ISSUE #546: Persist "Use with" selection
**AS A** user who prefers to use Yarn, **I WANT** the "Use with" selector to remember my preference **SO THAT** I can see my preferred install command quickly for all packages.

https://github.com/jsr-io/jsr/assets/910645/7339447e-4569-4b6c-a71f-fa709230a480

lucacasonato:
The usage box will now remember what package manager the user last selected for the next page load / session.

This data is stored on device, and thus does not sync between devices.

Closes #546


crowlKats (approved):
LGTM

marvinhagemeister (approved):

PR #554, opened 2024-05-23
docs: add missing @example tag

jprinaldi:


lucacasonato (approved):
The example tag is not needed, but sure, let's add it here.

jprinaldi:
> The example tag is not needed, but sure, let's add it here.

Without it I wasn't getting the proper JSR score.

PR #556, opened 2024-05-23
feat: include module paths in Dependencies

### LINKED ISSUE #310: Suggestion: Finer grained module dependencies
Many packages are treated as libraries, where individual modules are imported, rather than a single entrypoint.

JSR `Dependencies` currently only displays the high-level overview of packages as dependencies, and doesn't reflect that only certain modules of a package are actually depended on.

I think Dependencies really needs to reflect this, to inform the viewer of the true nature of the dependency, and whether its just of a particular module or of a whole package.

Take this view for example: https://jsr.io/@http/fns/dependencies

It just lists out a set of packages, misleadingly implying that it depends on the whole of each package, yet the majority of imports for this packages is actually individual modules.

I think for full clarity, the Dependencies page should also shown which individual modules are directly imported by the package.

(I also think Dependencies should list usage of global variables to reflect using of web standard and runtime specific APIs, but that's another issue)

jollytoad:
Closes #310

I feel the Dependencies tab is currently quite misleading, and doesn't reflect that only certain modules from a package are actually depended upon.

This PR introduces add the individual modules paths with links to the whole package and to the individual module docs (for npm the module path is not linked as there is nowhere really for it to link to).

Screenshot 2024-05-23 at 12 13 42 pm

### DIFF
--- a/frontend/routes/package/dependencies.tsx
+++ b/frontend/routes/package/dependencies.tsx
@@ -27,14 +27,28 @@ export default function Deps(
 ) {
   const iam = scopeIAM(state, data.member);
 
-  const deps: Record<string, { link: string; constraints: Set<string> }> = {};
+  const deps: Record<
+    string,
+    {
+      packageName: string;
+      packageLink: string;
+      moduleName?: string;
+      moduleLink?: string;
+      constraints: Set<string>;
+    }
+  > = {};
 
   for (const dep of data.deps) {
-    const key = `${dep.kind}:${dep.name}`;
+    const key = `${dep.kind}:${dep.name}${dep.path ? `/${dep.path}` : ""}`;
     deps[key] ??= {
-      link: `${
+      packageName: `${dep.kind}:${dep.name}`,
+      packageLink: `${
         dep.kind === "jsr" ? "/" : "https://www.npmjs.com/package/"
       }${dep.name}`,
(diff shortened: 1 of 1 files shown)


lucacasonato:
Hey, sorry for the late review.

I don't like the idea that the same package will show up in multiple rows. Can you change this code so that we still only have one row per package + version constraint, but then have a smaller text below the package name that has the list of imported "exports"?

jollytoad:
Hey, no problem, glad you had chance to review.

So yeah, I agree, and I was wondering about this. I'd considered grouping paths like versions are, although if one import is for one version and another for a different version then the list of what is import from what version becomes misleading. But I expect that to be quite an unusual situation anyway.

Happy to do some tweaking now I know you are happy with the general idea of showing this additional information.

jollytoad:
Screenshot 2024-06-10 at 10 35 54 AM

jollytoad:
Any thoughts on progressing this?

crowlKats:
@jollytoad apologies for the delays on this. I am in favour of this and i like the design (though maybe the version column should be the last one?), though i will pass this through an aditional person in regards of design

josh-collinsworth:
I like it. 👍

josh-collinsworth (approved):

PR #557, opened 2024-05-23
fix: display the default package manager

ryuapp:
This fixes a bug that occurred in #548.

If you select Bun in a package(e.g. https://jsr.io/@std/assert) and move to another package(e.g. https://jsr.io/@std/archive) where Bun does not work, the package manager will not display correctly.

image


marvinhagemeister (approved):

lucacasonato (approved):
Thanks.

PR #573, opened 2024-05-27
Add `@internal`, `@invalid`, `@local` and `@private` to the list of reserved scopes

kachkaev:
These scopes are something like [special-use domain names](https://en.wikipedia.org/wiki/Special-use_domain_name). They will allow developers to create non-public packages knowing that they won’t shadow the public ones. `@example` workspace has been added in https://github.com/jsr-io/jsr/pull/371.


lucacasonato (approved):
Thanks

PR #574, opened 2024-05-27
chore: remove deno_std workspace setup from local dev instructions

manzt:
I can't find the `_tools/convert_to_workspace.ts` in `deno_std`. I believe it's likely been removed since `deno_std` is now published on jsr under `@std/`?


lucacasonato (approved):
Correct - thanks!

PR #592, opened 2024-06-03
fix(frontend): Big and small accessibility improvements

timharek:
This PR includes a bunch of big and small accessibility improvements spread across the whole frontend-part of JSR.

If you have any questions please let me know :)

Here is a list of what I've changed, fixed and improved:

- Removed all alt-text from the logos on the frontpage.
  - Since they are purely decoration they only add noise to a screen reader, eg: "Works with any runtime... <paragraph>... Deno logo. Bun logo." etc.
- Moved the timestamp in the footer inside a [`<time>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/time)-element.
- Replaced the `type="text"` for global and symbol search with `type="search"`.
- Fixed contrasting colors a bunch of different places
  - This is the one I'm most uncertain of because I don't know if you are attached to the failing colors, if they are coming from a design system or something similar.
- Fixed the "Powered by Orama" string, the screen reader was reading just "Powered by"
- Added `aria-label` to the logo/home item in the navbar.
- Added `aria-hidden` to a bunch of icons and other misc. stuff that are purely decorative.
  - As well as the canvas on the frontpage.
- Fixed "Works with" on the package page. Now it reads everything in one go instead of the full phrase: "This package works with Deno" + "This package works with Bun" etc.
- Added `aria-label` to the tooltip.
- Made @<scope>/<package> assessible with screen reader in PackageHeader

Again, if you have any questions or additions, please let me know! :)

### DIFF
--- a/frontend/components/Header.tsx
+++ b/frontend/components/Header.tsx
@@ -57,6 +57,7 @@ export function Header({
             <a
               href="/"
               class="outline-none focus-visible:ring-2 ring-cyan-700"
+              aria-label="Home"
             >
               <HeaderLogo class="h-8 flex-none" />
             </a>
--- a/frontend/components/Header.tsx
+++ b/frontend/components/Header.tsx
@@ -94,7 +94,7 @@ export function Header({
               )}
             {searchKind !== "docs" && (
               <>
-                <span class="text-gray-200 select-none">|</span>
+                <Divider />
                 <a
                   href="/docs"
                   class="link-header"
@@ -103,7 +103,7 @@ export function Header({
                 </a>
               </>
             )}
-            <span class="text-gray-200 select-none">|</span>
+            <Divider />
             {user
               ? <UserMenu user={user} sudo={sudo} logoutUrl={logoutUrl} />
               : (
@@ -127,4 +127,8 @@ export function Header({
       </div>
     </>
   );
+}
+
+function Divider() {
+  return <span class="text-gray-200 select-none" aria-hidden="true">|</span>;
 }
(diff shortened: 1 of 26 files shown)


josh-collinsworth (changes requested):
This generally looks good, from review of the code. One nit, however: I would personally prefer that instances of `aria-label="foo"` be replaced with inner text using the `.sr-only` class instead, anywhere it makes sense to do so, like this:

```html
<span class="sr-only">Text that was previously inside the aria-label attribute on the parent element</span>
```

While `aria-label="foo"` is useful for AT users, actual text on the page as `<span class="sr-only">foo</span>` accomplishes the same task, but is better for i18n/translations, since browsers can usually more easily translate content than attribute values. (It can sometimes be better in reader mode situations as well.)

timharek:
Thank you for the feedback, @josh-collinsworth!

> While aria-label="foo" is useful for AT users, actual text on the page as <span class="sr-only">foo</span> accomplishes the same task, but is better for i18n/translations, since browsers can usually more easily translate content than attribute values. (It can sometimes be better in reader mode situations as well.)

I didn't know this, thanks for letting me know! ☺️ 

> I would personally prefer that instances of `aria-label="foo"` be replaced with inner text using the `.sr-only` class instead, anywhere it makes sense to do so

I have updated the PR now with all the places I think that makes sense, and works. An example of where I think it doesn't fit is in `PackageHeader` without adding more `aria-hidden="true"` to normal text inside of the different links for repo, scope, package etc.

Let me know if you have any further suggestions or feedback!

josh-collinsworth (approved):

lucacasonato:
Thanks @timharek!

PR #616, opened 2024-06-22
fix: allow all github flavor markdown codeblock styles for examples

### LINKED ISSUE #449: Example code in the readme is not recognized
I've included example code in my readme file, but it doesn't reflect in my JSR score:

<https://jsr.io/@hongminhee/localtunnel>

katlyn:
This PR adds additional checks for [tilde fenced code blocks](https://github.github.com/gfm/#fenced-code-blocks) and [indented code blocks](https://github.github.com/gfm/#indented-code-blocks) when scoring a package on readme examples. Both fenced and indented codeblocks are checked for in the readme, whereas the `@module` JSDoc is only checked for fenced code blocks.

Closes #449

### DIFF
--- a/api/src/analysis.rs
+++ b/api/src/analysis.rs
@@ -32,6 +32,7 @@ use deno_semver::package::PackageNv;
 use deno_semver::package::PackageReqReference;
 use futures::FutureExt;
 use once_cell::sync::Lazy;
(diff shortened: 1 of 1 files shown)


lucacasonato (approved):
LGTM, thanks. Sorry for the delay.

Issue #618, opened 2024-06-24
Package's github link 


lorens-osman-dev:
Why is there no GitHub link for the packages?  It will be a nice feature if you provide a link that points to the page of package's GitHub page.

marvinhagemeister:
Maybe we should move this to a different place. Could be an effect of ["banner blindness"](https://en.wikipedia.org/wiki/Banner_blindness#:~:text=Banner%20blindness%20is%20a%20phenomenon,ignore%20is%20called%20banner%20noise.) where it's to close to something not immediately relevant to the user and being missed because of that.

<img width="817" alt="Screenshot 2024-06-24 at 09 36 06" src="https://github.com/jsr-io/jsr/assets/1062408/16274119-4fce-4773-abca-4322d0e0b199">

lorens-osman-dev:
oh my god i am blind , sorry for wasting your time

lorens-osman-dev:
what about this package ? that very confusing 

![IMG_20240624_120746](https://github.com/jsr-io/jsr/assets/114411575/7a2ea318-d6ac-4ebc-8eda-4415839341de)

marvinhagemeister:
Looks like they haven't added a link to the GitHub repository in the package metadata. The simply didn't fill out this field.

marvinhagemeister:
@lorens-osman-dev Filed an upstream issue on their tracker about the missing GitHub link: https://github.com/bsorrentino/pdf-tools/issues/9

lorens-osman-dev:
From a user experience (UX) standpoint, whenever there are files listed in
the "Files" tab, including a mandatory GitHub link would be beneficial.
While filing issues for each package might seem like a solution, it doesn't
guarantee that a GitHub link will be provided. This creates an additional
burden for JSR users and can lead to negative feedback loop amoung JSR
users.   A simpler and more efficient approach would be to enforce
including a GitHub link during package submission.

best

alectrocute:
Addressing this in https://github.com/jsr-io/jsr/pull/625. I feel this chip needs to be more obviously an external link. We should address, in the long-term, making GitHub links mandatory.

PR #625, opened 2024-07-01
#618: Add `ExternalLink` icon to GitHub Repo chip

alectrocute:
In this PR:

- An `ExternalLink` icon is added, sourced from [Hero Icons](https://heroicons.dev/).
- `ExternalLink` icon is added to the GitHub repo chip on `[routes/package/(_components)/PackageHeader.tsx](https://github.com/jsr-io/jsr/compare/main...alectrocute:AA/618?expand=1#diff-582c4864144cf2df26b1937b120ca9b37e29168c391792a36abefcfdf8e00639)` to help signify that it's an external link, a needed #618.

### DIFF
--- a/frontend/components/icons/ExternalLink.tsx
+++ b/frontend/components/icons/ExternalLink.tsx
@@ -1,0 +1,19 @@
+// Copyright 2024 the JSR authors. All rights reserved. MIT license.
+export function ExternalLink(props: { class?: string }) {
+    return (
+        <svg
+          fill="none"
+          stroke-width="1.5"
+          stroke="currentColor"
+          viewBox="0 0 24 24"
+          xmlns="http://www.w3.org/2000/svg"
+          class={`size-4 ${props.class ?? ""}`}
+        >
+          <path 
+            stroke-linecap="round"
+            stroke-linejoin="round"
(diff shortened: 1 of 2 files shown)


frontend/components/icons/ExternalLink.tsx
```diff
@@ -0,0 +1,19 @@
+// Copyright 2024 the JSR authors. All rights reserved. MIT license.
+export function ExternalLink(props: { class?: string }) {
+    return (
+        <svg
+          fill="none"
+          stroke-width="1.5"
```
donjo, line 6: I think it'd be nice to have a stroke width prop on this so it could be increased to 2.25 when it sits next to bold text like it does in this pill. 1.5 is a good default but would be nice to override that in this implementation

Screenshot 2024-07-05 at 10 10 50 AM

donjo (changes requested):
thanks for your contribution! one small design nit on the component

frontend/components/icons/ExternalLink.tsx
alectrocute: You got it. I've made these changes, let me know if you need anything else. :)

donjo (approved):

Issue #443, opened 2024-04-24
Newly published packages show a blank "Works with" in the header.


donjo:
<img width="1209" alt="Screenshot 2024-04-24 at 10 01 38 AM" src="https://github.com/jsr-io/jsr/assets/776987/90824438-323e-4a7c-9d2a-12840fd36f78">

Right now when you first publish a package the runtime compatibility options are unset so the "works with" field in the package header is blank. If you go into settings it looks like they are all set to "unknown compatibility", however, the UI doesn't reflect that until you hit save for the first time. I would suggest one of the following options:

1. New packages are set to "compatibility unknown" upon publish which would render some content in the "Works with" section of the package header and the form fields on the settings page would be correct.

**or** _(if we want compatibility definition to be fully opt-in)_

2. The form fields on the settings page are set to an empty state. The "works with" section of the package header is not visibility unless at least 1 runtime compatibility (even unknown compatibility) is defined.

alectrocute:
Addressing this in https://github.com/jsr-io/jsr/pull/627.

PR #627, opened 2024-07-01
fix(#443): Improve `RuntimeCompatIndicator` sort and accessibility

alectrocute:
In this PR:

- Improve `RuntimeCompatIndicator` sorting and accessibility
  - Sort unknown runtimes to beginning of list to improve readability at quick glance
  - Add `title` and `alt` attributes where applicable to improve accessibility and UX
- If `runtimeCompat` prop is an empty object, then show all runtimes with unknown status instead of rendering empty container, which resolves https://github.com/jsr-io/jsr/issues/443.

Screenshot 2024-07-01 at 12 53 27 PM

### DIFF
--- a/frontend/components/RuntimeCompatIndicator.tsx
+++ b/frontend/components/RuntimeCompatIndicator.tsx
@@ -25,9 +25,6 @@ export function RuntimeCompatIndicator(
     compact?: boolean;
   },
 ) {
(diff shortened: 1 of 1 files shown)


alectrocute:
@marvinhagemeister When you get a spare moment, would you mind looking this over? I feel it's the highest priority merge out of all my open PR's.

donjo (approved):
Thank you for getting to this one!

PR #638, opened 2024-07-07
feat(ogp): Add Content-Type to OGP Image

nakasyou:
I added Content-Type to OGP Image.
When it's in production environment, Content-Type be `text/html` so I fixed it.


lucacasonato (approved):
Thanks!

PR #644, opened 2024-07-12
fix: allow various crawlers for OG data scraping

### LINKED ISSUE #608: Social preview image not working in production
It seems like the recently added social preview image (#417) doesn't work in production while it was working in dev. Currently seeing a 404 for the url that is supposed to be hosting the image.

lucacasonato:
Fixes #608


josh-collinsworth (approved):
LGTM

marvinhagemeister (approved):

PR #647, opened 2024-07-12
fix: update deno_doc and other deno crates

crowlKats:
* https://github.com/denoland/deno_doc/pull/611
* https://github.com/denoland/deno_doc/pull/605


dsherret (approved):
LGTM

PR #656, opened 2024-07-16
feat: background importer for download counts

lucacasonato:


marvinhagemeister:
LGTM, CI fails because query cache wasn't updated.

crowlKats (approved):
LGTM. not familiar with bigquery, but code looks correct

PR #675, opened 2024-08-01
Small style updates and UI fixes (including one-col symbol layout)

donjo:
A few small things addressed in this PR

- One column symbol layout 
- Link to code snippets visible on hover instead of by default 
- H1 at the top of readme gets a special style treatment
- Fixed some bugs with mobile viewport rendering of the usage block

Screenshot 2024-08-01 at 4 17 56 PM

Screenshot 2024-08-01 at 4 17 10 PM

Screenshot 2024-08-01 at 4 16 46 PM

https://jam.dev/c/6a322747-f14b-4f3e-be9d-2b81669ff8d0

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -59,7 +59,7 @@ jobs:
       - uses: Swatinem/rust-cache@v2
 
       - name: Setup postgres for tests
-        run: docker-compose up -d
+        run: docker compose up -d
 
       - name: Build
         run: cargo build --all-targets --tests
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -59,7 +59,7 @@ jobs:
       - uses: Swatinem/rust-cache@v2
 
       - name: Setup postgres for tests
-        run: docker-compose up -d
+        run: docker compose up -d
```
donjo, line 62: apparently `docker-compose` is deprecated and you're supposed to use `docker compose` now. tests were failing without this change

frontend/static/styles.css
```diff
@@ -271,11 +278,60 @@ body .ddoc {
     }
   }
 
+  .usageContent .markdown {
+    @apply max-w-full;
+  }
+    
+  /* START Upstream to deno_doc  */
```
donjo, line 285: @crowlKats  this is the section that can go upstream

crowlKats (approved):

PR #678, opened 2024-08-05
Global color palette upgrade

donjo:
This PR is an attempt to get our app using the 3 main JSR theme colors (cyan, yellow, and gray) and establishing a new JSR-gray palette along the way that fits with the cyan color  a bit better. The PR is mostly a lot of find/replace for `gray-` to `jsr-gray-` (and the same for yellow and cyan) but there are some style changes made along the way to accommodate the new color palette in places where the new color didn't work in the same way, or if the original color was never a great pick in the first place. 

I may still want to do some additional tweaking to the theme to get the color palette in better shape for hitting some global accessibility goals but by changing all the tailwind variable references we would only have to change a few lines in the tailwind config and not have to do these sweeping find/replaces again anytime soon.

We do still use some other standard tailwind colors like reds for alerts and destructive actions but there isn't a global palette for those yet since they are infrequently used and not a core part of the brand.

Screenshot 2024-08-05 at 2 49 10 PM

Screenshot 2024-08-05 at 4 27 34 PM

Screenshot 2024-08-05 at 4 28 42 PM


frontend/routes/package/(_components)/PackageHeader.tsx
```diff
@@ -82,7 +82,7 @@ export function PackageHeader(
                 <span>
                   <a
                     href={`/@${pkg.scope}`}
-                    class="link font-bold no-underline"
+                    class="link font-bold pr-1 no-underline"
```
donjo, line 85: Not color related but may help with an issue of readability people have expressed where they didn't necessarily know the scope and package were distinct links here

frontend/routes/package/source.tsx
```diff
@@ -62,19 +62,23 @@ export default function PackagePage(
...
+        <div class="flex flex-row gap-1 items-center pt-1 pl-3">
           {data.sourcePath.split("/").filter((part, i) =>
             !(part === "" && i !== 0)
           ).map((part, i, arr) => {
             if (part === "") {
               // @ts-ignore ok
-              part = <span class="italic">root</span>;
+              part = (
+                <span class="text-lg font-semibold">
+                  Package root
```
donjo, line 73: Root seemed like an ambiguous word, open to other suggestions but went for something a bit more verbose that alludes to what files we're looking at here.

frontend/routes/package/source.tsx
```diff
@@ -111,9 +115,9 @@ export default function PackagePage(
               : (
                 data.source.source.view
                   ? (
-                    <div class="ddoc border border-cyan-300 rounded">
+                    <div class="ddoc">
                       <div
-                        class="markdown children:!bg-transparent"
+                        class="markdown ddoc-full children:!bg-transparent"
```
donjo, line 120: fixes a bug introduced by recent ddoc changes that were done for the docs site. This content didn't fill the container anymore like it should

frontend/components/Header.tsx
```diff
@@ -55,7 +55,7 @@ export function Header({
           {isHomepage ? <div></div> : (
             <a
               href="/"
-              class="outline-none focus-visible:ring-2 ring-cyan-700"
+              class="outline-none focus-visible:ring-2 ring-jsr-cyan700"
```
josh-collinsworth, line 58: I think this one might be a typo (shouldn't there be a hyphen between, as in `cyan-700`?)

frontend/components/Table.tsx
```diff
@@ -26,12 +26,12 @@ export function Table(
   return (
     <>
       <div
-        class={`-mx-4 md:mx-0 ring-1 ring-jsr-cyan-950 sm:rounded overflow-hidden ${
+        class={`-mx-4 md:mx-0 ring-1 ring-jsr-cyan-700/30 sm:rounded overflow-hidden ${
```
josh-collinsworth, line 29: I wouldn't block over this, but here and other places, I'd prefer we use lighter tints of brand colors (editing or adding as necessary) rather than ad-hoc transparency. I just think the probability of inconsistency is a lot higher with non-opaque values.

That is, unless there's a reason we want transparency in these cases, in which case, ignore me.

josh-collinsworth (changes requested):
One typo to address, one color issue to maybe consider or disregard, then we're good.

frontend/components/Table.tsx
donjo: Good point, I ended up changing them to a lighter tint and it was close enough but knowing I have a bunch of these strewn about all over that can become a good guide as to how best to further tweak these colors.

josh-collinsworth (approved):

PR #687, opened 2024-08-12
fix: display code copy button on hover

ryuapp:
From #675, code copy button is no longer displayed when hovered. This PR fixes it.

### DIFF
--- a/frontend/static/styles.css
+++ b/frontend/static/styles.css
@@ -301,13 +301,17 @@ body .ddoc {
     @apply rounded no-underline !important;
   }
 
-  .context_button::before {
-    content: "View code";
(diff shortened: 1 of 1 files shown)


donjo:
Could you provide a screenshot of the type of page where this button wasn't working correctly before your fix? I wasn't able to reproduce the problem on my end but maybe I wasn't looking in the right place.

Thanks!

ryuapp:
It seems that my explanation was unclear.
Currently, the copy button does not appear when we hover over `code` tags, but we can see it in this PR.
This following `code` tags are the target:

image

donjo (approved):

PR #699, opened 2024-08-25
fix: allow slsa spec compliant subjects in provenance

littledivy:
Deno PR: https://github.com/denoland/deno/pull/25200


lucacasonato (commented):
I think we just have to ignore incorrectly formatted statements for now rather than erroring. Otherwise this would have to land at the exact same time as the new CLI release going out.

Maybe you can make it an enum that supports both `Vec<Subject>` and `Subject` for the moment, and we remove the old one in a month or so?

lucacasonato (approved):
I applied the change.

LGTM now

PR #706, opened 2024-09-01
fix: correctly mark logos on docs page as undraggable

kidonng:
Fix JSX `draggable={false}` incorrectly used as HTML `draggable="false"`.


lucacasonato (approved):
LGTM, thanks

PR #718, opened 2024-09-10
fix: remove trailing slash in usage of some modules & add direct import for deno

crowlKats:


api/src/docs.rs
```diff
@@ -883,7 +883,7 @@ impl HrefResolver for DocResolver {
     let (is_main, path) = current_resolve
       .get_file()
       .map(|short_path| (short_path.is_main, &*short_path.path))
-      .unwrap_or_default();
+      .unwrap_or((true, ""));
```
marvinhagemeister, line 886: Should we add a test for this?

crowlKats: we need to rework the enture usage block infrastructure in the next few days anyways, so i think lets just land as-is

lucacasonato (approved):
Test would be nice, but I'm not sure we have tests for doc usage in general right now

crowlKats:
yea adding a test for this is tricky; i'd love to set up tests infra that lets us test individual elements on our pages, but probably not soon

Issue #717, opened 2024-09-10
module author feedback


panva:
So I gave jsr.io some time to cook since its initial release and now gave it a fair shot. Here's feedback as to the experience I had and some issues I'm seeing.

I come from a place where I will not abandon established registries and distribution means for jsr.io, jsr.io is a complementary distribution for a module.

I've published https://github.com/panva/oauth4webapi -> https://jsr.io/@panva/oauth4webapi

**The publishing**

I went straight to a GHA workflow. Setting everything up was straight forward and publishing came through on a first try.

In the "How to publish" page on jsr.io I was told that "You will need to run `deno publish` in your action." but then all the steps told me to use `npx jsr publish` instead. This confused me. I don't understand the difference between the two and nor do I want to.

<details>
<summary>Screenshot</summary>

![Screenshot 2024-09-10 at 9 46 48](https://github.com/user-attachments/assets/aada8204-da47-4a87-a7a2-62e2fba950cb)

</details>

**The registry**

I only have questions here

- Is there a way to deprecate a version to hide it from the UI?
- Is there a way to deprecate a whole module?
- Is there a way to unpublish a version?
- Is there a way to unpublish a whole module? Under what conditions?
- Can I maintain multiple semver major releases at once? If so, will the latest publish be the front page or the latest by semver?

**The package Web UI - Score**

I don't know if `Has docs for most symbols` is looking for all symbols or just the exported ones. I assume the latter but a simple wording change would suffice to resolve this.

```diff
+ At least 80% of the packages' exported symbols should have [symbol documentation](https://jsr.io/docs/writing-docs#symbol-documentation). Currently 58% of symbols are documented.
- At least 80% of the packages' symbols should have [symbol documentation](https://jsr.io/docs/writing-docs#symbol-documentation). Currently 58% of symbols are documented.
```

I'd like the score checks to be included in the CLI so that I can check them before publishing.

**The package Web UI - "Use with"**

![Screenshot 2024-09-10 at 10 01 42](https://github.com/user-attachments/assets/91894bdb-5757-4fac-817a-bfacf9a194e1)

1) This straight up doesn't work in Node.js because the trailing "/" in the import is a subpath export (`./`) which obviously isn't in the package.json file (`.` is, `./` isn't). FWIW, another question - how would I even define subpath exports for Node.js users in the jsr.json config file?
2) "mod" - let me choose the string that will be used, I doubt users appreciate pasting multiple module imports and having all of them named "mod".

[...]

PR #734, opened 2024-09-19
fix: make links relative to github for readme if github repo is present

crowlKats:
For https://github.com/jsr-io/jsr/issues/717

> We could do a fallback mechanism though: either we link to JSR source if the file exists, or to GitHub if that is specified, or it is a dead link.

_Originally posted by @lucacasonato in https://github.com/jsr-io/jsr/issues/717#issuecomment-2343859084_

### DIFF
--- a/api/src/db/models.rs
+++ b/api/src/db/models.rs
@@ -565,6 +565,12 @@ pub struct GithubRepository {
   pub created_at: DateTime<Utc>,
 }
 
+impl GithubRepository {
+  pub fn get_head_url(&self) -> String {
+    format!("https://github.com/{}/{}/blob/HEAD", self.owner, self.name)
+  }
+}
+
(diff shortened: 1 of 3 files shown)


api/src/db/models.rs
```diff
@@ -565,6 +565,12 @@ pub struct GithubRepository {
   pub created_at: DateTime<Utc>,
 }
 
+impl GithubRepository {
+  pub fn get_head_url(&self) -> String {
+    format!("https://github.com/{}/{}/blob/HEAD", self.owner, self.name)
```
lucacasonato, line 570: I feel like images should link to raw.githubusercontent.com

crowlKats: @lucacasonato adressed

lucacasonato (approved):
LGTM!

PR #736, opened 2024-09-23
feat: package archiving

### LINKED ISSUE #250: Feature Request: Yanking an entire package
Currently, you can only yank one specific version of a package, and depending on the amount of versions in that package, it is gonna take a while, yanking the entire package, wouldn't mean yanking all of the versions, but basically making all of them (including the package) act yanked, so the versions aren't yanked, but the package is, so they'll all act as if they were yanked, and yanked packages could perhaps be hidden in a tab or checkbox when looking at a scope.

crowlKats:
Closes #250


frontend/components/List.tsx
```diff
@@ -7,6 +7,7 @@ import { ComponentChildren } from "preact";
 export interface ListDisplayItem {
   href: string;
   content: ComponentChildren;
+  parentOverwrite?: string;
```
lucacasonato, line 10: `parentClass`?

api/src/api/package.rs
```diff
@@ -229,8 +229,14 @@ pub async fn list_handler(
   let db = req.data::<Database>().unwrap();
   db.get_scope(&scope).await?.ok_or(ApiError::ScopeNotFound)?;
 
-  let (total, packages) =
-    db.list_packages_by_scope(&scope, start, limit).await?;
+  let iam = req.iam();
+  let (total, packages) = if iam.check_scope_admin_access(&scope).await.is_ok()
```
lucacasonato, line 233: Maybe or if the user is admin (with sudo enabled?). May not be needed if we add a package admin page.

crowlKats: `check_scope_admin_access` already does that

crowlKats:
> We probably need to add a "packages" table in the admin dashboard now too then - because there will be packages that are not visible in the dashboard anymore.

Sure, we oculd do that, though If an admin has sudo enabled they can see the archived packages anyways, which i think is good enough

lucacasonato (approved):
Okay, LGTM

PR #739, opened 2024-09-25
fix: filter out packages starting with INTERNAL in the title from orama

crowlKats:


frontend/static/styles.css
```diff
@@ -406,14 +422,14 @@ body .ddoc:has(.toc) {
...
+    --color-accent-fg: theme(colors . slate 0.100) !important;
+    --color-border-default: theme(colors . slate 0.300) !important;
+    --bgColor-neutral-muted: theme(colors . jsr-gray 0.100) !important;
+    --color-neutral-muted: theme(colors . jsr-gray 0.600) !important;
+    --color-accent-fg: theme(colors . cyan 0.700) !important;
+    --fgColor-muted: theme(colors . jsr-gray 0.600) !important;
     --color-fg-default: theme("colors.jsr-gray.700") !important;
     --base-text-weight-semibold: 500;
-    --bgColor-muted: theme(colors.slate.50);
+    --bgColor-muted: theme(colors . slate 0.50);
```
marvinhagemeister, line 433: This formatting seems wrong. Is this intended or maybe a bug in our formatter?

crowlKats: ugh, seems so. i'll replace them all with strings for now then

marvinhagemeister (approved):

PR #765, opened 2024-10-17
fix(docs): fixes href for "Learn more about scope and package names"

tbrockman:
Current link isn't valid, I'm guessing this was the intended target.


crowlKats (approved):
LGTM, thanks

PR #776, opened 2024-10-23
feat: add symbols to orama

crowlKats:
- [x] add deploying of the index

search itself will be added next week

### DIFF
--- a/api/src/analysis.rs
+++ b/api/src/analysis.rs
@@ -65,6 +65,7 @@ pub struct PackageAnalysisOutput {
   pub data: PackageAnalysisData,
   pub module_graph_2: HashMap<String, ModuleInfo>,
   pub doc_nodes_json: Bytes,
+  pub doc_search_json: serde_json::Value,
   pub dependencies: HashSet<(DependencyKind, PackageReqReference)>,
   pub npm_tarball: NpmTarball,
   pub readme_path: Option<PackagePath>,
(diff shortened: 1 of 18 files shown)


api/src/analysis.rs
```diff
@@ -227,17 +228,52 @@ async fn analyze_package_inner(
...
+      browser: None,
+      deno: None,
+      node: None,
+      workerd: None,
+      bun: None,
+    },
+    registry_url.to_string(),
+  );
+  let search_index = deno_doc::html::generate_search_index(&ctx);
+  let doc_search_json = if let serde_json::Value::Object(mut obj) = search_index
```
lucacasonato, line 265:
```suggestion
  let doc_search_json = if let serde_json::Value::Object(mut obj) = search_index
```

You can replace this with `search_index.as_object().unwrap()`.

api/src/orama.rs
```diff
@@ -120,6 +126,89 @@ impl OramaClient {
...
+  }
+
+  #[instrument(name = "OramaClient::upsert_symbols", skip(self))]
+  pub fn upsert_symbols(
+    &self,
+    scope_name: &ScopeName,
+    package_name: &PackageName,
+    search: serde_json::Value,
+  ) {
+    let package = format!("{scope_name}/{package_name}");
```
lucacasonato, line 141: Why two calls, can't we do one?

api/src/orama.rs
```diff
@@ -120,6 +126,89 @@ impl OramaClient {
...
+          } else {
+            unreachable!()
+          }
+        })
+        .collect::<Vec<_>>()
+    } else {
+      unreachable!()
+    };
+
+    let body = serde_json::json!({ "upsert": search });
```
crowlKats, line 191: TODO: handle too big data size

api/src/analysis.rs
crowlKats: no, it returns a reference

terraform/scheduler.tf
```diff
@@ -42,4 +42,12 @@ resource "google_cloud_scheduler_job" "orama_deploy" {
       "Authorization" = "Bearer ${var.orama_package_private_api_key}"
     }
   }
+
+  http_target {
```
lucacasonato, line 46: I don't think you can have two targets. I think you need to create a second `google_cloud_scheduler_job`

crowlKats: ah damn, yea wasnt sure, was hoping it would work

lucacasonato (approved):

PR #785, opened 2024-10-25
fix: copy button overlap issue

### LINKED ISSUE #782: Copy button overlaps with content
image

> The copy button goes over the text i'm trying to highlight here, would be nice to add a larger right padding on the internal element so that you can scroll completely to the right

akmittal:
fixes #782

## Copy button overlaps with content
Add padding-right when copy button is displayed


crowlKats:
I will close this as #791 is a better solution instead, as it does not add a constant blank area

Issue #748, opened 2024-10-02
Exported Symbol Documentation is written as 80% but does not meet the 80% threshold


Sorikairox:
@danet/core package was shown has having 80% of exported symbol documented, but only get 4/5 points as if it didn't meet it

![image](https://github.com/user-attachments/assets/ffd30e64-71e4-4731-879a-6b84cdbc4a97)

( I am adding some documentation to get 5 points, so it may not show anymore. But that probable rounding issue leads to misleading information)

PR #795, opened 2024-10-28
fix(frontend): displayed percentage of documented symbols

mbhrznr:
closes https://github.com/jsr-io/jsr/issues/748.

as already expected in the issue there was some rounding happening, caused by `.toFixed(0)`:
Bildschirmfoto 2024-10-28 um 16 24 54

changes the implementation to make use of `Math.floor`.


crowlKats (approved):
LGTM

Issue #711, opened 2024-09-03
`npm:@jsr/scope__package` dependencies should link to JSR package page, not NPM


zigazajc007:
Navigate to https://jsr.io/@rabbit-company/cloudky-api@0.0.1/dependencies and click on a dependency name.

You will get redirected to https://www.npmjs.com/package/@jsr/rabbit-company__argon2id instead of https://jsr.io/@rabbit-company/argon2id

PR #796, opened 2024-10-28
fix(frontend): let `npm:@jsr/scope__package` dependency links link to `jsr`

mbhrznr:
closes https://github.com/jsr-io/jsr/issues/711.


crowlKats (approved):
LGTM

PR #804, opened 2024-11-01
feat: render package search filters

crowlKats:
Screenshot 2024-11-01 at 17 41 22


josh-collinsworth (changes requested):
One thing this doesn't seem to account for is longer text in a search box. This example might not be super likely, but with a larger font size/zoom and/or a small screen, this overflow can happen fairly easily: 

CleanShot 2024-11-01 at 16 16 43@2x

When and if that does happen, the user's cursor isn't actually where the UI shows it being, and it gets really weird. Can we find a way to account for this?

josh-collinsworth (approved):
There are still some weird edge case behaviors here. Most notably, you can't scroll the input on Safari, and sometimes, when there's lots of text, the visible cursor gets slightly out of sync with the actual cursor position.

That said, I think these are edge-case-y enough to file them as bugs and run with this.

PR #808, opened 2024-11-04
feat: support for OpenSearch

### LINKED ISSUE #457: Add OpenSearch support
This is a UX improvement which adds the ability to search JSR packages without having to navigate to jsr.io first. On Firefox, a plus icon appears on the search bar which allows the user to add jsr.io as another search engine.

npmjs.com supports this feature:

image

jsr.io does not:

image

References:

- [OpenSearch Documentation](https://github.com/dewitt/opensearch)
- [MDN: OpenSearch description format](https://developer.mozilla.org/en-US/docs/Web/OpenSearch)
- [The Chromium Project: Tab to Search](https://www.chromium.org/tab-to-search/)

ryuapp:
Closes #457 

When we open the site in Firefox's search field, we can see that new suggestions appear.
image

### DIFF
--- a/frontend/routes/_app.tsx
+++ b/frontend/routes/_app.tsx
@@ -29,6 +29,12 @@ export default async function App(
           type="image/svg+xml"
           href={asset("/logo-square.svg")}
         />
+        <link
+          rel="search"
+          type="application/opensearchdescription+xml"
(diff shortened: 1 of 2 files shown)


lucacasonato (approved):
LGTM, thanks!

PR #809, opened 2024-11-04
Add governance section

lambtron:


lambtron:
Just want to add this Governance page to include resources like the public meetings and calendar. we can iterate on copy and decide what else to add over time

lucacasonato (approved):

PR #812, opened 2024-11-05
chore: add `darwin-amd64` binaries to `tools`

mbhrznr:
whilst setting up my local env for running the entire jsr stack i couldn't help but notice that there are no compatible binaries for my machine.

this adds the required binaries as well as a new `services:macos:amd` command to `deno.json`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -114,7 +114,7 @@ making changes to the API.
 
 ### Running jsr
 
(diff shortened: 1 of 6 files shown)


crowlKats (approved):
LGTM, thanks!

Issue #713, opened 2024-09-05
Ban cjs & cts file extensions


dsherret:
These should be banned as they don't work.

mbhrznr:
how would such banning look like?
i'd assume it could be a thrown error during publish/dry run?

i'd love to give it a spin, as it might be a nice opportunity to check the inner workings of jsr.

crowlKats:
>  i'd assume it could be a thrown error during publish/dry run?

correct; and i think then on JSR's backend we should handle it as well in case

PR #818, opened 2024-11-06
fix: ban cjs and cts file extensions

mbhrznr:
closes https://github.com/jsr-io/jsr/issues/713.


api/src/analysis.rs
```diff
@@ -186,6 +186,7 @@ async fn analyze_package_inner(
       .analyzer
       .get_parsed_source(module.specifier())
     {
+      check_for_banned_extensions(&parsed_source)?;
```
mbhrznr, line 189: actually at this point we would've already run into a `graph` error.
still thought it might be a good idea to have a final sanity check.

api/src/gcp.rs
```diff
@@ -573,12 +573,18 @@ impl FakeGcsTester {
...
     assert!(self.proc.is_none());
 
-    #[cfg(target_os = "macos")]
+    #[cfg(all(target_os = "macos", target_arch = "aarch64"))]
     let p = concat!(
       env!("CARGO_MANIFEST_DIR"),
       "/../tools/bin/darwin-arm64/fake-gcs-server"
     );
 
+    #[cfg(all(target_os = "macos", target_arch = "x86_64"))]
```
mbhrznr, line 582: after already introducing the binaries in https://github.com/jsr-io/jsr/pull/812, i've introduced this to enable tests on my machine.

api/src/tarball.rs
```diff
@@ -158,6 +158,12 @@ pub async fn process_tarball(
       });
     }
 
+    if path.ends_with(".cjs") | path.ends_with(".cts") {
+      return Err(PublishError::BannedCommonJsExtension {
+        specifier: path.to_string(),
+      });
+    }
```
lucacasonato, line 165: I don't think we should do this here - users may have `.cjs` files in the tarball they are publishing, even if that `.cjs` file is not ever imported (for example because it is part of the projects' tests).

Can you remove this check, and only keep the other check?

api/src/tarball.rs
```diff
@@ -517,6 +523,9 @@ pub enum PublishError {
     column: usize,
   },
 
+  #[error("CommonJS is not allowed {specifier}")]
+  BannedCommonJsExtension { specifier: String },
```
lucacasonato, line 527: I think we can just re-use `PublishError::CommonJs` for this instead of adding a new error code.

lucacasonato (commented):
Can you also add a sentence about this in the relevant section in `frontend/docs/troubleshooting.md`

api/src/tarball.rs
mbhrznr: ah, yes. absolutely makes sense, thanks for the hint!

lucacasonato:
Why did you remove the e2e test? I thought that was good 👍

mbhrznr:
> Why did you remove the e2e test? I thought that was good 👍

the e2e test was only successful due to the `path.ends_with` check and would now throw a `graphError` returned by `deno_graph`.
would it still make sense to have that one in place, wdyt?

lucacasonato (approved):
LGTM, thanks!

PR #791, opened 2024-10-28
fix: Copy button overlaps with content

third:
Proposed fix for #782

- Added right padding
- Handled overflowing text when scrolling to the right

Screenshot 2024-10-27 at 9 27 39 PM

PR #856, opened 2024-12-03
fix: copy button overlaps with content

### LINKED ISSUE #790: Missing right padding in the "Import symbol" example
Url: https://jsr.io/@nic/decimal-literal

image

The "Import symbol" and "Import directly with a jsr specifier" code snippets should have some right padding, similarly to how they have left padding. Ideally, it should be enough padding that (when scrolling all the way to the right) the "copy" icon does not overlap with the text.

thecoder93:
Fixes #790


thecoder93:
#791 is also covered by this fix.

crowlKats:
@thecoder93 sorry for the delayed review. I just tested locally, but seems that the spacing is gone between the different spans:

Screenshot 2025-01-15 at 03 21 37

thecoder93:
@crowlKats 
Probably there was a mismatch with the old version of firefox or new styles were added. By removing inline-flex, with my modification it should work on both Chrome and firefox. With triple click it selects all the text. What do you think?

crowlKats (approved):
this looks good now! thanks!

Issue #342, opened 2024-03-31
Show number of dependencies / dependents on the web


shinebayar-g:
Currently number of dependencies / dependents are not available on the web.
![image](https://github.com/jsr-io/jsr/assets/3091558/c02a57b9-73fb-4ff5-a557-84dcc4127680)

I find them super useful to quickly check how many dependencies / dependents it has.
![image](https://github.com/jsr-io/jsr/assets/3091558/67a5ef2e-bb54-4003-916b-aaa7f8bc17cb)

aspizu:
Why is `package_lastest_version` a optional?

marvinhagemeister:
@aspizu Because a package might have been created on JSR with no code being been published to it yet. Creating a package on JSR just means mostly registering the package name. It's required before being able to publish to it. In that time frame there are no package versions.

aspizu:
so, number of dependencies / dependents would need to be optional? I'm trying to get the toolchain working on my system but sqlx is failing to permission errors, I was running the first command with `sudo`.

aspizu:
Could you please guide me on setting up the test database.

```
error returned from database: password authentication failed for user "aspizu"
```

My `api/.env` is
```ini
DATABASE_URL=postgres://aspizu@localhost/registry
```

BorisLord:
Hey, i'd like to work on this feature,
Is there any requirements to follow ?

PR #857, opened 2024-12-04
feat : Show dependencies dependents count #342

BorisLord:
Fixes https://github.com/jsr-io/jsr/issues/342.

I think a should export some code to improve visibility in of the `get_handler` function in `package.rs` into an other file but where ? `utils.rs` ?

Ready to read suggestions and make changes if needed !

### DIFF
--- a/api/src/api/package.rs
+++ b/api/src/api/package.rs
@@ -71,6 +71,7 @@ use crate::iam::ReqIamExt;
 use crate::ids::PackageName;
 use crate::ids::PackagePath;
 use crate::ids::ScopeName;
+use crate::ids::Version;
 use crate::metadata::PackageMetadata;
 use crate::metadata::VersionMetadata;
 use crate::npm::generate_npm_version_manifest;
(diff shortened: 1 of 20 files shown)


api/src/api/package.rs
```diff
@@ -302,12 +304,38 @@ pub async fn get_handler(req: Request<Body>) -> ApiResult<ApiPackage> {
...
+      .list_package_version_dependencies(&scope, &package, &latest_version)
+      .await?;
+
+    api_package.dependency_count = {
+      let mut unique_deps = HashSet::new();
+      deps
+        .iter()
+        .filter(|dep| unique_deps.insert(&dep.dependency_name))
+        .count() as u64
+    };
```
crowlKats, line 327: maybe convert this to a database COUNT as well, especially since this is an active hotpath and recomputing this constantly seems a bit unecessary

crowlKats:
@BorisLord overall this PR looks good to me; any reason this is marked as a draft?

api/src/api/package.rs
BorisLord: ok i'll do it.

BorisLord:
@crowlKats  There's no longer a good reason being marked at draft

crowlKats (commented):
LGTM, just waiting on the change to use a database COUNT instead of manual

BorisLord:
@crowlKats Everything looks good from my side !

crowlKats (approved):
LGTM!

api/src/api/package.rs
```diff
@@ -333,12 +334,30 @@ pub async fn get_handler(req: Request<Body>) -> ApiResult<ApiPackage> {
...
+  let res_package = db
     .get_package(&scope, &package)
     .await?
     .ok_or(ApiError::PackageNotFound)?;
 
-  Ok(ApiPackage::from(package))
+  let mut api_package = ApiPackage::from(res_package);
+
+  if let Some(latest_v) = &api_package.latest_version {
+    let latest_version = Version::new(latest_v).unwrap();
```
KnorpelSenf: @BorisLord @crowlKats the count should be based on the displayed version, not the latest one. See #1259

PR #867, opened 2024-12-11
fix: prevent layout shift when hovering over doc header

### LINKED ISSUE #866: Layout shift when hovering over header in docs
When you hover the mouse over an exported function's header, the "view code" button causes everything below the header to shift down several pixels. When scrolling through and exploring docs, this shift can cause you to lose your place on the page and get confused.

https://github.com/user-attachments/assets/9d649100-f5ec-4260-91bc-30aee81c2b82

Specifically, it appears the button goes from 26px tall to 28px tall because the `::before` containing the "View Code" text is 16px tall, while the `svg` is 14px tall.

To fix, couple of options:
1. Remove `line-height: 1rem` from `body .ddoc { & .symbolGroup article > div:hover > div + a.context_button::before` 
2. Change `line-height: 1rem` to `line-height: 0.75rem` (to match the font size)

https://github.com/user-attachments/assets/ee2430e7-58a1-4887-9a40-2f77e83e6ac4

Both seem to work fine for mobile and desktop, but I'll open a PR implementing option 2 to align the line-height and font-size, common for centering text vertically.

aklinker1:
This closes #866, see issue for video of original issue. Resolves the shift caused by mismatch in the SVG and text heights.

https://github.com/user-attachments/assets/1b533313-60bc-4637-97d9-f0b51c4f4106

### DIFF
--- a/frontend/static/styles.css
+++ b/frontend/static/styles.css
@@ -475,7 +475,7 @@ body .ddoc {
 
   .symbolGroup article > div:hover > div + a.context_button:before {
     content: "View code";
-    @apply hidden md:block text-xs;
+    @apply hidden md:block text-xs leading-none;
   }
 
   .docEntryHeader:hover > .context_button {


dsherret (approved):
Thanks! This annoyed me too.

PR #868, opened 2024-12-14
fix: allow publishing with no version in config file

dsherret:
https://github.com/dsherret/jsr-publish-on-tag/actions/runs/12332179143/job/34419677355

```
$ deno publish --set-version 0.2.0
Download https://registry.npmjs.org/@types/node/-/node-18.16.19.tgz
Check file:///home/runner/work/jsr-publish-on-tag/jsr-publish-on-tag/src/main.ts
Checking for slow types in the public API...
Check file:///home/runner/work/jsr-publish-on-tag/jsr-publish-on-tag/src/main.ts
Publishing @david/publish-on-tag@0.2.0 ...
error: Failed to publish @david/publish-on-tag@0.2.0
Caused by:
    Failed to publish @david/publish-on-tag at 0.2.0: invalid config file '/deno.json': missing field `version`
Error: Process completed with exit code 1.
```

### DIFF
--- a/api/src/tarball.rs
+++ b/api/src/tarball.rs
@@ -247,7 +247,6 @@ pub async fn process_tarball(
       publish_task_name: publishing_task_scoped_package_name,
     });
   }
-
   if let Some(config_file_version) = config_file.version {
     if config_file_version != publishing_task.package_version {
       return Err(PublishError::ConfigFileVersionMismatch {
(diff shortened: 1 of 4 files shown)


api/src/tarball.rs
dsherret, line 1: It would be nice if this code was broken up a bit so it was more easily testable.

crowlKats:
can you update `frontend/static/schema/config-file.v1.json`?

dsherret:
Done.

crowlKats (approved):

PR #878, opened 2024-12-20
fix: Package runtime compat is now shown properly

### LINKED ISSUE #838: Runtime compatibility options not correctly reflected in settings page
The runtime compatibility settings input fields in the settings page for a package displays "Compatibility unknown" as selected even when other options are saved. When selecting an option and hitting "Save changes", the selected option is saved correctly, but all other options are then set to "unknown". Ideally these options should be synced to the current selection by default so that updating one field does not require you to re-select each of the other options as they were.

image
image

### LINKED ISSUE #863: Package runtime compat isn't showing properly in packages settings page
No matter how I modify and save Runtime Compat, it always shows as `Compatibility unknown`.

image

EGAMAGZ:
Fixes #863
Fixes  #838

### DIFF
--- a/frontend/routes/package/settings.tsx
+++ b/frontend/routes/package/settings.tsx
@@ -145,9 +145,15 @@ function RuntimeCompatEditorItem({ name, id, value }: {
         name={id}
         value={value === undefined ? "" : value ? "true" : "false"}
       >
-        <option value="">Compatibility unknown</option>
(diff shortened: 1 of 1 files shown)


EGAMAGZ:
This PR also fixes #838

crowlKats (approved):
LGTM, thanks!

PR #910, opened 2025-01-30
docs: add governance information

crowlKats:


ry (commented):
can you combine this all into a single docs page - i feel like it's too spread out. Maybe jsr.io/docs/governance

crowlKats:
including the charter? feels a bit much to have that in the same page

lucacasonato (approved):

PR #913, opened 2025-02-02
fix: support max package length of 58

lucacasonato:


devsnek (approved):

xkcdstickfigure:
Why 58 specifically?

lucacasonato:
> /// 1 (@) + 20 (scope name) + 1 (/) + 58 (package name) = 80

So that the full `@scope/package` name does not exceed 80 chars (a very standard max column width).

TriMoon:
@lucacasonato 
> > /// 1 (@) + 20 (scope name) + 1 (/) + 58 (package name) = 80
> 
> So that the full `@scope/package` name does not exceed 80 chars (a very standard max column width).

In the old days where we had monitors with small resolution that was the max line length used true...
But nowadays, we have wide screen monitors where max 80 chars looks prehistoric :wink: 
So you might rethink that max. if you guys aim for JSR being the top new thing in future...

PR #933, opened 2025-02-04
dep(frontent): use JSR instead of npm

AugustinMauroy:


lucacasonato (commented):
Orama is not continually publishing their new versions to JSR right now AFAICT.

AugustinMauroy:
> Orama is not continually publishing their new versions to JSR right now AFAICT.

So should we close this pr ?

AugustinMauroy:
We can help them by updating this script https://github.com/oramasearch/orama/blob/main/scripts/release.mjs

crowlKats:
I don't think this is needed for now, so will close. we can revisit it later, but also, using npm specifiers is not a problem.

PR #935, opened 2025-02-04
chore(logo): introduce logo variant

AugustinMauroy:
In case of dark theme for JSR we need a "inverted" logo to unsure good contrast ratio
The orignal logo use `jsr-cyan-950` and light one `jsr-cyan-50`

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
 
 <picture>
    <source media="(prefers-color-scheme: dark)" srcset="./frontend/static/logo-light.svg">
-   <img src="./frontend/static/logo.svg" width="200px" alt="the jsr logo">
+   <img src="./frontend/static/logo.svg" width="200px" alt="the jsr logo" align="right" >
 </picture>
 
 This is the source code for https://jsr.io, the new JavaScript registry.
--- a/README.md
+++ b/README.md
@@ -1,6 +1,9 @@
 # jsr.io
 
-<img src="./frontend/static/logo.png" width="200" align="right" alt="the jsr logo">
+<picture>
+   <source media="(prefers-color-scheme: dark)" srcset="./frontend/static/logo-light.svg">
+   <img src="./frontend/static/logo.svg" width="200px" alt="the jsr logo">
+</picture>
 
 This is the source code for https://jsr.io, the new JavaScript registry.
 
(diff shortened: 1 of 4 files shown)


TriMoon:
Are you refering to https://jsr.io/logo-square.svg :thinking: 
If so i think you could embed a stylesheet into the SVG that automatically adjusts the colors used IIRC:
- https://developer.mozilla.org/en-US/docs/Web/SVG/Element/style

AugustinMauroy:
IMO having two different file is better.
Because if it's needed in an other context it's maybe create strange issue. For example electron based IDE

AugustinMauroy:
And look at Deno and node, which have different branding for light and dark 
https://deno.com/brand
https://nodejs.org/en/about/branding

TriMoon:
See: https://developer.mozilla.org/en-US/docs/Web/CSS/color_value/light-dark :wink:

TriMoon:
Here is a version of https://jsr.io/logo-square.svg, with modifications that auto change colors upon changing light/dark mode :wink: 
_(Try and see)_
```svg
<svg width="100%" height="100%" viewBox="0 -3 13 13" version="1.1" xmlns="http://www.w3.org/2000/svg">
  <style>
    :root {
      color-scheme: light dark;
    }
    svg {
      > path {
        fill: light-dark(#f7df1e, #083344);
      }
      > g {
        fill: light-dark(#083344, #f7df1e);
      }
    }
  </style>
  <path d="M0,2h2v-2h7v1h4v4h-2v2h-7v-1h-4"/>
  <g>
    <path d="M1,3h1v1h1v-3h1v4h-3"/>
    <path d="M5,1h3v1h-2v1h2v3h-3v-1h2v-1h-2"/>
    <path d="M9,2h3v2h-1v-1h-1v3h-1"/>
  </g>
</svg>
```

josh-collinsworth (commented):
I actually don't feel this version works better than the existing version; the colors both being light create more of a clash than the dark blue on a dark background, to my eye.

In any case, though: I'm not convinced a secondary version is needed currently. While the dark cyan of the existing logo admittedly has low contrast against a dark background, which isn't ideal, I don't think it's currently an issue, as the logo is still legible thanks to the yellow.

If dark mode is implemented on jsr.io, we can (and should) revisit this topic then, to holistically determine whether an alternate version is needed and how it might work best in that environment.

In the meantime, though—while I appreciate the effort that went into this and the forward-looking approach—in my opinion, this isn't something we should pursue right now, and we would probably want to explore other implementations when we do.

crowlKats:
I'll close this then as per @josh-collinsworth comment

PR #946, opened 2025-02-09
Fix cross-package links in docs

nicolo-ribaudo:
Trying to get the "Expression" link at https://jsr.io/@babel-test-6ae45912/parser@0.0.7-8.0.0-alpha.15/doc/~/parseExpression to work:
- it's missing the `/doc` in the URL
- if the type is imported from a non-default entrypoint, it should link to it
- if it's imported from a specific version, it should link to that version (this makes it work with pre-releases)


crowlKats (approved):
LGTM

PR #950, opened 2025-02-10
Mark the page wrapper as a container only in the inline direction and not block

nicolo-ribaudo:
Fixes https://github.com/jsr-io/jsr/pull/945#issuecomment-2647716883. 

Only tested by changing the value in the browser devtools.


crowlKats (approved):
LGTM

Issue #947, opened 2025-02-10
leaving scope din't have any "validation" act


AugustinMauroy:
I'd need an act system to leave a scope (which is what I've just done by mistake).
For example a modal with a “are you sure you want to leave the `@am` scope”.
Or maybe you'd have to type the name of the scope before being able to leave it.

EGAMAGZ:
Duplicated with #948

JakobJingleheimer:
Augustin actually accidentally left our joint scope, `@nodejs-loaders`, and he was the sole owner (I'm just an admin). It appears now impossible for someone to get ownership of the scope.

The UI does prevent me removing myself. It should also prevent the last owner being removed too.

AugustinMauroy:
> Duplicated with https://github.com/jsr-io/jsr/issues/948

i do nothing maybe something like GH issue

AugustinMauroy:
> The UI does prevent me removing myself. It should also prevent the last owner being removed too.

it's when there are only one left admin.

I'm working on a solution to prevent miss click. It's just and input where you have to write the name of scope

AugustinMauroy:
also other issues with this section 

![Image](https://github.com/user-attachments/assets/b3676ded-1e0f-4870-932a-8cc19aab492b)

PR #952, opened 2025-02-10
fix(UI/UX): add a validation step before leaving scope

AugustinMauroy:
#947 

# Visual


crowlKats:
The input box looks very different from our other input boxes. can you maybe copy the styling?

AugustinMauroy:
> The input box looks very different from our other input boxes. can you maybe copy the styling?

It's the exact same tailwind classes:
- my input: `inline-block w-full max-w-sm px-3 input-container text-sm input`
- `ScopeInviteForm` : `inline-block w-full max-w-sm px-3 input-container text-sm input rounded-l-none`

crowlKats:
just tested locally: while i am typing i get the error popup. this is not great, especially since it shifts the entire layout. there should be some input debouncing.

AugustinMauroy:
Layout shift should be removed.

AugustinMauroy:
bump @crowlKats

PR #966, opened 2025-02-19
feat(acout page): refine

AugustinMauroy:
- add GH icon
- give space between account info and link. because in future if there are more link it's will be interesting to have such as small section 
- change ring color of avatar to fit with the ring in header

Capture d’écran 2025-02-19 à 22 06 17
Capture d’écran 2025-02-19 à 22 06 32


BlackAsLight:
> give space between account info and link. because in future if there are more link it's will be interesting to have such as small section

It's not so much of a link, but I think it would be cool if we could provide our discord tags here since discord is a very common way to have general discussions

crowlKats (approved):
LGTM

Issue #931, opened 2025-02-04
$tabler_icons is Deprecated


AugustinMauroy:
https://github.com/hashrock/tabler-icons-tsx?tab=readme-ov-file#deprecated

Solution: 
1. https://jsr.io/@preact-icons/tb
2. Copy pasted used icons in this codebase

lucacasonato:
I think we can use [@preact-icons/tb](https://jsr.io/@preact-icons/tb)

PR #967, opened 2025-02-19
refactor: replace $tabler_icons with @preact-icons/tb icons

EGAMAGZ:
Removes deprecated $tabler_icons import and replaces specific Tabler icons with equivalent icons from @preact-icons/tb across multiple components and routes. Part of the icons migration proposed in #931


AugustinMauroy:
Is it possible to have screenshot, to help reviewer ?

crowlKats (approved):
LGTM. it would be nice if the imports could be cleaned up later on to not have to do `/tb` everywhere

PR #971, opened 2025-02-21
security: add openssf scorecard

### LINKED ISSUE #969: security: OpenSSF scorecard
The openssf [scorecard](https://scorecard.dev) is a tools that generate a code (/10) this is the follow-up to good security practice for an open source project.

it's an simple workflow that run on main branch.

AugustinMauroy:
Close #969

### DIFF
--- a/.github/workflows/scorecard.yml
+++ b/.github/workflows/scorecard.yml
@@ -1,0 +1,58 @@
+name: Scorecard supply-chain security
+on:
+  # For Branch-Protection check. Only the default branch is supported. See
+  # https://github.com/ossf/scorecard/blob/main/docs/checks.md#branch-protection
+  branch_protection_rule:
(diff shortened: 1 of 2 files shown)


crowlKats:
any specific reason that you are using commit hashes instead of actual version numbers?

AugustinMauroy:
> any specific reason that you are using commit hashes instead of actual version numbers?

It's an openSSF recommendation 
https://github.com/ossf/scorecard/blob/ea7e27ed41b76ab879c862fa0ca4cc9c61764ee4/docs/checks.md#pinned-dependencies

AugustinMauroy:
@lucacasonato ???

AugustinMauroy:
@lucacasonato can I have a tumbs up ?

AugustinMauroy:
oups miss click 😅

PR #974, opened 2025-02-22
dep(twas): use JSR

AugustinMauroy:
use a JSR version of twas

https://jsr.io/@augustinmauroy/twas (published with provenance)

### DIFF
--- a/frontend/deno.json
+++ b/frontend/deno.json
@@ -26,7 +26,7 @@
     "@std/front-matter": "jsr:@std/front-matter@1",
     "@std/semver": "jsr:@std/semver@1",
 
-    "twas": "npm:twas@^2.1.3",
+    "@augustinmauroy/twas": "jsr:@augustinmauroy/twas@^1.0.0",
     "$imagescript": "https://deno.land/x/imagescript@1.3.0/mod.ts",
 
     "@deno/gfm": "jsr:@deno/gfm@0.10",
--- a/frontend/deno.lock
+++ b/frontend/deno.lock
@@ -1,6 +1,7 @@
 {
   "version": "4",
   "specifiers": {
+    "jsr:@augustinmauroy/twas@1": "1.0.0",
     "jsr:@deno/gfm@0.10": "0.10.0",
     "jsr:@denosaurs/emoji@0.3": "0.3.1",
     "jsr:@fresh/core@^2.0.0-alpha.1": "2.0.0-alpha.25",
@@ -71,6 +72,9 @@
     "npm:twas@^2.1.3": "2.1.3"
   },
   "jsr": {
+    "@augustinmauroy/twas@1.0.0": {
+      "integrity": "e6e464001c8865f9a23cbe06c7f3e0d4e44ff3701b050596e41d435546e4baa3"
+    },
     "@deno/gfm@0.10.0": {
       "integrity": "51708205e3559a4aeb6afb29d07c5bfafe7941f91bb360351ef6621de9a39527",
       "dependencies": [
@@ -1872,6 +1876,7 @@
   },
   "workspace": {
     "dependencies": [
+      "jsr:@augustinmauroy/twas@1",
       "jsr:@deno/gfm@0.10",
       "jsr:@fresh/core@^2.0.0-alpha.25",
       "jsr:@fresh/plugin-tailwind@^0.0.1-alpha.7",
@@ -1890,8 +1895,7 @@
       "npm:preact-render-to-string@6.3.1",
       "npm:preact@10",
       "npm:prismjs@^1.29.0",
-      "npm:tailwindcss@3.4",
-      "npm:twas@^2.1.3"
+      "npm:tailwindcss@3.4"
     ]
   }
 }


EGAMAGZ (commented):
I consider that it is better to keep the current library, since it has been in use for a longer time, has been extensively tested and, therefore, is more stable and reliable.

On the other hand, the library you suggest you just created it recently, exactly one hour, which implies that it has not yet been sufficiently evaluated in different scenarios. Also, since Deno is compatible with NPM, it is not strictly necessary to migrate to a library published in JSR unless there is a compelling reason, such as a deprecation or a significant advantage.

AugustinMauroy:
> has been extensively tested

That's not right my lib have more test case and 100% test coverage

> On the other hand, the library you suggest you just created it recently, exactly one hour ago, which implies that it has not yet been sufficiently evaluated in different scenarios. Also, since Deno is compatible with NPM, it is not strictly necessary to migrate to a library published in JSR unless there is a compelling reason, such as a deprecation or a significant advantage.

I agree with you on much of it. But don't forget that SJR's source code represents the entire project. And if it's too dependent on NPM, it's not very serious: a registry that needs another registry to function is “the snake that dies on its own tail”.

EGAMAGZ:
After reviewing the original package alongside your proposed implementation, I've noticed that the way time interval and time difference calculations are handled is very similar to the code found in [this repository](https://github.com/vutran/twas) (the current package used), which is licensed under MIT. The logic and structure are nearly identical, and they even share the same naming conventions and package name—aside from a few added configuration and test files.

Could you please confirm whether both codebases have been carefully compared for any discrepancies, and that the proper MIT license attribution is maintained in both cases? This similarity reinforces my earlier point that the changes may not provide a significant improvement or benefit overall compared with current package.

AugustinMauroy:
> Could you please confirm whether both codebases have been carefully compared for any discrepancies, and that the proper MIT license attribution is maintained in both cases? This similarity reinforces my earlier point that the changes may not provide a significant improvement or benefit overall compared with current package.

That just a rewrite in typescript and "modernize".
According to my research, this should not violate the MIT license.

crowlKats:
I would actually be in favour of not using a dependency, and instead use https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DurationFormat (or Duration from temporal)

AugustinMauroy:
> I would actually be in favour of not using a dependency, and instead use https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DurationFormat (or Duration from temporal)

I'll try to see how to use temporal

frontend/utils/timeAgo.ts
```diff
@@ -0,0 +1,24 @@
...
+    years: Math.floor(diff / (1000 * 60 * 60 * 24 * 365)),
+    months: Math.floor(
+      (diff % (1000 * 60 * 60 * 24 * 365)) / (1000 * 60 * 60 * 24 * 30),
+    ),
+    days: Math.floor(
+      (diff % (1000 * 60 * 60 * 24 * 30)) / (1000 * 60 * 60 * 24),
+    ),
+    hours: Math.floor((diff % (1000 * 60 * 60 * 24)) / (1000 * 60 * 60)),
+    minutes: Math.floor((diff % (1000 * 60 * 60)) / (1000 * 60)),
+    seconds: Math.floor((diff % (1000 * 60)) / 1000),
```
crowlKats, line 17: don't think going into seconds is useful

frontend/utils/timeAgo.ts
```diff
@@ -0,0 +1,24 @@
...
+  };
+
+  // Force english because JSR is an English-only project
+  // @ts-ignore - TS doesn't know about this API yet
+  const formatter = new Intl.DurationFormat("en", { style: "long" });
+  return formatter.format(duration);
```
crowlKats, line 23:
```suggestion
  return formatter.format(duration) + " ago";
```

frontend/utils/timeAgo.ts
AugustinMauroy: the previous implementation take care of seconds so I support it. But idk if needed

crowlKats:
Screenshot 2025-03-26 at 15 44 50
Screenshot 2025-03-26 at 15 45 04

just looked at this live, and seems it is kinda too detailed. @kt3k could you advise here?

kt3k:
I would only show the largest unit like:
```js
if (duration.years >= 1) {
  return formatter.format({ years: duration.years }) + " ago";
} else if (duration.months >= 1) {
  return formatter.format({ months: duration.months }) + " ago";
} else if (duration.days => 1) {
  return formatter.format({ days: duration.days }) + " ago";
} else if (duration.hours >= 1) {
  return formatter.format({ hours: duration.hours }) + " ago";
} ...
```

FYI I do something like this in my blog implementation (using Temporal.Duration object) https://github.com/kt3k/times-kt3k/blob/8353c9c4f15e23b090b7a467f098348c968be7cd/components/post.tsx#L60-L72

AugustinMauroy:
hummm interesting !

frontend/utils/timeAgo.ts
iuioiua, line 1: Try [`@std/datetime/difference`](https://jsr.io/@std/datetime/doc/~/difference).

EGAMAGZ: I like the idea :D

iuioiua:
Also, take a look at https://github.com/denoland/saaskit/blob/ba68c58ead38e97fe6c469e4ca73a464703568a5/utils/display.ts#L31-L54

frontend/utils/timeAgo.ts
```diff
@@ -0,0 +1,37 @@
...
+  } else if (duration.days >= 1) {
+    return formatter.format({ days: duration.days }) + " ago";
+  } else if (duration.hours >= 1) {
+    return formatter.format({ hours: duration.hours }) + " ago";
+  } else if (duration.minutes >= 1) {
+    return formatter.format({ minutes: duration.minutes }) + " ago";
+  } else {
+    return formatter.format({ seconds: duration.seconds }) + " ago";
+  }
+}
```
iuioiua, line 37:
```suggestion
const UNITS = [
  "years",
  "months",
  "weeks",
  "days",
  "hours",
  "minutes",
  "seconds",
] as const;
const SMALLEST_UNIT = "seconds" as const;

export function timeAgo(timestamp: number): string {
  const time = Temporal.Instant.fromEpochMilliseconds(timestamp);
  const duration = Temporal.Now.instant()
    .since(time, { smallestUnit: SMALLEST_UNIT });
  const largestUnit = UNITS.find((unit) => duration[unit] > 0) || SMALLEST_UNIT;
  // @ts-ignore - TS doesn't know about this API yet
  return new Intl.DurationFormat("en", { style: "long" })
    .format({ [largestUnit]: duration[largestUnit] }) + " ago";
}
```

iuioiua (changes requested):
Lets the runtime do the heavy lifting.

frontend/utils/timeAgo.ts
crowlKats: we cannot use temporal, because this may run on client-side

iuioiua, line 37:
```suggestion
import { difference, type Unit } from "@std/datetime/difference";

const units = [
  "years",
  "months",
  "weeks",
  "days",
  "hours",
  "minutes",
  "seconds",
] as Unit[];

export function timeAgo(timestamp: number): string {
  const duration = difference(new Date(timestamp), new Date(), { units });
  if (duration.seconds === 0) return "0 seconds ago";
  const largestUnit = units.find((unit) => duration[unit]! > 0) || "seconds";
  // @ts-ignore - TS doesn't know about this API yet
  return new Intl.DurationFormat("en", { style: "long" })
    .format({ [largestUnit]: duration[largestUnit] }) + " ago";
}
```

EGAMAGZ: I like your implementation, is cleaner and used web APIs

PR #983, opened 2025-02-22
feat(UI): have red ring when full quotas

AugustinMauroy:
**Before**:
Capture d’écran 2025-02-22 à 20 59 54
**After**:
Capture d’écran 2025-02-22 à 21 08 25

### DIFF
--- a/frontend/components/QuotaCard.tsx
+++ b/frontend/components/QuotaCard.tsx
@@ -21,22 +21,27 @@ export function QuotaCard(
 }
 
 function QuotaUsage(props: { limit: number; usage: number }) {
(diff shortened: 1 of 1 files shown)


crowlKats:
@AugustinMauroy i can already see something that will bother me, can you maybe fix this:
there is some white between the border and the fill
Screenshot 2025-02-25 at 22 50 43

AugustinMauroy:
Oh wow good eyes ! I'll check what I can do but it's strange because it's just tailwind thing ....

AugustinMauroy:
okay little bit tricky but I go a solution

crowlKats (approved):
LGTM, thanks!

PR #984, opened 2025-02-22
UI: prevent jumping on NavItem

### LINKED ISSUE #982: Replace `border` to `outline` in `<NavItem>` to prevent jumping
Right now, when I click in `<NavItem>` (package‘s tabs like `Files`, `Versions`) elements in navigation bar jumps by 1 px. The same behavior by moving `:focus` there by pressing <kbd>Tab</kbd>.

Here is an example from [`@sitnik/nanoid`](https://jsr.io/@sitnik/nanoid@5.1.2).

https://github.com/user-attachments/assets/29bb812b-e669-4d27-b060-c10535a17522

It happens because on `:focus` we are adding `border` (which change element size). To avoid this problem it is recommended to use `outline` or `box-shadow`.

I suggest use `outline` for `:focus` style.

I also suggest use `:focus-visible` instead of `:focus`. This style will be added not on every click by only if user use keyboard navigation. It is also [a good practice](https://css-tricks.com/almanac/pseudo-selectors/f/focus-visible/).

*Sorry, I can’t send PR since I do not work with Tailwind CSS*

AugustinMauroy:
Close #982

### DIFF
--- a/frontend/components/Nav.tsx
+++ b/frontend/components/Nav.tsx
@@ -51,7 +51,7 @@ export interface NavItemProps {
 export function NavItem(props: NavItemProps) {
   return (
     <a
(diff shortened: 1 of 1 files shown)


crowlKats (approved):
LGTM

Issue #18, opened 2024-02-28
Implement a dark theme


lucacasonato:
Requires https://github.com/denoland/deno_doc/issues/377

chann44:
can i pickup. on this or someone already working on it

andreif:
or at least choose a middle ground so it's not like looking-at-the-sun bright white 🙏

NicestRudeGuy:
Oh man this was the first thing which came in to my mind after opening the JSR home page

csvn:
This feels even more important to me now. My "Dark Reader" extension is screwing up the code blocks, so they become unreadable. So my choice is between having unreadable code or shrugging back like a vampire facing the sun whenever I open JSR 😅

AugustinMauroy:
Also to accomplish that may be add something like storybook to have visula testing

ivands:
+1

PR #985, opened 2025-02-23
feat(ui): Add dark/light theme toggling sitewide

davis9001:
This PR adds dark/light theme to the frontend by adding the island: `<ThemeToggle />` and a script in the `<head>` tag to avoid FOUC (Flash of Unstyled Content).

Closes: #18 

### General TODO list:
- [x] Add `islands/ThemeToggle.tsx`
- [x] Add `<ThemeToggle />` to `components/Header.tsx`
- [x] Add `<script>` in `<head>` to `_app.tsx`
- [x] Save user selected theme in localStorage.theme on click of `<ThemeToggle />` 
- [x] Checked on mobile
- [x] Avoids FOUC
- [x] Add color CSS vars to styles.css
- [x] Add colors to Tailwind config > theme > extend
- [x] Apply colors site wide:
  - [x] /
  - [x] Body background and text
  - [x] Main links (header etc.)
  - [x] Scrollbars
  - [x] Form inputs
  - [x] Selected text
  - [x] Orama search bar
  - [x] Markdown body
  - [x] `pre` and `code` boxes
  - [x] /docs
  - [x] /new
  - [x] /packages/*
  - [x] /account
  - [x] /admin
- [x] Pass GitHub CI tests/checks

image

image

image

image

image

### DIFF
--- a/frontend/islands/ThemeToggle.tsx
+++ b/frontend/islands/ThemeToggle.tsx
@@ -3,7 +3,7 @@
 
 import { useState } from "preact/hooks";
 import { IS_BROWSER } from "fresh/runtime";
-
+import { TbMoon, TbSun } from "@preact-icons/tb";
 /**
  * The head script string to put in the <Head> element.
  * Important for avoiding FOUC (Flash Of Unstyled Content).`
@@ -49,31 +49,8 @@ export default function ThemeToggle() {
       aria-label="Toggle Theme"
     >
       {theme === "light"
-        ? (
-          <svg
-            class="w-6 h-6"
-            viewBox="0 0 20 20"
-            xmlns="http://www.w3.org/2000/svg"
-          >
-            <path d="M17.293 13.293A8 8 0 016.707 2.707a8.001 8.001 0 1010.586 10.586z">
-            </path>
-          </svg>
-        )
-        : (
-          <svg
-            class="w-6 h-6"
-            viewBox="0 0 20 20"
-            xmlns="http://www.w3.org/2000/svg"
-          >
-            <path
-              d="M10 2a1 1 0 011 1v1a1 1 0 11-2 0V3a1 1 0 011-1zm4 8a4 4 0 11-8 0 4 4 0 018 0zm-.464 4.95l.707.707a1 1 0 001.414-1.414l-.707-.707a1 1 0 00-1.414 1.414zm2.12-10.607a1 1 0 010 1.414l-.706.707a1 1 0 11-1.414-1.414l.707-.707a1 1 0 011.414 0zM17 11a1 1 0 100-2h-1a1 1 0 100 2h1zm-7 4a1 1 0 011 1v1a1 1 0 11-2 0v-1a1 1 0 011-1zM5.05 6.464A1 1 0 106.465 5.05l-.708-.707a1 1 0 00-1.414 1.414l.707.707zm1.414 8.486l-.707.707a1 1 0 01-1.414-1.414l.707-.707a1 1 0 011.414 1.414zM4 11a1 1
-                  0 100-2H3a1 1 0 000 2h1z"
-              fill-rule="evenodd"
-              clip-rule="evenodd"
-            >
-            </path>
-          </svg>
-        )}
+        ? <TbMoon class="w-6 h-6" />
+        : <TbSun class="w-6 h-6" />}
     </button>
   );
 }
--- a/frontend/islands/ThemeToggle.tsx
+++ b/frontend/islands/ThemeToggle.tsx
@@ -1,0 +1,79 @@
+#!/usr/bin/env -S deno run -A --watch
+// Copyright 2024 the JSR authors. All rights reserved. MIT license.
+
+import { useState } from "preact/hooks";
+import { IS_BROWSER } from "fresh/runtime";
+
+/**
+ * The head script string to put in the <Head> element.
+ * Important for avoiding FOUC (Flash Of Unstyled Content).`
+ */
+export const themeToggleHeadScript: string = `
+const isDarkMode = localStorage.theme === "dark"
+|| (!("theme" in localStorage)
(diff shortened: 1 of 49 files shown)


AugustinMauroy (changes requested):
you are doing it wrong tailwind have a `dark:` prefix to apply style in dark

frontend/static/markdown.css
```diff
@@ -0,0 +1,799 @@
+/* Override css vars for markdown and code blocks (light and dark). */
```
AugustinMauroy, line 1: wow you have spend a lot of time for that. But it's a bad practice to override something

AugustinMauroy:
also on provided screenshot, the contrast ratio is horrible

frontend/static/markdown.css
davis9001: I don't see anything wrong with overriding CSS vars in this case. In an ideal world I wouldn't do that but seeing as the targets in question are external (and they provide an easy way to update the theme with CSS vars) this was IMHO the best option.

AugustinMauroy: So if it's external let's contribute to the dep (pr already opened by me). You should never override your dep. If you need something from your dep:
- Propose change if it's OSS
- Use an other dep that do the same job
- Write your own implementation

davis9001:
> you are doing it wrong tailwind have a `dark:` prefix to apply style in dark

I use `dark:` prefixes where it makes sense (sparingly) but I feel that it's better to not repeat that everywhere on every element, hence the use of CSS vars.

> also on provided screenshot, the contrast ratio is horrible

I see some room for improvement perhaps but could you be more specific please?

frontend/static/markdown.css
AugustinMauroy: also I just verify and it's already supported by dep
https://github.com/denoland/deno-gfm?tab=readme-ov-file#styling

AugustinMauroy:
> I use dark: prefixes where it makes sense (sparingly) but I feel that it's better to not repeat that everywhere on every element, hence the use of CSS vars.

Your point of view is right. But the project use tailwind. So IMO we should use tailwind approche.

> I see some room for improvement perhaps but could you be more specific please?

on landing page:

- button border
- search bar you should do something like `border-jsr-cyan-600 dark:border-jsr-cyan-400`
- all link

EGAMAGZ:
I like the way you implemented dark theme, it looks beatiful

frontend/islands/ThemeToggle.tsx
```diff
@@ -0,0 +1,79 @@
...
+          >
+            <path
+              d="M10 2a1 1 0 011 1v1a1 1 0 11-2 0V3a1 1 0 011-1zm4 8a4 4 0 11-8 0 4 4 0 018 0zm-.464 4.95l.707.707a1 1 0 001.414-1.414l-.707-.707a1 1 0 00-1.414 1.414zm2.12-10.607a1 1 0 010 1.414l-.706.707a1 1 0 11-1.414-1.414l.707-.707a1 1 0 011.414 0zM17 11a1 1 0 100-2h-1a1 1 0 100 2h1zm-7 4a1 1 0 011 1v1a1 1 0 11-2 0v-1a1 1 0 011-1zM5.05 6.464A1 1 0 106.465 5.05l-.708-.707a1 1 0 00-1.414 1.414l.707.707zm1.414 8.486l-.707.707a1 1 0 01-1.414-1.414l.707-.707a1 1 0 011.414 1.414zM4 11a1 1
+                  0 100-2H3a1 1 0 000 2h1z"
+              fill-rule="evenodd"
+              clip-rule="evenodd"
+            >
+            </path>
+          </svg>
+        )}
```
EGAMAGZ, line 76:
```suggestion
      {theme === "light"
        ? (
          <TbMoon class="w-6 h-6"/>
        )
        : (
          <TbSun class="w-6 h-6" />
        )}
```

frontend/islands/ThemeToggle.tsx
```diff
@@ -0,0 +1,79 @@
+#!/usr/bin/env -S deno run -A --watch
+// Copyright 2024 the JSR authors. All rights reserved. MIT license.
+
+import { useState } from "preact/hooks";
+import { IS_BROWSER } from "fresh/runtime";
+
```
EGAMAGZ, line 6:
```suggestion
import { TbMoon, TbSun } from "@preact-icons/tb";
```

EGAMAGZ (commented):
I suggest the usage if the `@preact-icons/tb` package to keep the icons uniform

frontend/islands/ThemeToggle.tsx
davis9001: agreed thanks!

frontend/islands/ThemeToggle.tsx
davis9001: I merged these changes (with the others) with the VS Code GitHub Pull Request extension hoping it would add you as an author commit but... 🤷 

I added your changes and switched from using `fill` to `text`.

EGAMAGZ: Don't worry if I don't appear in the commit :D

crowlKats:
I just read through this PR; I have to mostly agree with @AugustinMauroy. 
https://github.com/denoland/deno_doc/pull/697 is a pre-requisite for this, and shouldnt just be completely achieved in the jsr repository.
The contrast ratio on some of the pages and elements is fine, but on others it is indeed somewhat dubious.
I would like to mention that #18 is marked as "Needs Plan", and it does also say that https://github.com/denoland/deno_doc/pull/697 is a prerequisite; we probably should make it clearer by making it part of contributing guideline or PR templates that issues marked as "Needs Plan" should not be worked on without prior discussion with maintainers.
I hate to waste effort, especially since it seems a decent amount of work went into this, but i will have to decline this PR.

PR #993, opened 2025-02-27
fix: unifed size for report package button

BorisLord:
Fullscreen view (almost) unchanged
Button was oversized on smaller screen

Before :
image

After :
image

### DIFF
--- a/frontend/routes/package/(_components)/PackageHeader.tsx
+++ b/frontend/routes/package/(_components)/PackageHeader.tsx
@@ -227,10 +227,10 @@ I would like to report a package for the following reason:
 
           <div class="flex flex-row items-center gap-2">
             <a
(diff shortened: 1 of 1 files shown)


AugustinMauroy:
I'm the one who implemented this and I did it voluntarily. Why did you do it? Because it makes the button more accessible on mobile (by increasing its size) because it's finger-operated. And on desktop/large screens, you don't need a huge click zone. That's why it's smaller/discreet.

BorisLord:
I believe the button size is not consistent with the other buttons on the page, such as the "use with" button.
If buttons need to be larger for mobile use, we should review all button sizes to keep the user experience consistent.
In my opinion, the button is too large on mobile, but it's okay if you think a larger button is better for the user experience.

EGAMAGZ (approved):
LGTM

As you also have a good point regarding its dimensions and its relationship to the other buttons.

crowlKats (approved):
sounds reasonable, LGTM

PR #998, opened 2025-03-01
fix: properly error for invalid tarballs

lucacasonato:


piscisaureus (approved):

api/src/tarball.rs
```diff
@@ -686,6 +686,15 @@ impl PublishError {
   }
 }
 
+fn from_tarball_io_error(err: io::Error) -> PublishError {
+  match err.downcast::<reqwest::Error>() {
+    Ok(err) => {
+      return PublishError::GcsDownloadError(GcsError::Reqwest(err));
```
lucacasonato, line 692:
```suggestion
      PublishError::GcsDownloadError(GcsError::Reqwest(err))
```

PR #1004, opened 2025-03-05
refactor: Usage of @std/fmt instead of manual size calculation

EGAMAGZ:
Showcase of @std/fmt usage:

image

### DIFF
--- a/frontend/routes/package/source.tsx
+++ b/frontend/routes/package/source.tsx
@@ -8,6 +8,7 @@ import { PackageHeader } from "./(_components)/PackageHeader.tsx";
 import { TbFolder, TbSourceCode } from "@preact-icons/tb";
 import { ListDisplay } from "../../components/List.tsx";
 import { scopeIAM } from "../../utils/iam.ts";
+import { format as formatBytes } from "@std/fmt/bytes";
 
 export default define.page<typeof handler>(function PackagePage(
   { data, params, state },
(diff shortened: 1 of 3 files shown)


frontend/routes/package/source.tsx
```diff
@@ -8,6 +8,7 @@ import { PackageHeader } from "./(_components)/PackageHeader.tsx";
 import { TbFolder, TbSourceCode } from "@preact-icons/tb";
 import { ListDisplay } from "../../components/List.tsx";
 import { scopeIAM } from "../../utils/iam.ts";
+import { format as formatBytes } from "@std/fmt/bytes";
```
AugustinMauroy, line 11:
```suggestion
import { format } from "@std/fmt/bytes";
```

I'd tend to use the name provided by the lib, at least everyone has the same name whatever the project.

AugustinMauroy (approved):
LGTM !

frontend/routes/package/source.tsx
EGAMAGZ: Yep, you're right but to be more descriptive and don't use a generic/common name pretty used that might cause confusion to the LSP I rather used an alias

AugustinMauroy: In this case I would be in favor of using lib name. But open an issue on `denoland/std` (not sure of the repo)

EGAMAGZ: No, is not needed ;-;. What I meant is that as it is a generic name, is common to be used, so there might more that one function named `format` from the project or a library.

crowlKats (approved):
LGTM

Issue #899, opened 2025-01-21
ability to delete of published version under certain circumstances


crowlKats:
We need a way to delete a package version, however we should keep this to be under very specific circumstances. one idea would be:
- less than X downloads
- published less than X ago
- No dependents

this should be sufficient to enable this capability for the usecase people have, while still being true to the rule of being immutable.

Inbestigator:
Maybe if not latest version? (Unless only version)

BlackAsLight:
To add to this, I think it would be beneficial to exclude these versions when downloading an existing lib and looking for libs to deduplicate but still included if adding a new dependency to a project.

Inbestigator:
Maybe yanking could actually be more effective (like yanking all versions will allow you to delete the project etc.)

crowlKats:
no, yanking allows versions to still be available, and deleting could break a lot. ie lets say std/collections yanked all versions and then deleted the module, that would completely destroy everything

yuhr:
Versions that satisfy the circumstances must be shown with a noticeable BIG RED WARNING label about the endangerment in the Web UI. Otherwise random users might experience broken builds unexpectedly (like me experienced the same in npmjs.com).

yuhr:
I mean, we can't measure number of dependents at all, because not all actual dependents are published on JSR. Internal use only softwares might depend on such a low-reputation version.

On the other hand, number of downloads would be trustable. Thresholding at ~10 downloads sounds completely viable to me.

NetOpWibby:
I would like to delete a scope but I'm prevented from doing so because I have a module with releases. Those releases have been yanked, no one has installed them, but I cannot delete them...I feel that I should be able to delete the package and scope.

James-Ansley:
Would it not be possible to have a kind of package retirement plan?

For example, if all versions of a package have been yanked for some number of years, say three or five years, and downloads/use falls below a certain threshold after that time, a package can be deleted by the owner? Or even having "marking for deletion" be a separate step on top of yanking all versions which initiates a cool-down period before the package can be deleted? That at least seems like a vaguely reasonable trade off between not having packages pulled out from under anyone – giving everyone ample warning about a package's future, while also giving package owners some autonomy over their packages.

PR #1011, opened 2025-03-10
feat(admin): allow deleting package versions

crowlKats:
this PR adds an endpoint to be able to delete package versions, but only for admins. This is not exposed in the UI for admins either, as the fact that this is for admins only is just a temporary restriction, and the idea is to come up with sensible restrictions that apply to users so they can delete versions (see #899).


api/src/api/package.rs
```diff
@@ -942,8 +946,98 @@ pub async fn version_update_handler(
...
+  let count = db
+    .count_package_dependents(
+      crate::db::DependencyKind::Jsr,
+      &format!("@{}/{}", scope, package),
+    )
+    .await?;
+
+  if count > 0 {
+    return Err(ApiError::DeleteVersionHasDependents);
+  }
```
crowlKats, line 1009: this check is not correct, as it doesnt use the version as a requirement, and as such this check applies to all versions and is stricter than should be. This is good enough for now, however ideally this would build a graph and check if there are any other versions that fit the constraints used by dependents, and only if there is no such constraint would it error.

ry (commented):
Add admin UI for deleting packages

api/src/db/database.rs
```diff
@@ -1649,6 +1649,26 @@ impl Database {
     .await
   }
 
+  #[instrument(name = "Database::yank_package_version", skip(self), err)]
+  pub async fn delete_package_version(
```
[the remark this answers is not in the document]
ry: i think you should log or instrument the package name here

ry (approved):
RSLGTM if you feel confident

api/src/db/database.rs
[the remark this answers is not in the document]
crowlKats: thats already done in the handler function that calls this (same setup we have for all other db calls & endpoints)

lucacasonato (commented):
@crowlKats We had discussed that users should not be able to publish a version that was deleted anew - you didn't implement that yet. Can you do that?

frontend/routes/package/versions.tsx
```diff
@@ -254,6 +260,19 @@ function Version({
             </button>
           </form>
         )}
+        {isPublished && iam.isStaff && (
```
lucacasonato, line 263: Does this only show up when sudo is on?

crowlKats: good catch, fixed in https://github.com/jsr-io/jsr/pull/1027

crowlKats:
@lucacasonato this is already handled. opened https://github.com/jsr-io/jsr/pull/1027 for adding an additional test to showcase this.

PR #1012, opened 2025-03-11
fix: add nullable:true to rekorLogId

### LINKED ISSUE #885: Bug: openapi schema, `rekorLogId` should be nullable
I am using the openapi schema (https://api.jsr.io/.well-known/openapi) to generate client.
And I noticed an error in the schema 

Here is the schema for `rekorLogId`
```yaml
rekorLogId:
          type: string
          description: Rekor log ID for the published package version.
```

But when a package isn't signed by github action, its `rekorLogId` is null.
I believe a `nullable: true` should be added to the schema.

Jayanth-Parthsarathy:
# Add Github Repository property to Packages
Fixes #885

- [x] Added nullable:true to rekorLogId

### DIFF
--- a/api/src/api.yml
+++ b/api/src/api.yml
@@ -1883,6 +1883,7 @@ components:
           description: The date and time when the package version was last updated.
         rekorLogId:
           type: string
+          nullable: true
           description: Rekor log ID for the published package version.
       required:
         - scope


crowlKats (approved):
LGTM

Issue #110, opened 2024-02-29
Request for Mirror Synchronization Capability


elrrrrrrr:
Dear jsr maintainers,

Greetings from the [cnpm](https://github.com/cnpm) team, the provider of stable and free npm registry package mirror services in China, including the frontend at https://npmmirror.com/ and the registry at https://registry.npmmirror.com/.

After testing, we have found that we cannot directly access the npm.jsr.io content within mainland China. In order to provide better services in collaboration with jsr in the future, we would like to request mirror synchronization capabilities similar to those offered by npm registry.

We rely on the changesStream interface provided by npm registry to deliver streaming version change information at https://replicate.npmjs.com/_changes.

Additionally, for frontend display purposes, we also hope to provide manifest interfaces to showcase metadata information.

Looking forward to your response.

elrrrrrrr:
@lucacasonato

🤩

Here is the [REPLICATE-API](https://github.com/npm/registry/blob/master/docs/REPLICATE-API.md#overview) interface, which returns corresponding changes through a stream. 

https://registry.npmmirror.com/_changes?since=1, we have **also** developed our own changesStream interface, which returns changes in JSON format, It also requires a since parameter for pagination.

Additionally, we noticed that the manifest does not contain data related to maintainers, which might introduce some differences. If there are already existing interfaces for updates, please inform us how to call them.

Lastly, we hope to find a way to return all npm tarball list for initialization purposes. such as [ref](https://github.com/cnpm/cnpmcore/blob/sync-all-packages/test/SyncAllPackages.ts#L1) .

elrrrrrrr:
We've been attempting some synchronization and adaptation work locally and have successfully completed it for a single package, happ pass. 
![image](https://github.com/jsr-io/jsr/assets/5574625/37d43c8a-a418-435d-852d-e1d3738cda74)


Here are some of the services or changes we urgently need:
1. `Maintainers` in manifests,  (all packages in the npm registry have maintainers information even security holder). Are there any plans to add a default value for this field, maybe like `jsr@jsr.io`?
2. There is a need for a way to export the names of all existing packages for init the mirror.
3. We need a changesStream service. Currently, it seems  `PACKAGE_VERSION_ADD` event is enough ?

ryoppippi:
@elrrrrrrr @lucacasonato 
Hi! Any updates?
I just got an issue from a developer in China that JSR is slow.

PR #1025, opened 2025-04-02
feat: add minimal implementation of changesStream for registry mirroring

elrrrrrrr:
Hi, jsr team 

This PR introduces a minimal implementation of the changesStream interface (#110) to enable registry mirroring capabilities. Given that `11 months` have passed since our initial discussion and the feature remains unscheduled, I've taken the liberty to propose this implementation based on our proven approach in npmmirror.

## Key implementation notes:
1. Implements simple polling mechanism via `application/json` endpoint
2. Follows npmmirror pattern (compared to npm's heavier stream interface)
3. Maintains backward compatibility with existing infrastructure

## For maintainers:
- This implementation prioritizes simplicity and maintainability
- Being new to Rust, I welcome guidance on code quality and best practices 😋

**Context**:
As developers in mainland China experience consistent network challenges accessing the JSR registry directly, this feature would significantly improve developer experience for one of the world's largest developer communities.

Thanks for building this vital infrastructure for the JavaScript ecosystem. 
Looking forward to your feedback! 🙏🏻

### DIFF
--- a/api/migrations/20250402143550_create_changes.sql
+++ b/api/migrations/20250402143550_create_changes.sql
@@ -1,0 +1,16 @@
+CREATE TYPE change_type AS ENUM (
+    'PACKAGE_VERSION_ADDED',
+    'PACKAGE_TAG_ADDED'
+);
+
+CREATE TABLE changes (
+    seq BIGSERIAL PRIMARY KEY,
+    change_type change_type NOT NULL,
+    package_id VARCHAR(255) NOT NULL,
+    data TEXT NOT NULL,
+    created_at TIMESTAMPTZ NOT NULL DEFAULT NOW()
+);
+
+-- 创建索引
+CREATE INDEX changes_package_id_idx ON changes (package_id);
+CREATE INDEX changes_created_at_idx ON changes (created_at);
--- a/api/migrations/20250402143550_create_changes.sql
+++ b/api/migrations/20250402143550_create_changes.sql
@@ -11,6 +11,5 @@ CREATE TABLE changes (
     created_at TIMESTAMPTZ NOT NULL DEFAULT NOW()
 );
 
--- 创建索引
 CREATE INDEX changes_package_id_idx ON changes (package_id);
 CREATE INDEX changes_created_at_idx ON changes (created_at);
--- a/api/src/api/changes.rs
+++ b/api/src/api/changes.rs
@@ -1,12 +1,10 @@
-use crate::api::ApiError;
 use hyper::{Body, Request};
-use routerify::Router;
 use routerify::prelude::*;
 use serde::Serialize;
 
 use crate::{
     db::{Change, Database},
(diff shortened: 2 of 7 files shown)


api/migrations/20250402143550_create_changes.sql
```diff
@@ -0,0 +1,15 @@
+CREATE TYPE change_type AS ENUM (
+    'PACKAGE_VERSION_ADDED',
+    'PACKAGE_TAG_ADDED'
+);
+
+CREATE TABLE changes (
+    seq BIGSERIAL PRIMARY KEY,
+    change_type change_type NOT NULL,
+    package_id VARCHAR(255) NOT NULL,
```
lucacasonato, line 9: Seperate `scope_name` and `package_name`.

lucacasonato (commented):
Does the changes endpoint need to return changes in order?

If so, this doesn't work. There is no guaruantee in Postgres that a sequential table is actually sequential on insert because multiple transactions can be in flight at once, and sequential IDs are assigned at the start of a transaction, not once the transaction commits.

I suggest you change this as follows:

- Instead of a seperate database query to insert changes into the change stream, do this using `TRIGGER` in postgres on the `package_versions` table. Add a trigger that on insert creates the `changes` entry.

- Add a `row_version` column to the `changes` table that is typed as `xid8 NOT NULL DEFAULT pg_current_xact_id()`.

- On query, limit the returned results to `WHERE row_id <= pg_snapshot_xmin(pg_current_snapshot())`.

api/migrations/20250402143550_create_changes.sql
elrrrrrrr: updated 😄

fengmk2:
Is there any progress?

elrrrrrrr:
> Does the changes endpoint need to return changes in order?
> 
> If so, this doesn't work. There is no guaruantee in Postgres that a sequential table is actually sequential on insert because multiple transactions can be in flight at once, and sequential IDs are assigned at the start of a transaction, not once the transaction commits.
> 
> I suggest you change this as follows:
> 
> * Instead of a seperate database query to insert changes into the change stream, do this using `TRIGGER` in postgres on the `package_versions` table. Add a trigger that on insert creates the `changes` entry.
> * Add a `row_version` column to the `changes` table that is typed as `xid8 NOT NULL DEFAULT pg_current_xact_id()`.
> * On query, limit the returned results to `WHERE row_id <= pg_snapshot_xmin(pg_current_snapshot())`.
> 
> What this will do is ensure that we only show changes where all previous transactions have already committed or failed.

Hello @lucacasonato

We still hope to have the changesStream capability. 
I noticed that jsr has higher requirements for time sequences.

I wanted to confirm if there are already any other solutions or plans scheduled in jsr regarding this feature.
If not, I will proceed to modify the PR according to the suggestions provided.

Looking forward to your reply 🙏🏻

PR #1035, opened 2025-04-15
fix: support `virtual:` and `cloudflare:` schemes

### LINKED ISSUE #986: Supporting `virtual:*` modules?
Hi,

[Vite](https://vite.dev) offers a plugin construct where you can define "virtual modules". ( `import packageName from 'virtual:{package-name}'`)
In short, when bundling you can refer to this "module" and inject code into the bundle.
This can be very useful for including build-time generated framework metadata for example.

Currently, when you try to compile a JSR package using this construct you get an error:
```txt
error: Unsupported scheme "virtual" for module "virtual:{package-name}". Supported schemes:
 - "blob"
 - "data"
 - "file"
 - "http"
 - "https"
 - "jsr"
 - "npm"
    at file:///home/runner/work/file.mts:3:27
Child process exited with: 1
 ELIFECYCLE  Command failed with exit code 1.
Error: Process completed with exit code 1.
```

I'm not sure if it's in JSR's interest to support this, because it seems to check whether your dependencies actually resolve.
(I tried replacing it with `npm:virtual:{package-name}` and got an error because of that.
But I figured I'd make an issue anyway so it's documented if JSR doesn't want to support it.

### LINKED ISSUE #1034: Add the ability to ignore schemes
There are situations in which a platform will introduce an import scheme that aren't supported by deno. These schemes should still be publishable to JSR. 

The biggest one right now that I'm aware of is cloudflare's `cloudflare:*` scheme. For example, if you're creating a durable object:

```ts
import { DurableObject } from "cloudflare:workers"
```

Likewise with importing `env`

```ts
import { env } from "cloudflare:workers"
```

Right now packages that reference these imports _can't_ be published to JSR even though JSR ostensibly supports cloudflare as a suggested deployment target. 

---

See also #986

crowlKats:
Closes #986
Closes #1034

@zephraph this should address your problem, alongside https://github.com/denoland/deno/pull/28922


zephraph:
Thanks @crowlKats!

ry (approved):

PR #1051, opened 2025-04-18
fix(admin): align notifcation on ticket list, wrap assign scope button, make failure case of tasks red

AugustinMauroy:
- add color to differentiate error/success
- remove layout shift when there are unread notifications 

iPhone-14-Plus-jsr test
Macbook-Air-jsr test

### DIFF
--- a/frontend/routes/admin/tickets.tsx
+++ b/frontend/routes/admin/tickets.tsx
@@ -12,9 +12,6 @@ export default define.page<typeof handler>(function Tickets({
   data,
   url,
 }) {
-  const isNotification = (ticket: Ticket) =>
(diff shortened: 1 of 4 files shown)


frontend/routes/admin/tickets.tsx
```diff
@@ -12,6 +12,9 @@ export default define.page<typeof handler>(function Tickets({
   data,
   url,
 }) {
+  const isNotification = (ticket: Ticket) =>
```
crowlKats, line 15: instead of this function, create a variable in the tickets.map call

AugustinMauroy: good idea

crowlKats (approved):
LGTM

PR #1052, opened 2025-04-18
feat(PackageHeader): min height + outline button

### LINKED ISSUE #386: package headers should have a minimum height so nav doesn't move
Screenshot 2024-04-16 at 9 54 27 AM
Screenshot 2024-04-16 at 9 54 12 AM

If you're browsing a package with a short or non-existent description clicking a tab that isn't tied to a specific package version, like the score tab, will cause the UI to shift around since the "published" content block is no longer displaying. The overall header should probably have a minimum height set to whatever the shortest possible header with a published date is.

AugustinMauroy:
Close #386

And also add a lighter version of button for reporting (issue was reported on JSR's discord)

iPhone-13-PRO-jsr test

### DIFF
--- a/frontend/islands/TicketModal.tsx
+++ b/frontend/islands/TicketModal.tsx
@@ -19,7 +19,7 @@ export function TicketModal(
   { user, kind, style, fields, children, extraMeta, title, description }: {
     children: ComponentChildren;
     kind: TicketKind;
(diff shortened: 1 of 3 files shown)


AugustinMauroy:
during dark theme or ticket changes header as change not anymore an issue

PR #1053, opened 2025-04-18
fix(input): safari contraint

### LINKED ISSUE #923: "invite members" section doesn't have good UX
Image

The select component isn't great. IMO it's to sticky from input.

- First soltion add gap between select and input
- Second, remove border-radius on right side of the select

AugustinMauroy:
close #923


crowlKats:
I dont think this fix is acceptable; its not clear anymore that its a dropdown

crowlKats:
@AugustinMauroy can you add an up-to-date screenshot?

crowlKats (approved):

PR #1057, opened 2025-04-21
chore: remove alias infrastructure

crowlKats:
removes the alias infrastructure that we never ended up using


lucacasonato (approved):
LGTM

PR #1060, opened 2025-04-22
fix(frontend): fix data point aggregation logic for download chart

kt3k:
This PR fixes the rendering of download chart when the recent 5 versions don't have enough x values in the chart.

This PR fixes the download chart of `@ecopages/lit` package:

BEFORE
Screenshot 2025-04-22 at 11 49 45

AFTER
Screenshot 2025-04-22 at 11 49 31

### DIFF
--- a/frontend/routes/package/(_islands)/DownloadChart.tsx
+++ b/frontend/routes/package/(_islands)/DownloadChart.tsx
@@ -5,6 +5,7 @@ import type {
   DownloadDataPoint,
   PackageDownloadsRecentVersion,
 } from "../../../utils/api_types.ts";
+import type ApexCharts from "apexcharts";
 
 interface Props {
   downloads: PackageDownloadsRecentVersion[];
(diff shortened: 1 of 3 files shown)


frontend/routes/package/(_islands)/DownloadChart.tsx
```diff
@@ -158,7 +158,7 @@ export function normalize(
   dataPoints: DownloadDataPoint[],
   xValues: string[],
   aggregationPeriod: AggregationPeriod,
-): [Date, number][] {
+): [number, number][] {
```
kt3k, line 161: Looks like this needs to be number according to the type definition

frontend/routes/package/(_islands)/DownloadChart.tsx
```diff
@@ -193,7 +193,7 @@ function getSeries(
   ).flat();
 
   const xValues = collectX(
-    dataPointsToDisplay.map((version) => version.downloads).flat(),
+    dataPointsWithDownloads.map((version) => version.downloads).flat(),
```
kt3k, line 196: Only this line is the actual fix of the issue. We need to collect x values based on 'all' data, instead of recent '5' versions.

crowlKats (approved):
LGTM!

PR #1065, opened 2025-04-23
fix: slowness in download endpoint

lucacasonato:
1. It was meant to limit to most recent 5 versions, but it was querying all.
2. It was doing a bunch of queries in series instead of querying all versions in a single call.


crowlKats (approved):
LGTM

PR #1055, opened 2025-04-20
feat: dark mode

crowlKats:
Closes #18

PR #1069, opened 2025-04-24
fix: correct default dark-mode preference

csvn:
Thanks for shipping Dark mode for JSR! (#1055) 🎉 

Though I noticed that the media `prefers-color-scheme` detection is not working correctly. During loading the correct theme will be applied (via `_app.tsx` script), but when Preact renders it will override with incorrect theme. This is because the `isDarkStored ?? isDarkPreference` expression basically always returning `isDarkStored` (`isDarkStored` is `boolean`, not `boolean | undefined | null`).

This PR attempts to fix the issue by ensuring the [logic is the same as for `_app.tsx`](https://github.com/jsr-io/jsr/blob/main/frontend/routes/_app.tsx#L74-L81) for when Dark mode should be applied, which avoids dark -> light flashing for when dark mode is preferred without any stored value.

## Current behavior

This is how it looks for me (3G throttled) when I have no `darkMode` localeStorage item.

https://github.com/user-attachments/assets/56ff105f-5a39-46ff-bf6c-40f518366ed2

### DIFF
--- a/frontend/islands/DarkModeToggle.tsx
+++ b/frontend/islands/DarkModeToggle.tsx
@@ -6,10 +6,11 @@ export default function DarkModeToggle() {
   const [isDark, setIsDark] = useState(false);
 
   useEffect(() => {
-    const isDarkStored = localStorage.getItem("darkMode") === "true";
+    const isDarkStored = localStorage.getItem("darkMode");
     const isDarkPreference =
       globalThis.matchMedia("(prefers-color-scheme: dark)").matches;
-    const initialDarkMode = isDarkStored ?? isDarkPreference;
+    const initialDarkMode = isDarkStored === "true" ||
+      isDarkStored === null && isDarkPreference;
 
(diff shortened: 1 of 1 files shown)


csvn:
FYI: I also had issues on windows with the `cloudtrace:v2.ts` file, as it seems it's [semicolon may not be a valid character in file names](https://stackoverflow.com/a/31976060). So if possible, it would be nice for Windows contributors to change this file name if possible, e.g. `cloudtrace.v2.ts` or something.

image

crowlKats:
@csvn yep seems good to change that name; was just named as such as that was the name of it when we vendored it, but feel free to rename it

crowlKats (approved):
LGTM

csvn:
> @csvn yep seems good to change that name; was just named as such as that was the name of it when we vendored it, but feel free to rename it

Cool, I'll see if I can make a separate PR for that, to make contributions easier 😃

PR #1075, opened 2025-04-25
fix(ui/ux): reduce spacing

### LINKED ISSUE #1073: Weird spacing on lib page
Image

This may just be personal preference, but that is way too much space between the header and the navbar.

Here's a more space efficient layout I propose:

Image

Although, flattening it into one line and maybe moving the report button further down would probably be better, akin to NPM's layout:

Image

AugustinMauroy:
close #1073

I think this spacing was added for report button which is isn't there anymore


AugustinMauroy:
Humm
here I've removed the superfluous space. But I don't think removing more spacing will be a good idea for UX/A11Y

PR #1077, opened 2025-04-26
refactor: Usage of badge-maker to generate badges

### LINKED ISSUE #1044: Use badge-maker package instead of proxying shields.io
Since we currently provide the feature of customizing logo color, I think we can consider moving to badge-maker after https://github.com/badges/shields/pull/10955. It should be easier to change the logo color than base64 images.

EGAMAGZ:
Fixes #1044


crowlKats:
@EGAMAGZ whats the status of this PR?

PR #1085, opened 2025-04-29
refactor: Added spacing between docs and report package

EGAMAGZ:
The margin given is the same one in Overview to keep constant the gap

Before:
imagen

Now:
imagen


crowlKats (approved):
lgtm

PR #1093, opened 2025-05-02
feat(scope): add description

### LINKED ISSUE #217: Add the ability to add a description field to a scope
On the scope page it should render the description in markdown.

It might also be nice to have it for user profiles.

AugustinMauroy:
close #217


crowlKats (changes requested):
instead of it being nullable, always only support a string, and check for empty string instead as the description not being set (just like we do for package description).

api/src/api.yml
AugustinMauroy, line 1: TODO(@AugustinMauroy): update this when pr is cleaned

api/src/api/scope.rs
```diff
@@ -163,6 +165,11 @@ async fn update_handler(
       )
       .await?
     }
+    ApiUpdateScopeRequest::Description(description) => {
```
AugustinMauroy, line 168: DO we need that ?

AugustinMauroy:
bump @crowlKats now it's should be good !

crowlKats (approved):
LGTM, thanks!

PR #1107, opened 2025-05-13
refactor(frontend): use Preact Signals

iuioiua:
`useSignal()` is more ergonomic and performant than `useState()`.


crowlKats (approved):
LGTM

PR #1116, opened 2025-05-16
feat: scope-level symbol search

crowlKats:


lucacasonato (approved):
LGTM - let's serve the css and such from Fresh though as discussed offline.

crowlbot (commented):
Thanks for this — scope-level symbol search is a nice addition, and moving the ddoc CSS/JS out of every docs API response into static assets is a good cleanup that also addresses the earlier "serve the CSS from Fresh" note. I did a thorough pass; a few things look like genuine bugs, plus some concerns on the new scope endpoint.

Heads up first: the branch currently conflicts with `main` and the `test` job is red, so it'll need a rebase + green CI before it can land.

### Bugs

**1. `<script>` uses `href` instead of `src` — the ddoc script never loads.**
In `frontend/routes/@[scope]/index.tsx`, `frontend/routes/package/(_components)/Docs.tsx`, and `frontend/routes/package/source.tsx`:

```jsx
<script hidden href={asset("/ddoc/script.js")} />
```

`<script>` loads from `src`, not `href`; the browser ignores `href` here, so `script.js` is never fetched or executed. That regresses the interactive bits that used to be inlined via `dangerouslySetInnerHTML={{ __html: docs.script }}` (copy buttons, the usage-selector `mouseup` handler). Should be `src={asset("/ddoc/script.js")}`.

**2. `NavOverflow` outside-click handler throws on every document click.**
In `frontend/components/NavOverflow.tsx`:

```js
function outsideClick(e) {
  if (!navOverflowMenuEl.parent.contains(e.target)) {
```

`Element` has no `.parent` property (it's `parentElement`/`parentNode`), so `navOverflowMenuEl.parent` is `undefined` and `.contains(...)` throws a `TypeError` on every click on the document. You already compute `const navOverflowButton = navOverflowMenuEl.parentElement;` above — that's likely what you want: `!navOverflowButton.contains(e.target)`.

**3. Leftover debug `console.log` in the nav script.**
Same file, inside the click handler:

```js
while (el && el !== e.currentTarget) {
  console.log(el.tagName);
  ...
```

Logs on every click-bubble step; should be removed.

### Concerns on `get_docs_search_html_handler` (`api/src/api/scope.rs`)

**4. Silent 100-package cap.** `list_packages_by_scope(&scope, false, 0, 100)` only ever covers the first 100 packages, so symbol search in a scope with >100 packages silently misses the rest. Worth either paginating through all of them or making the limit explicit.

**5. One missing-docs package 500s the whole scope search.** Inside the loop, a missing GCS docs object maps to `ApiError::InternalServerError`, failing the entire request — so a single package whose docs haven't been generated takes down search for the whole scope. Probably want to `continue` + log here, like the `None`-version case just above it.

**6. Sequential, unbounded work per request.** For up to 100 packages this does, serially: a `get_package` round-trip, a `get_latest_unyanked_version` round-trip, a GCS download, a JSON parse, and full all-symbols HTML generation — then concatenates all of it into one string. On a cache miss (the cache is only 1 minute) that's a lot of latency and memory for a large scope; consider parallelizing the downloads and/or bounding the response. Minor: you already have `package` from `list_packages_by_scope`, but re-fetch it with `get_package` (for `repo`) — two queries per package.

**7. `getElementById(doc.target_id.toLowerCase())!` can be null.** In both `ScopeSymbolSearch` and the cloud path of `PackageSymbolSearch`, if the Orama index and the freshly-rendered HTML ever diverge (symbol removed in a newer publish, async index lag, etc.), `getElementById` returns `null` and the next `.parentElement` throws, killing the whole results loop so nothing renders. A null check that skips the hit would be more robust.

### Build hygiene

**8. `build.rs` writes into the frontend source tree, and the output is committed + gated by `git diff --exit-code`.** `api/build.rs` writes `../frontend/static/ddoc/{style.css,comrak.css,script.js}` on every `cargo build`, those files are checked in, and CI fails if they drift. It works, but (a) a Rust build mutating a sibling `frontend/` directory as a side effect is surprising, and (b) the committed assets are now coupled to the pinned `deno_doc = "=0.183.0"` build-dependency — bumping deno_doc means regenerating and re-committing them or CI goes red. Might be worth a note in `build.rs`/CONTRIBUTING about that, or generating to `OUT_DIR` and serving the assets from the API rather than committing them.

### Nits
- `@oramacloud/client` is bumped `^1 → ^2` (major) — just confirming the `OramaClient` constructor + `.search({ term, where, mode, limit })` surface is v2-correct.
- The `NAV_OVERFLOW_SCRIPT` template string mixes tabs and spaces in a few spots (`deno fmt` won't touch string contents).
- `api/Cargo.toml` is missing a trailing newline.
- `hidden` on `<link rel="stylesheet">` is a no-op (harmless).

Happy to re-review once it's rebased and CI is green.

PR #1127, opened 2025-05-30
chore: add dependabot

AugustinMauroy:
### Description

Adding dependabot to track update on our dependency. Sadly dependabot didn't track changes on deno

Action is weekly check to unsure there are no vulnerabilities on supply chain. But rest in montly do reduce noice and maintenance.

### PR Checklist

- [X] The PR title follows

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,21 @@
+version: 2
+updates:
+  - package-ecosystem: github-actions
+    directory: /
+    schedule:
+      interval: weekly
+
+  - package-ecosystem: cargo
+    directory: /
+    schedule:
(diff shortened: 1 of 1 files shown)


AugustinMauroy:
Do you know of an alternative? Or should we just keep it for GHAs?

crowlKats:
i think we should just not have anything of the sort

AugustinMauroy:
I disagree. A tool that avoids having to check deposits is crucial for project maintenance.

crowlKats:
I'll close this for now, as this just doesnt match what we do across other projects and the infrastructure for this just isnt available yet in the rust ecosystem, but I am more than happy to discuss again

PR #1139, opened 2025-06-10
chore: update vlt logo

ejkorol:
Updates the vlt logo with new branding icon. Adjusts icon styles for better visual consistency with the existing icon list.


crowlKats (approved):
thanks!

PR #1161, opened 2025-07-17
feat: store licenses

### LINKED ISSUE #44: Store licenses for packages
Each package / package version should have a `license` field. This field should be a SPDX ID of a license.

This license should be visible via the API and UI, and could be useful for auditing purposes.

We should sniff the LICENSE file (like GitHub does) to populate this field. There should be a fallback where you can explicitly set the license SPDX ID via either a field in the package settings, or via a directive in the license file.

crowlKats:
closes #44


api/src/tarball.rs
```diff
@@ -273,6 +283,49 @@ pub async fn process_tarball(
...
+      break;
+    }
+  }
+
+  if license.is_none() {
+    if let Some(license) = &config_file.license {
+      if store.get_original(&license).is_none() {
+        return Err(PublishError::InvalidLicense);
+      }
+    }
```
lucacasonato, line 320: I think this should take priority over the file.

PR #1170, opened 2025-08-11
docs: Update publishing-packages.md

fry69:
Add small hint to run the `publish.yml` workflow only when a file containing the `version` changes, not on every commit to `main`.


KnorpelSenf:
I'm not sure I agree that this should be merged. I think the docs should target people who have no idea how to use JSR (otherwise they wouldn't be reading the docs, eh?) and for this audience, it's great to explain things in the simplest possible way. 

Reducing the number of CI runs a bit is something that people can do later on, so I'd avoid the complexity at this step. 

Just my 2c

PR #1218, opened 2025-10-24
feat: Allow dots in organization/scope names

philkunz:
Updates validation rules to allow dots (.) in organization names while maintaining backward compatibility with existing names.

Backend changes:
- Allow dots in ScopeName validation (api/src/ids.rs:30)
- Prevent leading/trailing dots (e.g., .org, test.)
- Prevent consecutive dots (e.g., foo..bar)
- Add new error variants for dot validation
- Add comprehensive test coverage for dot patterns

Frontend changes:
- Update regex to allow dots (frontend/utils/ids.ts:9)
- Remove underscore support to match backend validation
- Add validation for leading/trailing dots
- Add validation for consecutive dots
- Update error messages for consistency

Valid examples: my.org, foo.bar, foo.bar.baz, org.name-123
Invalid examples: .foo, foo., foo..bar

## Reasoning
Allow Open Source projects to mirror their project domain as org name.
 
### PR Checklist

- [x] The PR title follows
- [x] Is this closing an open issue? If so, link it, else include a proper
      description of the changes and rason behind them.
- [x] Does the PR have changes to the frontend? If so, include screenshots or a
      recording of the changes.
      <br/>If it affect colors, please include screenshots/recording in both
      light and dark mode.
- [x] Does the PR have changes to the backend? If so, make sure tests are added.

### DIFF
--- a/api/src/ids.rs
+++ b/api/src/ids.rs
@@ -27,7 +27,9 @@ impl ScopeName {
     if !name
       .chars()
       // temp allow underscores
-      .all(|c| c.is_ascii_lowercase() || c.is_ascii_digit() || c == '-' || c == '.')
+      .all(|c| {
+        c.is_ascii_lowercase() || c.is_ascii_digit() || c == '-' || c == '.'
+      })
     {
       return Err(ScopeNameValidateError::InvalidCharacters);
     }


Ionaru:
While I don't think allowing dots in scope names is wrong in principle, it provides some challenges beyond just widening the allowed characters on the jsr website and backend.

@philkunz Maybe discuss this with the JSR maintainers during their open office meeting: https://discord.gg/V3z7A25S?event=1388890175240667197

These are the problems I can think of:

---

On Windows machines, it is problematic to install packages using a dot in their name.
image
And the package name will need to be wrapped in quotes like this: `npx jsr add "@foo.example/bar"`.
This will cause the default copy-able command in jsr.io to fail.
image psd

I am not sure if wrapping the package name in quotes has effects on other operating systems

---

The jsr CLI tool will also need to be modified, as it does its own check for valid scope/package names here: https://github.com/jsr-io/jsr-npm/blob/main/src/utils.ts

image


(Maybe `deno add` and other package managers too)

philkunz:
@KnorpelSenf So... Have I convinced you and you throw your support behind this one?

philkunz:
@KnorpelSenf Cool :) Can you review the change?

Issue #1219, opened 2025-10-26
CopyButton returns undefined when copying workflow code on publish page


katayama8000:
## Description

On the package publish instructions page, clicking the copy button for the GitHub Actions workflow code results in `undefined` being copied to the clipboard instead of the actual YAML code.

## Steps to Reproduce

1. Navigate to a package's publish page (e.g., `/@scope/package/publish`)
2. Scroll to the "Publish from CI" section
3. Click the copy button (📋) next to the workflow code block
4. Paste the clipboard contents
5. Expected: Clean YAML workflow code
6. Actual: `undefined`

PR #1221, opened 2025-10-31
fix: (potential) workaround for transpiler's error

kt3k:
Currently the below part:

```js
const { text } = props;
...
navigator.clipboard.writeText(text)
```

is transpiled like the below (`props.value` is used instead of `props.text`)
Screenshot 2025-10-31 at 10 46 46

and it causes copying of 'undefined' text (ref https://x.com/JLarky/status/1984050194896580874 )

This PR tries to fix it by avoiding using destructuring syntax.


crowlKats (approved):

BlackAsLight:
Closed: https://github.com/jsr-io/jsr/issues/1219

PR #1243, opened 2025-12-26
feat: webhooks

### LINKED ISSUE #1030: Add change feed / webhook API
@coreybutler 
https://x.com/goldglovecb/status/1909638974802387128

crowlKats:
Closes #1030

Needs to setup a job to delete old (7 days) deliveries


KnorpelSenf (changes requested):
Thank you for working on this!

Managing webhooks for a scope is still broken for me, see the comment below.

I think we should have some details on how the payload objects are structured if JSON is selected. This can either be a short example in the settings page, but either way we should likely have a page in the docs. The webhook UI should then link to this page.

api/src/api/package.rs
```diff
@@ -191,6 +195,34 @@ pub fn package_router() -> Router<Body, ApiError> {
...
+      util::auth(util::json(create_webhook_handler)),
+    )
+    .get(
+      " /:package/webhooks/:webhook",
+      util::auth(util::json(get_webhook_handler)),
+    )
+    .get(
+      "/:package/webhooks",
+      util::auth(util::json(list_webhooks_handler)),
+    )
```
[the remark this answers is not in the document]
crowlKats: not sure i understand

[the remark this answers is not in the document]
KnorpelSenf: When you view the scope settings, there's a 400 and the settings page can't be displayed. This is caused because it tries to fetch this endpoint, which then panics because there is no package name. Instead, it should fetch the webhook settings for the scope (a different endpoint).

KnorpelSenf:
Absolutely! I didn't think it could live in this repo (so I didn't think of including it in this PR) but that makes perfect sense!

KnorpelSenf (approved):
LGTM

crowlbot (commented):
This is a big one — the overall shape (event/endpoint/delivery split, Cloud Tasks queue with 3 retries, AES-256-GCM at-rest encryption of secrets, fan-out via DB INSERT-SELECT, HMAC-SHA256 signature for JSON payloads) is right and the schema design is clean. KnorpelSenf approved on 2026-02-03; there have been ~13 commits since then (the April 3 "cleanup, tests, fixes, database manager" + follow-ups), and that's where the current problems sit. Three buckets to address before this can re-merge:

## 1. CI is red on `check`, `test`, and `frontend`

All three are real compile/format errors introduced after the approval:

- **`test` fails to compile.** Four separate issues:
  - The big import-block reshuffle in `api/src/tasks.rs:1-28` dropped `use chrono::Duration;`, but lines 379 (`Duration::hours(1)` in `clean_oauth_states_handler`) and 399 (`Duration::days(7)` in `clean_download_counts_4h_handler`) still use the unqualified form. Line 388 (`chrono::Duration::days(30)` in your new `clean_webhook_data_handler`) is fully qualified and works. Pick one and apply consistently.
  - `api/src/tasks.rs:851` and `:922` import/use `crate::gcp::BigQueryQueryResult`, which doesn't exist on main anymore (grep returns nothing in `api/src/gcp.rs`). Looks like it was removed in a prior refactor and this test got merged forward without an update.
  - `api/src/api/package.rs:3047,3058` use `WebhookPayload` / `WebhookEventKind` without `use` statements (same names imported at `api/src/tasks.rs:11-12`).
  - `api/src/db/models.rs:1236` uses `<sqlx::Postgres as sqlx::database::HasArguments<'q>>::ArgumentBuffer` — the `HasArguments` trait was removed in sqlx 0.8 (the lockfile bumped 0.7→0.8 in this PR). The replacement is `<sqlx::Postgres as sqlx::Database>::ArgumentBuffer<'q>`. There's a 0.7→0.8 migration note for custom `Encode`/`Decode` impls — worth `grep -rn HasArguments` to make sure this is the only site.
- **`check` (cargo fmt) is red** on `api/src/api/scope.rs` (the `decrypt_webhook_secret` `map_err` block) and `api/src/util.rs:455`. `cargo fmt` fixes both.
- **`frontend` (deno fmt) is red** with one unformatted file — a trailing blank line in one of the new webhook routes.

Get the test job green first — there's a real chance the sqlx 0.8 upgrade has surfaced other latent breakage hidden behind the failure-to-compile.

## 2. SSRF: delivery-time URL validation is missing

This is the most important substantive concern. `validate_webhook_url` (util.rs:411) is solid for create-time — it blocks the IP-literal forms of private/loopback/link-local/CGNAT/`fc00::/7`, `localhost`, `*.local`, and `metadata.google.internal`. But at *delivery* time (`dispatch_webhook` in tasks.rs:564), the URL string is fetched straight out of the DB and handed to `reqwest::Client::new().post(webhook.url)...send()` with no further IP check.

Three real attack paths this enables:

1. **A hostname that resolves to a private IP.** Set up `webhook.attacker.com` with an A record pointing at `10.0.0.5` (or `169.254.169.254`, GCP metadata, AWS IMDS, internal Kubernetes services, internal admin endpoints, etc.). The create-time check sees a hostname (not an IP literal), so all the IP-range guards are bypassed, and only `localhost` / `.local` / `metadata.google.internal` get rejected by string match. Then at delivery time, DNS resolves to the private IP and the HTTP request hits an internal service.
2. **DNS rebinding.** Public IP at create time, private IP at delivery time. Same outcome.
3. **HTTP redirects.** `reqwest::Client::new()` follows up to 10 redirects by default. Even if the initial connection is to a public IP, a redirect target can point anywhere.

The blast radius is significant because *the response body is stored in `webhook_deliveries.response_body` and shown to the endpoint owner* (the whole point of the deliveries UI). So an attacker can read internal HTTP responses — service-account tokens from `169.254.169.254/computeMetadata/v1/instance/service-accounts/default/token`, internal admin pages, anything reachable from the Cloud Run egress. This isn't theoretical; it's near-textbook stored-SSRF.

Mitigation outline (one PR, all together):
- Build a `reqwest::Client` once at startup that uses a custom DNS resolver (`hyper-trust-dns` or `reqwest::dns::Resolve`) which, after resolving the hostname, rejects the request if the resolved IP is in any of the same ranges your `is_private_ip` / `is_link_local` / `is_loopback` covers. The same client should set `redirect::Policy::none()` (or `limited(0)`) so each hop has to be checked again, and `.timeout(Duration::from_secs(30))` to bound how long a slow target can hold the dispatch task.
- Cap the response body at, say, 1 MiB before storing. Unbounded `.text().await` lets a malicious target fill arbitrary amounts of Postgres `TEXT` (1GB max per row).
- The same restricted client should be used for the `redeliver` and "test webhook" code paths too — they all funnel through `dispatch_webhook`.

None of this is hard, and it closes the open hole. The fact that the *delivery owner* sees the response body is what makes the SSRF impactful; without that flow it'd just be a blind SSRF. So this isn't optional.

## 3. Smaller items

- Cleanup window mismatch: PR description says "7 days", `clean_webhook_data_handler` uses 30 days (`tasks.rs:388`). Pick one and align both.
- `WEBHOOK_SECRET_ENCRYPTION_KEY=AAAAAAAA...A=` in `.env.example` — the all-zero key is fine for local dev but worth a `# DO NOT USE IN PRODUCTION` comment, since it's the kind of thing that gets copy-pasted into a staging env once and forgotten.
- No encryption-key versioning. Rotating `WEBHOOK_SECRET_ENCRYPTION_KEY` makes every stored secret undecryptable. Not blocking for the initial ship, but worth a `key_id` byte prefixed to ciphertext if rotation is ever expected (very common pattern: 1 byte for key generation, then nonce, then ciphertext).
- `reqwest::Client::new()` inside `dispatch_webhook` rebuilds the entire client + TLS state on every dispatch. Once you build the SSRF-safe client, lift it into shared state (`Arc<reqwest::Client>` in `MainRouterOptions`) so connection pools survive.
- HMAC is only computed for `WebhookPayloadFormat::Json`. Discord/Slack don't get signed, which matches their conventions, but worth a one-line comment so future me doesn't "fix" this thinking it's a bug.

Rest is inline.

api/src/util.rs
```diff
@@ -405,6 +405,119 @@ where
...
+      return Err(ApiError::MalformedRequest {
+        msg: "webhook URL must not target private or loopback addresses".into(),
+      });
+    }
+  }
+
+  // Block well-known dangerous hostnames
+  let lower = host.to_lowercase();
+  if lower == "localhost"
+    || lower.ends_with(".local")
```
crowlbot, line 442: This blocks IP **literals** in private ranges, but a hostname like `webhook.attacker.com` that A-records to `10.0.0.5` (or `169.254.169.254` for the GCE/AWS metadata endpoint) passes through here unscathed. The string-match block at lines 440-446 catches `localhost`, `*.local`, `metadata.google.internal` — but not arbitrary attacker-controlled DNS.

The full fix has to be at delivery time (see top-level comment): use a `reqwest::Client` with a custom DNS resolver that re-applies these IP-range checks against the *resolved* address, with redirects disabled. The create-time check here can stay as a UX shortcut ("don't even let users save a webhook that points at an obvious literal"), but the security boundary lives in the dispatcher, not here.

api/src/util.rs
```diff
@@ -405,6 +405,119 @@ where
...
+    || lower == "metadata.google.internal"
+  {
```
crowlbot, line 444: AWS IMDS lives at `169.254.169.254` (covered by `is_link_local` / `169.254.0.0/16` already) but **also** at `metadata.aws.internal` and `instance-data.ec2.internal`. Azure uses `169.254.169.254` too plus several `*.azure.com` internal-only hosts. If JSR runs on AWS or Azure in any environment (staging? eu region?), worth adding those hostnames to this string-match block. GCP-only deployments are fine as-is.

api/src/tasks.rs
```diff
@@ -403,3 +446,644 @@ fn deserialize_version_download_count_from_analytics(
...
+  let trace_id = span_ctx.trace_id().to_string();
+  headers.insert("x-deno-ray", trace_id.parse().unwrap());
+
+  if let Some(signature) = signature {
+    headers.insert("X-JSR-Signature", signature.parse().unwrap());
+  }
+
+  let request_headers = serde_json::to_value(headers_to_map(&headers))?;
+
+  let response = match reqwest::Client::new()
```
crowlbot, line 771: `reqwest::Client::new()` here means:

1. New client + new TLS state per dispatch — no connection pooling, every dispatch pays the TLS handshake.
2. No timeout — a slow webhook target can hold this future open until the Cloud Tasks queue's own task deadline.
3. Default redirect policy of 10 hops — a redirect chain can land on any host, including private IPs after the initial connect.
4. No restriction on body size — `.text().await` at line 801 reads the entire response into memory, which then goes straight into `webhook_deliveries.response_body TEXT`.

A shared client built once at startup (`MainRouterOptions::webhook_client: Arc<reqwest::Client>` alongside the encryption key) with:

```rust
reqwest::Client::builder()
    .timeout(Duration::from_secs(30))
    .redirect(reqwest::redirect::Policy::none())
    .dns_resolver(Arc::new(SsrfSafeResolver))  // see top-level comment
    .build()?
```

fixes all four. The SSRF concern is the load-bearing one; the other three are robustness.

api/src/tasks.rs
```diff
@@ -403,3 +446,644 @@ fn deserialize_version_download_count_from_analytics(
...
+      return Err(anyhow::Error::from(err).into());
+    }
+  };
+
+  let status = response.status();
+  let success = status.is_success();
+  let response_http_status = status.as_u16() as i32;
+  let response_headers =
+    serde_json::to_value(headers_to_map(response.headers()))?;
+  let response_body = response.text().await.map_err(anyhow::Error::from)?;
```
crowlbot, line 801: `response.text().await` reads the response body in full with no cap. Webhook deliveries' `response_body` column is `TEXT` (1GB Postgres limit, but practical memory exhaust well before that). A malicious target can return arbitrarily large responses on every retry.

Cap it. `bytes_stream` with a running total, or `Response::bytes` after `Content-Length` check, or pull in `bytes_with_limit`-style helper. 1 MiB feels like a generous ceiling for "useful diagnostic output" without enabling abuse.

api/src/tasks.rs
```diff
@@ -348,6 +382,15 @@ pub async fn clean_oauth_states_handler(req: Request<Body>) -> ApiResult<()> {
   Ok(())
 }
 
+#[instrument(name = "POST /tasks/clean_webhook_data", skip(req), err)]
+pub async fn clean_webhook_data_handler(req: Request<Body>) -> ApiResult<()> {
+  let db = req.data::<Database>().unwrap().clone();
+  let cutoff = Utc::now() - chrono::Duration::days(30);
```
crowlbot, line 388: 30-day retention here, but the PR description says "7 days". Pick one. (My read: 30 days is generous and probably what most users want for debugging missed events, but if the description is correct it should be 7. Either way the docs should match the code.)

Also: this is the one site in the file that uses fully qualified `chrono::Duration::days(...)`. Lines 379 and 399 use the unqualified `Duration::days(...)` / `Duration::hours(...)`, which doesn't compile because the import-block reshuffle at the top of the file dropped `use chrono::Duration;`. Easiest fix: add the `use` back.

api/src/tasks.rs
```diff
@@ -403,3 +446,644 @@ fn deserialize_version_download_count_from_analytics(
...
+  use chrono::Utc;
+  use serde_json::json;
+  use uuid::Uuid;
+
+  use crate::db::DownloadKind;
+  use crate::db::EphemeralDatabase;
+  use crate::db::ExportsMap;
+  use crate::db::NewPackageVersion;
+  use crate::db::PackageVersionMeta;
+  use crate::gcp::BigQueryQueryResult;
```
crowlbot, line 851: `crate::gcp::BigQueryQueryResult` doesn't exist on main today — `grep -rn BigQueryQueryResult api/src/gcp.rs` returns nothing. This test was probably written against an earlier shape of `gcp.rs` that's since been refactored away during one of the merges from main. Either restore the type as a test-only helper, or update the test to use whatever the new query-result shape is.

api/src/db/models.rs
```diff
@@ -1105,3 +1104,197 @@ impl FromRow<'_, sqlx::postgres::PgRow> for AuditLog {
...
+    let s: sqlx::types::Json<WebhookPayload> =
+      sqlx::Decode::<'_, sqlx::Postgres>::decode(value)?;
+    Ok(s.0)
+  }
+}
+
+impl<'q> sqlx::Encode<'q, sqlx::Postgres> for WebhookPayload {
+  fn encode_by_ref(
+    &self,
+    buf: &mut <sqlx::Postgres as sqlx::database::HasArguments<'q>>::ArgumentBuffer,
```
crowlbot, line 1236: The sqlx 0.7→0.8 upgrade removed the `HasArguments` trait. The replacement for `<sqlx::Postgres as sqlx::database::HasArguments<'q>>::ArgumentBuffer` is `<sqlx::Postgres as sqlx::Database>::ArgumentBuffer<'q>` — `ArgumentBuffer` is now a GAT directly on the `Database` trait. Worth `grep -rn HasArguments` to make sure this is the only call site needing the update.

api/src/util.rs
```diff
@@ -405,6 +405,119 @@ where
...
+/// 32-byte key for AES-256-GCM encryption of webhook secrets at rest.
+#[derive(Clone)]
+pub struct WebhookSecretEncryptionKey(pub [u8; 32]);
+
+/// Encrypt a webhook secret using AES-256-GCM. Returns base64(nonce || ciphertext || tag).
+pub fn encrypt_webhook_secret(
+  key: &WebhookSecretEncryptionKey,
+  plaintext: &str,
+) -> String {
+  use aes_gcm::{Aes256Gcm, KeyInit, AeadCore, AeadInPlace};
```
crowlbot, line 484: `Aes256Gcm` with random nonce + base64(nonce ‖ ciphertext ‖ tag) — correct. Two follow-ups worth doing in a separate PR before this hits prod:

1. **Key versioning.** Reserve byte 0 of the encoded blob for a key generation ID. Today rotating `WEBHOOK_SECRET_ENCRYPTION_KEY` orphans every existing secret (`decrypt_webhook_secret` errors and the webhook silently stops being signed). Even if you don't implement rotation now, the byte is cheap to reserve and very painful to retrofit later.
2. **Confirm no `tracing::*!` log writes the raw `secret` field anywhere** in the create/update path — webhook secrets going across the wire are HTTPS-protected, but `#[instrument(skip(secret))]` or explicit body redaction in `WebhookEdit`-handling endpoints is worth a visual scan.

api/.env.example
```diff
@@ -14,4 +14,5 @@ S3_ACCESS_KEY=minioadmin
 S3_SECRET_KEY=minioadmin
 REGISTRY_URL=http://jsr.test
 NPM_URL=http://npm.jsr.test
+WEBHOOK_SECRET_ENCRYPTION_KEY=AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA=
```
crowlbot, line 17: All-zero key is fine for `jsr.test` local dev. Worth a one-line `# Generate a real 32-byte key with: openssl rand -base64 32 — DO NOT use this value in any non-local environment` so the staging Helm chart maintainer doesn't paste this in once and forget.

api/src/api/package.rs
```diff
@@ -2648,6 +2759,373 @@ pub async fn get_score_handler(
...
+  let enc_key = req.data::<crate::util::WebhookSecretEncryptionKey>().unwrap();
+
+  let iam = req.iam();
+  iam.check_scope_admin_access(&scope).await?;
+
+  let webhook = db
+    .get_webhook_endpoint(&scope, Some(&package), webhook_id)
+    .await?;
+
+  let payload = WebhookPayload::PackageVersionPublished {
```
crowlbot, line 3047: This is one of the test/compile errors — `WebhookPayload` and `WebhookEventKind` aren't imported in this file's `use` block. Same pattern as `api/src/tasks.rs:11-12`:

```rust
use crate::db::WebhookEventKind;
use crate::db::WebhookPayload;
```

PR #1265, opened 2026-01-24
feat: fallback registry system

crowlKats:
Adds env vars letting an instance name a *fallback registry* that serves packages the instance itself does not host, so a self-hosted JSR can depend on packages from jsr.io (or another instance) without mirroring them.

Two env vars configure it:

- `FALLBACK_REGISTRY_URL` (api) — consulted when resolving `jsr:` dependencies at publish time, and when building a package's dependency graph.
- `FALLBACK_ROOT_URL` / `FALLBACK_NPM_URL` (lb) — serve artifact requests the modules / npm bucket misses.

### Publish time

When a `jsr:` dependency has no matching version in this registry, the fallback is asked whether it has one that satisfies the constraint and exports the requested subpath. If it does, the dependency row records the fallback's URL in the new `dependency_fallback_url` column; if it doesn't, the publish fails with `unresolvableJsrDependency` as before.

Failures *talking to* the fallback (unreachable, 5xx, malformed metadata) are retryable task errors rather than publish failures — a fallback outage is this instance's infrastructure failing, not the publisher's mistake, and must not be reported to them as "your dependency doesn't exist".

Only the fallback's identity is stored, not the version it resolved to at publish time. That matches how a locally-resolved dependency is recorded: the row keeps the constraint and consumers re-resolve it at install time.

### Serve time

The lb serves artifact requests the bucket misses from the fallback. This is deliberately limited to *artifacts*: a package resolved from the fallback is **linked, not mirrored**, so the API (`/api/scopes/…`, package/version metadata, search) still 404s for it, and the frontend links out to the fallback registry rather than rendering a local page.

The fallback request is constructed fresh — method and path/query only — and never forwards the caller's inbound request, which would send their `Authorization` / `Cookie` / `Host` headers to a third-party origin on every bucket miss. It is restricted to GET/HEAD within the scoped (`/@…`) namespace, so an arbitrary 404 on this instance can't be turned into an attacker-chosen outbound request from our egress IP.

### Frontend

Dependencies resolved from the fallback link to the fallback registry and are drawn in a distinct color in the dependency graph.

### Tests

The fallback tests run against an in-process fake registry (`FakeFallbackRegistry`) rather than the real jsr.io, so they don't depend on network access or on what `@std/assert` currently publishes. Coverage includes the credential-forwarding regression, the non-scoped/unsafe-method rejections, and the fallback-is-down retryable path.


crowlbot (commented):
First-pass review on a PR that's been waiting a while — there's a coherent feature here (fallback registry for missing JSR deps, surfaced at both publish-time and serve-time, with a frontend signal), and the publish-flow plumbing is structured cleanly via the `ResolvedDependency` struct.

That said, there are a few things I'd want resolved before this could land — the most important being a credential-leak vector in the LB proxy fallback. Roughly in order of severity:

### 🚨 1. LB fallback proxies the *original* `Request` to a third-party origin (lb/proxy.ts:107–112)

```ts
if (response.status === 404 && fallbackUrl) {
  response = await fetch(
    new URL(url.pathname + url.search, fallbackUrl),
    request,
  );
}
```

Passing the inbound `Request` as `fetch`'s second arg forwards **all of its headers and body** to whatever `fallbackUrl` points at — including `Authorization`, `Cookie`, custom `X-*` API keys, and the original `Host`. For a self-hosted JSR instance with `FALLBACK_ROOT_URL=https://jsr.io`, every cache-miss on the modules bucket becomes an outbound request to jsr.io carrying the original client's credentials. That's:

- A **secret leak** on every misrouted/expired/path-guessed module URL.
- A **request smuggling** primitive — a malicious 404 path on the local instance lets an attacker probe the fallback registry from the local instance's egress IP.
- A behavior the user can't observe or opt out of from the client side.

Concrete fix: construct a new minimal request, no credentials, GET only, no body:

```ts
response = await fetch(new URL(url.pathname + url.search, fallbackUrl), {
  method: "GET",
  redirect: "follow",
});
```

And consider an **allowlist** rather than catch-all 404: only fall back for URL shapes that look like immutable module artifacts (e.g. paths matching `/@<scope>/<package>/<version>/...`), not arbitrary 404s — that closes the path-enumeration channel.

### 🚨 2. `fetch(url, request)` is also wrong on the publish-time tarball loader (api/src/api/package.rs:1860–1867)

```rust
let response = reqwest::get(&fallback_specifier).await.ok();
```

This one uses `reqwest::get(...)` (no headers, fine), but it's still:

- **No timeout** — `reqwest::get` builds a default client with no `timeout()` / `connect_timeout()`. A slow or hanging fallback registry stalls the entire dep-tree resolution. Build a shared `reqwest::Client` with `connect_timeout(5s).timeout(30s)` and reuse it for all fallback fetches.
- **No User-Agent**: hits to jsr.io will look like the generic `reqwest/0.x` UA, which makes incident triage on the fallback side harder. Use the existing `crate::util::USER_AGENT`.

Same applies to `tarball.rs::resolve_from_fallback` — `let client = reqwest::Client::new();` per call, no timeouts, no UA. Hoist the client construction.

### ⚠️ 3. `fallback_packages` set is racy and over-marks (api/src/api/package.rs:1850–1888)

```rust
async move {
  // ...
  let mut bytes = bucket.download(...).await?;
  let mut from_fallback = false;

  if bytes.is_none() && let Some(fallback_url) = &fallback_registry_url {
    let response = reqwest::get(&fallback_specifier).await.ok();
    if let Some(response) = response && response.status().is_success() {
      bytes = response.bytes().await.ok();
      from_fallback = true;
    }
  }

  if from_fallback {
    let package_id = format!("@{}/{}", scope.as_str(), package.as_str());
    fallback_packages.lock().await.insert(package_id);
  }
  // ...
}
```

`DepTreeLoader::load` is invoked concurrently for many specifiers. The same package can be loaded along multiple paths in the same dep tree (`@scope/foo/a.ts` and `@scope/foo/b.ts`). If `a.ts` loads from local and `b.ts` falls back, the whole package gets marked `from_fallback` and then `GraphDependencyCollector` tags it as fallback-resolved — but its contents are actually a mix from both registries. Worse: the consumer reading the dep graph sees a deceptive "this came from jsr.io" link that doesn't reflect what was actually bundled.

If the intent is "any miss is enough to taint the whole package," that should be stated explicitly (comment + maybe an `Ordering::Relaxed` `AtomicBool` to make it cheaper). If the intent is "each file records where it came from," the set needs to be `(package_id, path)`-keyed, and the collector logic needs to follow.

### ⚠️ 4. `resolve_from_fallback` returns the resolved version but the caller throws it away (tarball.rs ~225 → package.rs ~1842)

```rust
if let Some(fallback_url) = &fallback_registry_url
  && let Ok(Some(_resolved_version)) = resolve_from_fallback(...).await
{
  resolved_dependencies.insert(ResolvedDependency {
    kind: *kind,
    req: req.clone(),
    registry_url: Some(fallback_url.to_string()),
  });
}
```

Two consequences:

- The DB row stores the constraint (`req.req.version_req`), not the actual version pinned by the fallback at publish time. If the fallback registry yanks or replaces versions, the local instance's recorded dep effectively floats — clients re-resolving at consume time may pick a different version than was used at publish-time analysis.
- The publish flow loses an audit trail of "what was actually there in the fallback when we accepted this publish."

Either persist the resolved version explicitly (new column or extend `dependency_constraint` semantics), or document why the constraint is sufficient. For npm-like supply-chain guarantees, the former matters.

### ⚠️ 5. Trust boundary: fallback flake fails the publish

`PublishError::FallbackRegistryError { url, error }` is returned as a hard publish failure (`Some("fallbackRegistryError")`). Any network blip while talking to the fallback during a publish turns a publish that *would have succeeded if the fallback weren't checked* into a `Failure` task.

Concrete failure mode: jsr.io is degraded, every locally-published package that imports `jsr:@std/*` now fails to publish on the self-hosted instance because the fallback can't be reached, even when the local has its own copy. The "ensure all of the JSR dependencies are resolvable" loop now silently depends on `fallback_url`'s availability whenever the local DB doesn't carry the dep.

Two cleaner shapes:
- Treat fallback errors as "skip fallback, fail back to `UnresolvableJsrDependency`" so the publish only fails on `not in local AND not in fallback OR fallback is down`.
- Or: treat fallback errors as soft, mark the dep `registry_url = None` (treat as local), and let the consume-time fallback handle it.

### Smaller things

**6. (api/src/api/package.rs:1849)** — the fallback `reqwest::get(&fallback_specifier)` constructs `fallback_specifier` by `replacen(registry_url.as_str(), fallback_url.as_str(), 1)`. That's a string-level substitution: if `registry_url = https://jsr.test/` and the specifier is `https://jsr.test/...`, fine — but if the specifier is `https://jsr.test/@scope/foo/1.2.3-jsr.test/mod.ts` (the path happens to contain the registry hostname), the `replacen` replaces only the *first* occurrence, which is the prefix, so this is actually OK. Worth a comment noting why `replacen(.., .., 1)` is safe (i.e. the registry URL is guaranteed to be a prefix) so future readers don't generalize it.

**7. `lb/main.ts` — fallback is wired into modules + npm, but not into the API meta endpoints**. A package that resolved from the fallback at publish time has its `dependency_fallback_url` recorded; the frontend then renders a link to the fallback registry. But the API itself (`/api/scopes/.../meta.json`, version metadata, etc.) still 404s locally for that package — only the *files* fall back. That seems intentional (the package is "linked, not mirrored"), but worth a comment in lb/main.ts so it's not later "fixed" by accident, and the API behavior should probably be documented in the PR description.

**8. The admin requeue path** (api/src/api/admin.rs:308–321) now picks up `fallback_registry_url` from request state. Requeuing a publish task that was originally accepted against fallback registry A and then run against B is a footgun. Either fail the requeue if the resolved deps' `dependency_fallback_url` doesn't match the current config, or document that requeue uses current config.

**9. Migration filename mismatch** (api/migrations/20260124120000_dependency_registry_url.sql): the file is named `_dependency_registry_url` but the column is `dependency_fallback_url`. Trivial, but renaming the file matches the column and saves grep friction later.

**10. Color `#a855f7` for fallback-resolved JSR deps** (frontend/.../DependencyGraph.tsx:240) — the rest of the file uses semantic color names via Tailwind. Hard-coding `#a855f7` (Tailwind `purple-500`) here breaks pattern. Either pull the value from the same source as the existing colors, or add a comment.

**11. The `frontend/utils/api_types.ts` change makes `fallbackUrl: string | null` non-optional**. Existing API responses don't carry this field, so deserialization on a client built against the new types reading an old server returns `undefined` for `dep.fallbackUrl`, not `null` — the type lies. Make it `fallbackUrl?: string | null` (or update the server to always emit `null` explicitly).

**12. CI**: the `test` job is failing on a workflow run from > a year ago and the logs are expired. A rebase + push to get a fresh CI run is the only way to see what's broken now.

---

Happy to re-review once the LB credential-forwarding and the timeout/UA fixes are in, and the design intent on the fallback_packages tainting + publish-error semantics is settled.

crowlbot (commented):
Went through this against every point from my earlier review — all of it is addressed, most with more care than I asked for:

1. **Credential leak (🚨 the blocker)**: `fetchFromFallback` in `lb/proxy.ts` no longer forwards the inbound `Request`. It builds a fresh `fetch(target, { method, redirect: "follow" })` with no headers/body, and `isFallbackEligible` restricts it to `GET`/`HEAD` on `/@...` paths only — closes both the credential-leak and the path-enumeration angle. `proxy_test.ts` has a test that asserts the outbound `Authorization` header is `null`, plus dedicated tests for the namespace restriction, method restriction, and fallback-miss/throw paths.
2. **No timeout/UA (🚨)**: `shared_http_client()` in `util.rs` sets `user_agent` + `connect_timeout`, and every fallback call chains `.timeout(FALLBACK_REQUEST_TIMEOUT)` (30s) — used consistently in both `tarball.rs::resolve_from_fallback` and the doc-loader path in `package.rs`.
3. **`fallback_packages` tainting semantics (⚠️)**: now has an explicit comment stating tainting is per-package by design ("a single file served by the fallback means this registry does not hold the package") — the ambiguity is resolved by documenting intent, which is what I asked for.
4. **Resolved version discarded (⚠️)**: still stores the constraint, not the pinned version — but now with a comment explaining that's deliberate and matches how a locally-resolved dependency is stored (re-resolved at install time). Documented rather than changed, which was the other acceptable option I gave.
5. **Publish fails hard on fallback outage (⚠️)**: this is the best fix of the batch — `FallbackRegistryError` now propagates via `?` instead of collapsing into `UnresolvableJsrDependency`, and `user_error_code()` returns `None` for it (same bucket as `S3DownloadError`/`DatabaseError`), so it's retried as infra failure rather than told to the publisher as "your dependency doesn't exist."
6. **`replacen` prefix assumption**: now guarded with an explicit `starts_with` check before the replace, plus a comment — better than just documenting the assumption.
7. **API doesn't know about fallback packages**: documented in `lb/main.ts` as intentional (artifacts are linked, not mirrored).
8. **Admin requeue uses current fallback config**: documented at the call site.
9. **Migration filename**: renamed to `20260821000000_dependency_fallback_url.sql`, matches the column.
10. **Hardcoded hex colors**: pulled into named `COLOR_*` constants with a comment on why they can't be Tailwind classes (graphviz node attrs, not DOM).
11. **`fallbackUrl` optionality in `api_types.ts`**: still `string | null` non-optional — on reflection this is fine, since the Rust side serializes `Option<String>` as `null` (not omitted) with no `skip_serializing_if`, so the field really is always present. Not flagging this further.
12. **CI on stale logs**: moot now, this branch has been rebased onto current main and has fresh runs.

Also appreciate the switch to an in-process `FakeFallbackRegistry` for the tests instead of hitting the real jsr.io — that was a latent flakiness/external-dependency risk in the original test suite that I hadn't called out but is a clear improvement.

Nothing outstanding from me. `check`/`test` were still pending as of this review — will take a final look once they're green, but I don't expect further comments.

crowlbot (commented):
This is a second self-driven round on top of the review-feedback commit — reviewed it in full, it's solid and catches a real bug I missed in my first pass:

**The important one — `process_tarball`'s local/fallback branching (`tarball.rs`)**: previously, when a *locally-hosted* package's constraint matched no local version (`found_local = false`, no subpath error either), the code fell through and validated the dependency against the fallback anyway, recording `registry_url: Some(fallback_url)`. But the lb only ever consults the fallback on a bucket miss — a locally-hosted package's `meta.json` is always served from the bucket, so the fallback is never actually queried for it at runtime. That meant a publish could succeed by validating a constraint against a version that exists on the fallback but that no consumer could ever actually resolve to, since the local copy (which doesn't satisfy the constraint) wins. Now correctly branches on `!versions.is_empty()`: any local presence means resolution happens strictly locally, fallback isn't consulted at all. Well covered — the two new tests configure an *unreachable* fallback URL and assert the publish still fails cleanly with `unresolvableJsrDependency`/`invalidJsrDependencySubPath`, which also proves the fallback was never contacted.

**Cache poisoning guard (`lb/proxy.ts`)**: fallback responses are now only cached in the lb's edge cache when the fallback explicitly sends `max-age`/`s-maxage` (and doesn't send `private`/`no-store`). Without this, an unconditionally-cached fallback response for a mutable resource (meta.json, npm packument) could shadow a local publish or upstream release indefinitely, since nothing in the publish path purges this cache key. Good catch, this closes a real staleness hole the first round didn't touch.

**Subpath-hosted fallback URLs**: both the lb (`target.pathname = base.replace(/\/+$/,'') + path`) and the API (`parse_base_url` clap value_parser normalizing to a trailing slash) now handle `FALLBACK_REGISTRY_URL=https://mirror.corp/jsr` correctly — `new URL(path, base)` with an absolute `path` was silently discarding the base's own path before, which would've broken any non-root-hosted fallback.

**Panic safety**: the doc-loader's `serde_json::from_slice(...).unwrap()` on fallback-fetched bytes is now a propagated `LoadError` — right call, since that data crosses a network boundary to a third party and was previously one malformed response away from a panic.

**`@jsr/`-mapped npm deps on a non-local package now record `dependency_fallback_url`** — closes a gap where the frontend would link to a local package page that 404s.

**Local dev no longer defaults to jsr.io as fallback** — good, avoids silently making local dev network-dependent and masking missing-upload bugs.

Everything here reads as intentional, well-tested cleanup, not new scope. No further concerns from me. Build passed, `test` job still pending as of this review.

crowlbot (approved):
Final pass after the two follow-up rounds — nothing new since my last review (still `c34f05c`), CI is fully green (build + test).

Recap of what's been verified across this review:
- The original credential-leak vector (forwarding the caller's `Authorization`/`Cookie` to the fallback registry) is closed, with a test asserting the header never leaves this instance.
- Fallback fetches use a shared client with a UA and timeout everywhere.
- Fallback registry outages are retryable infra errors, not fatal "your dependency doesn't exist" publish failures.
- The subtler bug from the second round — a locally-hosted package's dependency constraint could previously validate against the fallback even though the lb never actually queries the fallback for a package it hosts — is fixed and regression-tested with an *unreachable* fallback URL to prove it's no longer consulted.
- Cache poisoning of mutable fallback resources (meta.json, packuments) is guarded by honoring the fallback's own `Cache-Control`.
- Remaining design choices (constraint-not-pinned-version storage, per-package not per-file tainting, requeue using current fallback config) are deliberate and documented, not oversights.

Approving.

PR #1270, opened 2026-01-31
feat: private packages

crowlKats:
Adds support for private packages: packages that are only visible to (and readable by) members of their scope.

## How it works

- `packages.is_private` column; a package's visibility can be toggled from the package settings by scope admins, but only while the package has no published versions (artifacts are not migrated between buckets on a toggle).
- Private packages publish their module files, manifests, and npm tarballs to two new buckets (`modules-private`, `npm-private`) instead of the public ones.
- The lb serves module/npm requests from the public bucket first; on a 404 it checks the requester's access via the new `GET /api/scopes/:scope/packages/:package/check-access` endpoint (forwarding `Authorization`/cookie) and only then serves from the private bucket. Private responses never enter the shared edge cache and are served with `Cache-Control: private`.
- All package API endpoints mask private packages as `packageNotFound` for callers without access. Search (Algolia), the sitemap, package stats, and the global package list exclude private packages.
- New token permission `package/read` (package / scope / full granularity), plus a `full` granularity for `package/publish`. Token permissions are deserialized strictly (exact key-set match) so a malformed permission object can no longer fall through to a full-access grant. A matching `package/publish` permission also grants read access to the package, which the publish flow needs to fetch manifests back.
- The token creation flow in the frontend now offers separate publish / read / API usage flows.

Requires denoland/deno#36656 for `deno publish` to work with private packages (the CLI's unauthenticated package-existence check misreads a private package as missing).


crowlbot (commented):
The overall architecture — `is_private` column, private GCS buckets, `check-access` API endpoint gated on scope membership, lb-level fallback from public to private bucket on 404 — is the right shape for private packages.

A few things to address before this can merge:

**Debug `console.log` leaking auth headers (security)**

In `lb/main.ts`, `validatePackageAccess` has:
```ts
console.log(resp, authHeader);
```
This logs the `Authorization` header (which contains the user's token) to Worker logs on every private package access. That needs to be removed before this ships.

**CI is failing**

Both the `frontend` and `test` jobs are red on the latest CI run. The PR needs to be brought up to green before review can conclude.

**Unused `KVNamespace` interface**

`lb/types.ts` adds a `KVNamespace` interface to the global scope, but nothing in this PR (or apparently anything in `lb/`) references it. If it's prep work for a future feature it probably shouldn't land here; if it's needed, a usage should be visible.

**`is_scope_member` vs `check_package_read_access` inconsistency**

`get_handler` gates private-package visibility on `iam.is_scope_member(&scope)`, while `check_access_handler` (which the lb calls for npm/module file access) uses `iam.check_package_read_access(&scope, &package)`. If `check_package_read_access` is strictly more permissive or restrictive than `is_scope_member`, there could be cases where a user can see the package page but not download it (or vice versa). Worth making sure these two code paths are consistent, or adding a comment explaining the intended difference.

**Missing PR description**

The PR body is empty. Even a brief summary of what the feature does and any design decisions (e.g., why 404-and-retry rather than a header-driven bucket selection) would help reviewers and future readers.

crowlbot (approved):
Confirmed: `get_recently_published_packages_by_user` now maps `is_private` into the `Package` struct (fixing the compile error), and additionally filters `WHERE packages.is_private = false` — good catch beyond just fixing the build, since without that filter the public user-profile page would have revealed the existence and names of a user's private packages via the "recently published" list. CI is now fully green.

Between this and the earlier round (permission-deserialization fix, unified access-check helper, tarball/diff endpoint guards, cache-poisoning prevention), I'm satisfied with the security posture of this PR. Approving.

PR #1297, opened 2026-02-17
fix: correctly handle licenses with the same content

### LINKED ISSUE #1296: `GPL-3.0-or-later` rejected as `invalidLicense` despite being a valid SPDX identifier
`deno publish` fails with `invalidLicense` when `"license": "GPL-3.0-or-later"` is set in `deno.json`. `--dry-run` passes but the actual publish is rejected by the server.

Is this a known issue, or is there a workaround I'm missing? (e.g. a different identifier format, etc.)

Likely related to #1161.

crowlKats:
Closes #1296


kt3k (approved):
LGTM (when clippy issue resolved)

Issue #1316, opened 2026-02-27
npm.jsr.io returns 404 for URL-encoded scoped package requests


fengmk2:
## Bug Description

The npm.jsr.io registry returns 404 when the scoped package name contains a URL-encoded slash (`%2F`), but works fine with a literal `/`. This breaks pnpm which URL-encodes the package name in registry requests.

## Reproduction

```bash
# URL-encoded slash — returns 404
curl -v "https://npm.jsr.io/@jsr%2Fstd__yaml"
# HTTP/2 404

# Literal slash — returns 200
curl -v "https://npm.jsr.io/@jsr/std__yaml"
# HTTP/2 200
```

## Impact

pnpm URL-encodes the scoped package name when fetching from the registry, so any project using `@jsr/*` packages via pnpm hits this error:

```
ERR_PNPM_FETCH_404  GET https://npm.jsr.io/@jsr%2Fstd__yaml: Not Found - 404

@jsr/std__yaml is not in the npm registry, or you have no permission to fetch it.
```

This affects `pnpm install`, `pnpm dedupe --check`, and other pnpm commands that resolve packages.

## Root Cause

This was likely introduced by https://github.com/jsr-io/jsr/pull/1306 which switched the npm bucket from GCS to Cloudflare R2.

The `proxyToR2` function in `lb/proxy.ts` uses the URL path directly as the R2 key:

```typescript
const key = path.slice(1);
```

When pnpm requests `/@jsr%2Fstd__yaml`, the key becomes `@jsr%2Fstd__yaml` (still URL-encoded). But the R2 object is stored as `@jsr/std__yaml` (with a literal `/`). R2 doesn't auto-decode `%2F` in keys, whereas the previous GCS proxy (via `cachedFetch` / HTTP) would have handled URL decoding at the HTTP layer.

## Suggested Fix

Decode the key before looking it up in R2:

```typescript
const key = decodeURIComponent(path.slice(1));
```

## Expected Behavior

Both `https://npm.jsr.io/@jsr%2Fstd__yaml` and `https://npm.jsr.io/@jsr/std__yaml` should return the same 200 response, since `%2F` is the URL-encoded form of `/`.

## Environment

- pnpm v10.28.0
- Tested on macOS (Darwin 25.2.0)

crowlKats:
@fengmk2 good catch, thanks for this. could you open a PR with the fix, so we can land it ASAP?

PR #1317, opened 2026-02-27
fix(lb): decode URL-encoded paths in R2 proxy for npm registry

fengmk2:
## Summary

- Add `decodeURIComponent()` to the R2 key lookup in `proxyToR2` so URL-encoded scoped package names resolve correctly

## Problem

After #1306 switched the npm bucket from GCS to R2, pnpm users get 404 errors when resolving `@jsr/*` packages because pnpm URL-encodes the `/` in scoped package names:

```bash
# pnpm sends this — returns 404
curl "https://npm.jsr.io/@jsr%2Fstd__yaml"

# Direct slash works — returns 200
curl "https://npm.jsr.io/@jsr/std__yaml"
```

The GCS proxy handled this transparently at the HTTP/fetch layer, but R2's `bucket.get(key)` requires the exact key — `%2F` doesn't match the stored `/`.

## Fix

Decode the path before using it as an R2 key:

```diff
- const key = path.slice(1);
+ const key = decodeURIComponent(path.slice(1));
```

Fixes https://github.com/jsr-io/jsr/issues/1316

### DIFF
--- a/lb/proxy_test.ts
+++ b/lb/proxy_test.ts
@@ -9,11 +9,11 @@ function createFakeBucket(
   objects: Record<string, { body: string; contentType?: string }>,
 ): PartialBucket {
   return {
-    async head(key: string): Promise<R2Object | null> {
+    head(key: string): Promise<R2Object | null> {
       const obj = objects[key];
-      if (!obj) return null;
+      if (!obj) return Promise.resolve(null);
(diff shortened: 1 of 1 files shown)


crowlKats:
more tests are always welcome!
thanks a lot for this fix, not sure how this was working as expected for me locally when i implemented these changes.

PR #1331, opened 2026-03-06
feat: JSR service account & let users delete their account themselves

crowlKats:
this adds a service account (imagine similar to @ghost), that in the future can be used for external management of the service by an admin account (like #1265); this service account has no scope limit.

Then this also adds 

- User deletion (admin + self-service): Admins can delete users via the admin panel, and users can delete their own accounts from account settings.                                                                                    
- Scope ownership transfer: When a user is deleted, scopes where they are the sole member are transferred to the service account. Scopes with other members transfer creator to the next admin (or any member).
- Audit log preservation: Audit logs from the deleted user are reassigned to the service account, while the "user_delete" audit entry retains the actual actor who performed the deletion.                                             
 - Confirmation modals: Both the self-service and admin delete modals require typing the username to confirm, and show affected scopes with transfer details.


api/migrations/20260304000000_jsr_service_account.sql
```diff
@@ -0,0 +1,15 @@
+ALTER TABLE users ALTER COLUMN scope_limit DROP NOT NULL;
+
+INSERT INTO users (id, name, avatar_url, is_staff, scope_limit)
+VALUES (
+       '00000000-0000-0000-0000-000000000000',
```
bartlomieju, line 5: I assume this is on purpose so it's a known value. Not sure if it would make sense to have it randomized

bartlomieju: I think it's fine after all

frontend/routes/admin/users.tsx
```diff
@@ -67,31 +68,34 @@ export default define.page<typeof handler>(function Users({ data, url }) {
...
+                    },
+                    {
+                      name: "isBlocked",
+                      label: "is blocked",
+                      type: "boolean",
+                      value: user.isBlocked,
+                    },
+                  ]}
+                />
+                <DeleteUser userId={user.id} userName={user.name} />
```
miracatbot: The admin table still renders a delete action for the service account row, but the backend always rejects deleting that user. Can we hide/disable the delete control for the nil-UUID account so admins do not go through a doomed destructive flow?

frontend/routes/account/settings.tsx
```diff
@@ -97,16 +100,18 @@ export default define.page<typeof handler>(function AccountInvitesPage({
             any scopes that you are the sole owner of will be orphaned. You will
```
miracatbot: The settings page still says scopes where the user is the sole owner will be "orphaned", but the new behavior transfers them to the service account. Can we update this copy to match the actual post-delete behavior? The modal text already uses the new wording.

miracatbot (commented):
A few targeted comments from the review:

- the new service account is protected from deletion, but it still looks mutable through the existing admin update path, which seems risky now that it acts as a built-in system principal
- the admin UI still offers a delete flow for the service account even though the backend always rejects it
- the account settings copy still says sole-owner scopes become orphaned, but the new behavior transfers them to the service account

Overall I think the direction here makes sense, but I’d address those before merging.

miracatbot (commented):
A few targeted comments from the review:

- the new service account is protected from deletion, but it still looks mutable through the existing admin update path, which seems risky now that it acts as a built-in system principal
- the admin UI still offers a delete flow for the service account even though the backend always rejects it
- the account settings copy still says sole-owner scopes become orphaned, but the new behavior transfers them to the service account

Overall I think the direction here makes sense, but I’d address those before merging.

api/migrations/20260304000000_jsr_service_account.sql
crowlKats: yeah and we already have a testing account on that ID that we delete in staging/prod, so just reusing that

api/src/auth/mod.rs
```diff
@@ -467,7 +468,7 @@ mod tests {
     eprintln!("resp status {}", resp.status());
     assert!(resp.status().is_success());
```
crowlbot, line 469: This is the test failure in CI. The whole test now PATCHes `mock_user_id = Uuid::nil()`, which the guard at `admin.rs:96` rejects with 400. The state-transition assertions you updated (e.g. `assert!(user.is_staff)` on line 453) never get exercised because the first PATCH returns BAD_REQUEST.

Cleanest fix: scrap this test entirely and rely on the negative coverage in `admin.rs:629-642` (you already have `expect_err_code(BAD_REQUEST, "cannotModifyServiceAccount")` there). If you want to keep positive coverage of `PATCH /api/admin/users/:user_id` working on a regular user, replace `mock_user_id` with a fresh non-nil UUID via `TestSetup` and revert the asserts to test-user defaults.

api/migrations/20260304000001_user_delete_cascades.sql
```diff
@@ -0,0 +1,33 @@
+-- Add proper ON DELETE behavior to foreign keys referencing users(id)
+-- so that deleting a user doesn't require manual cleanup of every table.
+
+-- scopes.creator: change from CASCADE (would delete scopes!) to SET NULL
+ALTER TABLE scopes DROP CONSTRAINT IF EXISTS scopes_creator_fkey;
+ALTER TABLE scopes ADD CONSTRAINT scopes_creator_fkey
+    FOREIGN KEY (creator) REFERENCES users(id) ON DELETE SET NULL;
```
crowlbot, line 7: Worth a comment on this line specifically: before this migration, deleting a user would have cascaded into *deleting every scope they created*. There was no user-delete path in the code so it was dormant, but the PR enables that path for the first time — so flipping from CASCADE to SET NULL isn't a nice-to-have, it's a correctness prerequisite. One-line `-- WARNING: was CASCADE before this PR — would have deleted scopes` makes future archeology much easier.

api/src/db/database.rs
```diff
@@ -410,17 +410,143 @@ impl Database {
...
+    let audit_actor = if *actor_id == id {
+      &service_account_id
+    } else {
+      actor_id
+    };
+    audit_log(
+      &mut tx,
+      audit_actor,
+      is_sudo,
+      "user_delete",
```
crowlbot, line 527: This reassignment is irreversibly lossy: every prior `audit_logs` row that referenced the deleted user now points to the service account, with no record of *who* originally did the action. After deletion you can't distinguish "actually performed by the system" from "performed by some user who later got deleted."

Compare what you (correctly!) do for the `user_delete` entry itself on line 519 — you stash `performed_by` in the JSONB `meta` to preserve forensics. The same pattern would work here, e.g. add an `original_actor_id uuid` column on `audit_logs` (default NULL) and `UPDATE audit_logs SET original_actor_id = $2, actor_id = $1 WHERE actor_id = $2`. Or, less invasively, splice the original `actor_id` into each row's `meta`.

Without this, the audit log loses its primary value ("who did this?") the first time a user gets deleted.

api/src/db/database.rs
```diff
@@ -3441,6 +3567,35 @@ gitlab_id: r.user_gitlab_id,
...
+    err
+  )]
+  pub async fn upsert_service_account_token(
+    &self,
+    hash: Option<String>,
+  ) -> Result<()> {
+    let user_id = Uuid::nil();
+    let mut tx = self.pool.begin().await?;
+    sqlx::query!("DELETE FROM tokens WHERE user_id = $1", user_id)
+      .execute(&mut *tx)
```
crowlbot, line 3582: Multi-instance startup race: every replica running `upsert_service_account_token` on boot does `DELETE` then `INSERT` within a transaction. If two replicas commit concurrently, the second's `DELETE` runs after the first's `INSERT`, leaving a brief window with no service-account token. Then the second's `INSERT` puts the same row back.

End state is correct, but during a rolling restart you can see flaps where a request authenticated with the service-account token momentarily fails. Easy fix:

```sql
DELETE FROM tokens WHERE user_id = $1 AND (hash <> $hash OR $hash IS NULL);
INSERT INTO tokens (...) VALUES (...) ON CONFLICT (hash) DO NOTHING;
```

or wrap the whole thing in a `pg_advisory_xact_lock`. Probably fine to skip if rolling deploys never restart faster than the lock-step of cluster traffic, but worth knowing it's there.

api/src/main.rs
```diff
@@ -176,6 +176,13 @@ async fn main() {
   .await
   .unwrap();
 
+  database
```
crowlbot, line 179: `config.service_account_token.as_deref().map(token::hash)` — note that `Some("")` (empty-string env var) flows through as `Some(hash(""))`, which would seat the hash of an empty bearer token into the DB. That's not a security issue (no one can present an empty `Authorization: Bearer ` and have it parse), but defensively:

```rust
config.service_account_token.as_deref().filter(|s| !s.is_empty()).map(token::hash)
```

avoids the surprising state where `SERVICE_ACCOUNT_TOKEN=` (unset to empty) looks the same as `SERVICE_ACCOUNT_TOKEN=<value>` in the DB.

api/src/db/database.rs
```diff
@@ -410,17 +410,143 @@ impl Database {
...
+    // Check user exists
+    let user_exists = sqlx::query!(r#"SELECT id FROM users WHERE id = $1"#, id)
+      .fetch_optional(&mut *tx)
+      .await?;
+    if user_exists.is_none() {
+      return Ok(None);
+    }
+
+    // For scopes where the user is the last member, transfer to service account.
+    // For scopes where the user is creator but others exist, transfer creator.
```
crowlbot, line 432: Memberships are read at READ COMMITTED isolation (Postgres default). Between this `SELECT` and the writes below, another transaction could add a new `scope_member` to one of these scopes — that would mean a scope you classify as "last member" actually has another member by the time you transfer it to the service account. The new member then becomes a peer of the service account with `is_admin: true`, which is OK semantically but means the "sole owner ⇒ service account" invariant the UI promises is best-effort, not strict.

Probably fine for an admin/self-delete flow (rare, not on a hot path), but if you want strict semantics, `BEGIN ISOLATION LEVEL SERIALIZABLE` + retry would close it. Minor.

frontend/routes/account/settings.tsx
```diff
@@ -94,19 +97,22 @@ export default define.page<typeof handler>(function AccountInvitesPage({
           <h2 class="text-xl mb-2 font-bold">Delete account</h2>
           <p class="mt-2 text-secondary max-w-xl">
             You may delete your account at any time. If you delete your account,
-            any scopes that you are the sole owner of will be orphaned. You will
-            not be able to recover your account after deletion.
-          </p>
-          {/* removing delete button until we offer that functionality */}
-          <p class="mt-4 text-red-600">
-            Please contact help@jsr.io to delete your account.
+            any scopes that you are the sole owner of will be transferred to a
```
crowlbot, line 100: Re @miracatbot's comment about "orphaned" copy — this line already says "transferred to a service account", which is the new behavior. The thread is stale and can be resolved.

crowlbot (commented):
Solid PR — the design (nil-UUID service account, transactional `delete_user` with scope/audit-log reassignment, `scopes.creator` FK flip from CASCADE to SET NULL) is right. The architecture covers the main edges. A few things to address before merge, and one really important latent-bug fix that's worth calling out.

**Primary blocker — the `auth::tests::user_admin_api` test failure.** This is the test failure in CI, and it's a real inconsistency in the PR rather than a flake. The test was updated to expect that `mock_user_id = Uuid::nil()` is now the service account (`is_staff = true`, etc.), but it still PATCHes that ID expecting success. With the guard you correctly added at `api/src/api/admin.rs:96-98` (`return Err(CannotModifyServiceAccount)`), every PATCH in this test now returns 400 and the test panics on the first `assert!(resp.status().is_success())`. The fix is either:
- replace `mock_user_id` with a different (non-nil) test user — `t.user1.user.id` or fresh — and drop the `is_staff` assertions that depended on the service account state, or
- rewrite this test to assert the rejection (`expect_err_code(BAD_REQUEST, "cannotModifyServiceAccount")`), which would duplicate the negative case you already have in `admin.rs:629`.
The latter is cleaner since you already have positive coverage for staff PATCH on regular users elsewhere.

**Worth highlighting — the `scopes.creator` FK migration is fixing a latent landmine.** The previous constraint was `ON DELETE CASCADE`. Before this PR, deleting a user (only possible via a direct DB delete) would have cascaded into deleting every scope they created. That's basically a footgun that's been dormant because there was no user-deletion code path. The migration's flip to `SET NULL` is correct, and it's worth a one-line comment in the migration to make clear it's not just enabling new behavior — it's preventing a destructive default.

The two miracatbot points still standing (delete button visible on the service account admin row; the test failure described above) are both real. The third (settings page `"orphaned"` copy) is actually already addressed — line 100 of `settings.tsx` says `"transferred to a service account"`. Worth a re-read with @miracatbot to close that thread.

Rest is inline.

BlackAsLight:
Personally I'm against the idea of the audit logs being edited to replace the user's name with the service's account name. I think this completely circumvents the idea of audit logs existing as the user can just delete their account to hide their activity.

crowlbot (commented):
Reviewed the new `deletion_hold` / identity-tombstone commit — this is new scope beyond what I flagged earlier, so a fresh pass:

- Wiring is complete end-to-end: migration → `User`/`ApiFullUser` → admin PATCH → both delete paths (`self_user::delete_account` and `admin::delete_user_admin`) check the flag and return `409 userDeletionHeld` → frontend hides the delete button / shows an explanation on both the admin table and account settings page.
- Confirmed `delete_user` (`api/src/db/database.rs:489`) has exactly two callers (the two handlers above), both of which check `deletion_hold` first — no path bypasses the hold.
- The "tombstone" is minimal and matches the comment's claim: only `github_id`/`gitlab_id` get preserved, and only inside the `user_delete` audit-log `meta` (which is where the actor-preservation work from the last round of comments also lives) — the `users` row itself is a hard `DELETE`, name/email are not retained anywhere. That's a defensible design for the stated GDPR 17(3)/CCPA 1798.105(d) carve-outs.
- Good test coverage (`deletion_hold_blocks_account_deletion`): hold blocks both self-service and admin deletion, lifting it unblocks both.

One minor/optional note, same class as the READ COMMITTED point from the last round: both delete handlers check `deletion_hold` from a `User` that was loaded before the call into `delete_user` (from IAM context in the self-service path, from a separate `get_user` in the admin path), not inside `delete_user`'s own transaction. An admin flipping the hold on in the small window between that check and the delete committing would not stop the deletion. Given how rarely this fires and that you already accepted an equivalent race elsewhere as an acceptable tradeoff, I wouldn't block on it — just flagging for the record. A `SELECT ... FOR UPDATE` (or re-checking the flag inside `delete_user`'s transaction) would close it if you want it airtight.

No blockers from me. `test` job was still pending as of this review — will follow up once it lands.

piscisaureus:
I think displaying ghost on the public website as a placeholder for deleted accounts makes sense, but in the internal logs you really want to keep some user data after they delete their account.

Otherwise it becomes very difficult to investigate abuse / supply chain attacks etc. because the user can delete their account and all forensic traces with it.

Would be better to anonymize after a year instead of immediately.

crowlbot (commented):
Reviewed the response to piscisaureus's retention concern — this is a solid resolution:

- `delete_user` now retains `name`/`email` in the `user_delete` audit tombstone alongside the already-retained `github_id`/`gitlab_id`, so abuse/supply-chain investigation isn't blocked by immediate anonymization.
- A new `Database::anonymize_deleted_users(older_than)` strips just `name`/`email` via `meta - 'name' - 'email'` once the tombstone passes the cutoff, gated on `meta ? 'name' OR meta ? 'email'` so re-runs are cheap no-ops. The pseudonymous identifiers (internal user ID, github_id/gitlab_id) are kept indefinitely, matching the original GDPR/CCPA-carve-out design.
- Wired to a new `/tasks/anonymize_deleted_users` endpoint, mounted the same way as the other `/tasks/*` handlers (only exposed when `expose_tasks` is set — internal-only, not publicly routable) — no new attack surface relative to its siblings.
- Scheduled daily via Cloud Scheduler with the same OIDC-token-to-service-account pattern as the existing `requeue_stuck_publishing_tasks` job.
- Test coverage checks both sides of the cutoff boundary (`now - 1 day` → 0 anonymized, `now + 1 day` → 1 anonymized) and confirms `github_id`/`user_id` survive while `name`/`email` are gone afterward.

This directly addresses the forensics-vs-erasure tension piscisaureus raised: a year-long retention window for name/email, permanent retention of pseudonymous IDs. Nothing further from me on the mechanism — the "one year" duration itself is a policy call for you two to settle, not something I can weigh in on technically. CI build passed, `test` job was still pending at review time.

crowlbot (approved):
Formal approval after review request — no new commits since my last pass (still `590a9e6`), CI fully green.

Summary of what's been verified across this review:
- The original 5 points (nil-UUID admin-mutability test, migration warning comment, audit-log actor attribution, multi-instance token-upsert race, empty-token defensiveness) are all fixed.
- The `deletion_hold` flag correctly gates both self-service and admin deletion, with no bypass path (`delete_user` has exactly two callers, both check the flag), and the frontend hides/explains accordingly.
- In response to piscisaureus's forensics concern: deleted-user tombstones now retain name/email for a 1-year window (anonymized after via a properly-gated internal-only cron task, boundary-tested), while pseudonymous identifiers (user ID, github_id/gitlab_id) are kept indefinitely — a solid balance between GDPR/CCPA erasure obligations and abuse-investigation needs.

One remaining non-blocking note from earlier (both delete-path handlers check `deletion_hold` from a `User` loaded before entering `delete_user`'s own transaction, not inside it) — a narrow TOCTOU race, same class as an already-accepted tradeoff elsewhere in this PR. Not blocking.

Approving.

PR #1339, opened 2026-03-10
chore: remove unused legacy infra

crowlKats:
lets wait at least to end of week to land this.

The buckets will be orphaned, so we will keep them around a while longer just as a precaution.

4h downloads is removed in favour of a broader analytics engine query that runs less frequently. keeps similar behaviour than before while removing one the largest tables we have.


piscisaureus:
`google_compute_ssl_policy` and `google_compute_managed_ssl_certificate` are kept in `https.tf`, but the forwarding rules and HTTPS proxy that referenced them are deleted. Are these still needed? If not, they should be cleaned up too to avoid terraform drift.

piscisaureus:
The Analytics Engine queries change from `INTERVAL '4' HOUR` to `INTERVAL '1' DAY`, and the insert now goes directly into `version_download_counts_24h` with `SUM(temp.count)` + `GROUP BY`. Is this change in behavior intentional?

The scheduler now runs every 3h with `intervalHrs=24`, and the upsert uses `ON CONFLICT ... DO UPDATE SET count = EXCLUDED.count` (replace, not accumulate). This means the last successful scrape of the day wins. If Cloudflare Analytics Engine returns cumulative daily counts that's fine, but if it returns counts scoped to just the query window, overlapping runs could overwrite a full day's count with a partial one.

Previously the 4h table accumulated granular buckets and then aggregated into the 24h table, which avoided this issue.

piscisaureus:
Potential bug in `api/src/util.rs`: `ensure_servers_started` now uses `std::thread::spawn` instead of the previous `std::thread::scope`. The `OnceLock::get_or_init` closure spawns the thread and returns immediately, so `FakeS3Tester::new()` may not have completed when tests start using it.

Previously `thread::scope` was used which joined both the GCS and S3 threads before returning. Now that only S3 remains, this could be simplified to just calling `FakeS3Tester::new()` directly inside the `get_or_init` closure (no thread needed).

crowlKats:
ok readded the 4h table, was really hoping we could remove it. well, i added a cleanup job for it so it doesnt balloon in size.

piscisaureus (approved):

PR #1345, opened 2026-03-18
fix: read  from correct database column in User/UserPublic FromRow impls

ayushkavad:
Both User::from_row() and UserPublic::from_row() were reading for the  field, causing every API response containing user data to silently return the account creation timestamp instead of the last-modified timestamp. This affected admin user listings, scope members, scope invites, publishing tasks, and ticket responses.


crowlKats (approved):
LGTM; thanks!

PR #1364, opened 2026-04-03
feat: add search filtering UI, score filter, and hide yanked-only packages

### LINKED ISSUE #287: suggestion: filter by 100% score
It doesn't have to be a full score specifically, but rather some way to filter based on score. In another form we could also sort by score in another form. However, filtering would be more straightforward. Either way, this would make looking for high-quality packages easier.

### LINKED ISSUE #311: Search ranks unpublished packages first
It appears that search ranks unpublished modules at the top. For example:

https://jsr.io/packages?search=kv

Returns:

kv_-_Packages_-_JSR

Where the "top" package is currently unpublished: https://jsr.io/@ts-rex/kv-spy

### LINKED ISSUE #312: request: Better Module Search
Can we have a better module searching feature on JSR? Something like the search used by image boards, discord, etc.

**Example**
`websocket` - Normally searches for `websocket`

`runtime:deno websocket` - Searches for `websocket` that runs on Deno.

`runtime:deno runtime:node websocket` - Searches for `websocket` that runs on both Deno and Node.

`scope:std websocket` - Searches for `websocket` under the `std` scope.

This would be even more useful if concepts like `tags` (eg. Markdown, ML, Crypto) were introduced for libraries.

### LINKED ISSUE #689: Don't show packages in search results that only have yanked versions

crowlKats:
## Summary
- **Search filter UI** (#312): Adds a collapsible filter panel on the `/packages` page with runtime compatibility toggle buttons (Deno, Node.js, Browsers, Cloudflare Workers, Bun) and a minimum score slider, providing a proper UI alternative to text-based filter syntax
- **Score filter syntax** (#287): Adds `score:>N`, `score:>=N`, `score:<N`, `score:<=N` support in the search query syntax (e.g. `score:>80` to find high-quality packages)
- **Hide yanked-only packages** (#689): Packages where all versions are yanked are now removed from the search index, both in the API's real-time Orama upsert and in the reindex script
- **Fix unpublished package ranking** (#311): The reindex script's filter logic was using `||` instead of `&&`, causing nearly all packages (including archived/internal ones) to pass through — this is now fixed along with the yanked-version exclusion

Closes #287, closes #311, closes #312, closes #689

## Test plan
- [ ] Verify `score:>80` in search bar highlights as a filter tag and correctly filters results
- [ ] Verify `score:>=90 runtime:deno` combined filters work
- [ ] On `/packages`, click "Filters" button to expand the filter panel
- [ ] Toggle runtime buttons and verify results update with correct URL params
- [ ] Drag the minimum score slider and verify filtering works
- [ ] Verify "Clear filters" link resets all filters
- [ ] Verify pagination preserves filter state
- [ ] Verify packages with only yanked versions no longer appear in search results
- [ ] Verify the FAQ docs page includes the new score filter documentation


KnorpelSenf:
Should the score be part of the ranking by default, rather than just letting users filter by score optionally?

crowlKats:
do you mean that the searchbar by default should filter by a score of like 70+?

KnorpelSenf:
No, I mean that the ranking should be influenced. Packages with better scores should appear first.

crowlKats:
we already set a value called _omc:number (or similar), which affects the search results based on the score

KnorpelSenf:
Ohh I see, never mind then. I felt like I always get packages with a low score very high, but then I'll try this a bit more systematically before making such suggestions lol

crowlbot (commented):
Strong PR — the bundling of #312 (UI), #287 (score syntax), #689 (yanked-only exclusion), and #311 (reindex bug) makes sense because they all touch the same code path, and the #311 catch is good (the old `||` chain was effectively letting everything through). A few things to address:

1. **`deno fmt --check` is red** on 2 frontend files. The two stray empty lines added in `GlobalSearch.tsx` (lines 79 and 104, both `  ` with trailing whitespace) are part of it — `deno fmt frontend/` should resolve. 
2. **The FAQ and PR description promise a *slider*, but the implementation is 5 discrete buttons** (`SCORE_OPTIONS = [Any, 60+, 70+, 80+, 90+]`). They're nice and tappable on mobile, but the docs are wrong as written. Either update the FAQ to "buttons" / "chips" or swap the buttons for an actual `<input type="range">`. The test plan checkbox "Drag the minimum score slider" also can't be exercised as written.
3. **`upsert_package` and the reindex now disagree about archival.** In `orama.rs:70`, archived packages cause a silent `return` (no `delete_package` call), but the reindex now *excludes* archived packages from the upsert batch. So if a package becomes archived after being indexed, realtime leaves the stale record in Orama until the next full reindex. Pre-existing — but this PR makes it more visible by fixing the reindex side. Worth either deleting on archive-transition in the realtime path or accepting that reindex is the reconciliation step.

The KnorpelSenf thread (default score-based ranking) is already resolved — `_omc:number` is doing that work. Nothing else needed there. Rest is inline.

tools/orama_packages_reindex.ts
```diff
@@ -48,8 +48,9 @@ while (true) {
 
 const entries = packages
   .filter((entry) =>
-    entry.versionCount > 0 || !entry.isArchived ||
-    !entry.description.startsWith("INTERNAL")
+    entry.versionCount > 0 && !entry.isArchived &&
+    !entry.description.startsWith("INTERNAL") &&
```
crowlbot, line 52: Good catch on the `||` → `&&` fix — that's #311. The old version short-circuited as soon as any one of the conditions was true, so basically every non-empty-versions package passed (archived + INTERNAL included). Just confirming the new filter covers everything `upsert_package` would reject:

| Reject condition | `orama.rs` realtime | reindex (post-PR) |
|---|---|---|
| `versionCount == 0` | silent return | `versionCount > 0` ✓ |
| `is_archived` | silent return | `!isArchived` ✓ |
| `latest_version is None` (all yanked) | `delete_package` | `latestVersion !== null` ✓ |
| `description starts with INTERNAL` | `delete_package` | `!startsWith("INTERNAL")` ✓ |

So the reindex is now a strict superset of "things that would survive a realtime upsert." Nice.

api/src/external/orama.rs
```diff
@@ -70,6 +70,12 @@ impl OramaClient {
       return;
     }
 
+    // Don't index packages where all versions are yanked
+    if package.latest_version.is_none() {
+      self.delete_package(&package.scope, &package.name);
```
crowlbot, line 79: Question on the realtime path generally (not introduced here, but worth thinking about while you're in this file): `version_count == 0 || is_archived` returns silently without calling `delete_package`. So an *already-indexed* package that flips to `is_archived = true` will stay in Orama until the next full reindex. With the reindex now properly excluding archived packages (#311), the gap is no longer a steady-state inconsistency, but it does leave a window between archival and the reindex run where the search index serves stale results.

If this matters, the symmetrical fix is:

```rust
if package.is_archived {
  self.delete_package(&package.scope, &package.name);
  return;
}
if package.version_count == 0 {
  return;
}
```

(version_count == 0 doesn't need to delete because it can never have been indexed in the first place — the same `upsert_package` early-returned on that condition before any version was published.)

frontend/islands/GlobalSearch.tsx
```diff
@@ -21,14 +22,14 @@ interface GlobalSearchProps {
...
-const searchHints: JSX.Element[] = [
-  <p key="scope:">
-    Hint: use <code>scope:</code> to search for packages by scope
-  </p>,
-  <p key="runtime:">
-    Hint: use <code>runtime:</code> to search for packages by compatible runtime
-  </p>,
-];
+
+const SCORE_OPTIONS = [
```
crowlbot, line 25: Re: "slider" in PR description and FAQ — two options: keep the chip buttons but update the FAQ (line 154–155 in `frontend/docs/faq.md`) to say "chips/buttons", or implement a real slider. The buttons are arguably better UX on mobile and they snap to clean percentages, so I'd vote for adjusting the docs.

frontend/islands/GlobalSearch.tsx
```diff
@@ -74,8 +76,7 @@ export function GlobalSearch(
     }
   }, [projectId, apiKey]);
 
-  const randomHint = useSignal<JSX.Element | null>(null);
-
+
```
crowlbot: Stray blank line with trailing whitespace — this and the matching one at line 104 are what `deno fmt --check` is rejecting.

frontend/islands/GlobalSearch.tsx
```diff
@@ -260,12 +260,51 @@ export function GlobalSearch(
...
+    }
+    triggerSearch(newValue);
+  }
+
+  function setScoreFilter(scoreValue: number | null) {
+    const tokens = tokenizeFilter(search.value);
+    const withoutScore = tokens.filter((t) => t.kind !== "score");
+    let newValue = withoutScore.map((t) => t.raw).join(" ");
+    if (scoreValue !== null) {
+      newValue = (newValue + " score:>=" + scoreValue).trim();
```
crowlbot, line 279: `setScoreFilter` always writes `score:>=N` (gte). If the user typed `score:>80` in the search bar (gt), then clicks the 80+ button, the existing token gets wiped and replaced with `score:>=80` — i.e. the operator silently changes from `>` to `>=`. Minor, but worth noting that the buttons are not a no-op when the existing filter matched semantically (in particular, `score:>79` and `score:>=80` are the same constraint but only one would highlight the 80+ chip).

frontend/islands/GlobalSearch.tsx
```diff
@@ -534,6 +713,24 @@ function tokenizeFilter(search: string): Token[] {
         });
         continue;
       }
+    } else if (part.startsWith("score:")) {
+      const rest = part.slice(6);
+      const match = rest.match(/^(>=|<=|>|<)(\d+)$/);
```
crowlbot, line 718: The regex `^(>=|<=|>|<)(\d+)$` rejects decimals and negatives. `score:>=80.5` falls through to a `text` token and ends up in the query string instead of the where clause, which means it pollutes the full-text search rather than failing visibly. Probably fine — scores in JSR are integers — but worth a one-liner comment that decimals are intentionally rejected, or you can clamp/round to int as a forgiving fallback.

frontend/islands/GlobalSearch.tsx
```diff
@@ -542,21 +739,31 @@ function tokenizeFilter(search: string): Token[] {
...
-  const filters: [string, boolean | string][] = [];
+): { query: string; where: WhereClause | undefined } {
+  const where: WhereClause = {};
   let query = "";
+  let hasFilters = false;
   for (const part of tokenizeFilter(search)) {
     if (part.kind === "text") {
       query += part.value + " ";
+    } else if (part.kind === "score") {
+      where["_omc:number"] = {
```
crowlbot, line 755: Spreading `where["_omc:number"]` like this means two conflicting score filters merge instead of overwrite: `score:>80 score:<90` produces `{ gt: 80, lt: 90 }` (the right thing). But `score:>80 score:>50` produces `{ gt: 50 }` because the later token's `gt` key clobbers the earlier one (later spread wins). That's also fine semantically (the user said `>50` last, so honor that), I just want to confirm it's the intended behavior since this codepath also gets exercised by the chip buttons when a user fat-fingers two clicks.

frontend/islands/GlobalSearch.tsx
```diff
@@ -260,12 +260,51 @@ export function GlobalSearch(
     }
   }
 
+  function toggleFilter(tokenRaw: string) {
```
crowlbot, line 260: `toggleFilter` does an exact `tokenRaw` match (e.g. `runtime:deno`). If a user typed `runtime:Deno` (capital D) into the search bar, the tokenizer at line 421-ish doesn't normalize case, so it becomes a runtime token with `value: "Deno"` and `raw: "runtime:Deno"`. Then clicking the Deno chip appends `runtime:deno` (lowercase, from `RUNTIME_COMPAT_KEYS`), leaving both tokens in the box. Easy fix: lowercase the value when tokenizing the runtime branch, so the chip toggle is case-insensitive.

api/src/external/orama.rs
crowlKats: Applied the symmetrical fix in d5019af — archived packages now call `delete_package` then return, so the realtime path matches the reindex.

frontend/islands/GlobalSearch.tsx
crowlKats: Updated the FAQ to say "score chips" in d5019af.

frontend/islands/GlobalSearch.tsx
crowlKats: Fixed by `deno fmt` in d5019af.

frontend/islands/GlobalSearch.tsx
crowlKats: Good catch. Fixed in d5019af — the tokenizer now lowercases the runtime value and normalizes `raw` to `runtime:<lowercase>`, so `runtime:Deno` + clicking the Deno chip toggles cleanly instead of double-adding.

tools/orama_packages_reindex.ts
crowlKats: Thanks for the table — confirms my read.

frontend/islands/GlobalSearch.tsx
crowlKats: Noted — leaving as-is for now since the operator drift is harmless (both are equivalent constraints when N is an integer). Happy to revisit if it actually trips someone up.

frontend/islands/GlobalSearch.tsx
crowlKats: Leaving as-is — scores are always integers, so decimals falling through to text isn't worth a defensive branch.

frontend/islands/GlobalSearch.tsx
crowlKats: Confirmed — "last write wins" is intended. The chip buttons always clear existing score tokens before appending, so a double-click can't produce a conflicting pair through the UI; this only applies when users hand-edit the bar.

crowlbot (approved):
All concerns from the prior review are addressed in d5019af:

- **`deno fmt`**: CI green.
- **FAQ**: now says "score chips" (matches the actual UI). ✅
- **Runtime tokenizer case normalization**: `part.slice(8).toLowerCase()` + normalized `raw` — chip toggle is now case-insensitive. ✅
- **Archived-package realtime/reindex symmetry** (`orama.rs`): archived packages now call `delete_package` before returning, and the new `latest_version.is_none()` guard cleans up yanked-all-versions packages too. The realtime path and the reindex now agree on every reject condition. ✅
- Decimal scores, operator drift, last-write-wins — all acknowledged and intentional.

The Orama integration, search tokenizer, and filter-bar UI all look solid. Ship it.

crowlKats:
Rebased onto main across the Algolia search migration (#1461). Since two of the three commits reworked the same UI, the branch is now a single squashed commit. Notable port decisions:

- `processFilter` now emits Algolia filter strings (`score >= 80`, `scope:"std"`, `runtimeCompat.deno:true`) instead of Orama where-clauses; the `_omc:number` score field is now the indexed `score` field.
- The yanked-only exclusion moved from `orama.rs` to `algolia.rs::upsert_package`; the reindex-script OR→AND fix carried over to `tools/algolia_packages_reindex.ts` via rename detection.
- Kept main's rotating search hints (with a new `score:>N` hint) instead of the static hint block, since the Filters toggle is now visible in the popup footer.
- Kept main's Algolia insights click-tracking alongside the new filter props in `SuggestionList`.

cargo check, deno fmt/lint, frontend tests, and the frontend build all pass locally. Worth a re-skim of the filter-string translation before merging since the prior approval predates the Algolia port.

PR #1365, opened 2026-04-03
fix: UI improvements across navbar, docs, downloads, and homepage

### LINKED ISSUE #600: Improve the Packages section
Looking at the Packages section, I have no idea what those listed packages do. I have to click each link to read the description.

Suggestions: 
- Add a short description for each package
- May split the section into 3 sections: featured, recently updated, and new

 image

### LINKED ISSUE #692: Add explanation of “works with” info
“Works with” info is represented using a cluster of icons, many of which will be unfamiliar to some users. (How many people recognize the Bun icon?)

It would be nice if clicking on it showed the same information as a list of names, perhaps with links to more information about each environment.

### LINKED ISSUE #1059: Empty graph after unarchiving a package
I had this package archived and unliked a couple of minutes ago, it displayed the weekly downloads. But now when I have it unarchived and linked to a repo,and want to see the complete graph it looks like these (the empty graph image)

### LINKED ISSUE #1079: Weird alignment in navbar elements of small devices
Image

### LINKED ISSUE #1089: add links to the docs page on a scope package list

### LINKED ISSUE #1099: In case of no default entrypoint, display a text indicating this situation

### LINKED ISSUE #1121: Package description overflows on wrap
This packages description seems to overflow off the screen on mobile

https://jsr.io/@salmidev/maths-objects
image

### LINKED ISSUE #1130: Weekly downloads shows really low number until previous week is hovered over
https://jsr.io/@david/dax

By default:

Image

Hovered:

Image

crowlKats:
## Summary

- **#1079**: Shorten "Browse packages" to "Packages" on small screens to prevent navbar items from wrapping/misaligning on mobile devices
- **#1089**: Add documentation links to package list items in scope and browse pages, with a new `actions` slot in `ListDisplayItem` to support sibling links alongside the main item link
- **#1121**: Add `break-words` class to package description on the package detail page to prevent long text from overflowing on mobile
- **#1130**: Drop the current incomplete week from download chart data so it doesn't appear in the graph or show misleadingly low numbers
- **#1059**: Show "No download data available yet" empty state in the download chart when package has no download data (e.g. after unarchiving)
- **#1099**: Show a notice on the all_symbols docs page when a package has no default entrypoint, instead of silently redirecting
- **#692**: Already addressed — RuntimeCompatIndicator already has title tooltips on each runtime icon explaining compatibility
- **#600**: Add package descriptions to the homepage Featured, Recently updated, and New to JSR sections (API + frontend changes including sqlx cache updates)

Closes #1079
Closes #1089
Closes #1121
Closes #1130
Closes #1059
Closes #1099
Closes #692
Closes #600

## Test plan

- [x] Verify navbar alignment on mobile viewport (<640px) — "Packages" label should display without wrapping
- [x] Verify "Browse packages" still shows on screens ≥640px
- [x] Verify docs link icon appears on package list items in scope pages and browse packages page
- [x] Verify clicking the docs link navigates to `/@scope/name/doc`
- [x] Verify package descriptions with long unbreakable text wrap correctly on mobile
- [x] Verify default weekly downloads number shows the last complete week's count
- [x] Verify hovering over chart points still shows the correct week's data
- [x] Verify download chart shows empty state message when no data available
- [x] Verify visiting `/doc` on a package with no default entrypoint shows a notice on the all_symbols page
- [x] Verify homepage package lists show descriptions below package names
- [x] Verify runtime compat icons show tooltips on hover (pre-existing)


crowlbot (commented):
Eight unrelated issues in one PR makes review harder than it needs to be — if any one of these turns out controversial, the other seven get stuck behind it. Not a blocker, but worth considering whether the `actions`-slot refactor (#1089) and the homepage descriptions (#600) belong in their own PRs since they each touch shared components.

Three things that should land before merge:

1. **`deno fmt --check` is red** on `frontend/components/List.tsx` (line 54 — the chevron `<a>` attributes are over the wrap threshold). One `deno fmt frontend/components/List.tsx` and commit fixes the `frontend` job.
2. **Mobile docs link is icon-only with no a11y label.** `<a title="Documentation">` falls back to a `title` tooltip, which screen readers don't reliably announce. `aria-label="Documentation"` on the `<a>` will fix it without touching layout.
3. **`DownloadChart` and `DownloadWidget` disagree about "no data"** now that `.slice(0, -1)` drops the current week in some places but not others. Worth picking one source of truth so the chart and the widget never show different empty-states for the same package.

Rest is inline.

frontend/components/List.tsx
```diff
@@ -34,16 +35,39 @@ export function ListDisplay(
...
+                  <a
+                    href={item.href}
+                    class="flex-1 flex items-center gap-2 min-w-0 outline-none focus:ring-2 ring-jsr-cyan-700 dark:ring-jsr-cyan-500 ring-inset rounded"
+                  >
+                    {item.content}
+                  </a>
+                  <div class="flex items-center gap-2 shrink-0">
+                    {item.actions}
+                  </div>
+                  <a href={item.href} class="shrink-0" aria-hidden="true" tabindex={-1}>
```
crowlbot: This is what `deno fmt --check` is rejecting. Running `deno fmt frontend/components/List.tsx` rewrites it as:

```tsx
<a
  href={item.href}
  class="shrink-0"
  aria-hidden="true"
  tabindex={-1}
>
```

Once that lands the `frontend` CI job goes green.

frontend/components/List.tsx
```diff
@@ -34,16 +35,39 @@ export function ListDisplay(
...
+                    <TbChevronRight class="text-jsr-cyan-800 dark:text-jsr-cyan-400 size-6" />
+                  </a>
+                </div>
+              )
+              : (
+                <a
```
crowlbot, line 65: Two small things on this variant:

1. The main `<a>` here uses `focus:ring-2 ring-jsr-cyan-700 ... rounded`, but the non-`actions` variant just below uses `focus:bg-jsr-cyan-50` *and* the ring. Two different focus styles for what is otherwise the same list row will feel jumpy when you tab through a list that mixes both. Pick one (probably the ring on its own, since the background-flash interacts oddly with the chevron column).
2. Wrapping `<a aria-hidden="true" tabindex={-1}>` around the chevron is fine for keyboard/SR users (they skip it), but mouse users still get a clickable chevron with no hover affordance. Consider letting the outer `hover:bg-jsr-cyan-50` cover it (which it does today), but also add a subtle scale or color hover on the chevron so it doesn't look dead.

frontend/components/PackageHit.tsx
```diff
@@ -63,5 +64,15 @@ export function PackageHit(pkg: OramaPackageHit | Package): ListDisplayItem {
         )}
       </div>
     ),
+    actions: (
+      <a
+        href={`/@${pkg.scope}/${pkg.name}/doc`}
+        class="flex items-center gap-1 text-sm text-secondary hover:text-jsr-cyan-700 dark:hover:text-jsr-cyan-400 px-2 py-1 rounded hover:bg-jsr-cyan-100 dark:hover:bg-jsr-cyan-900 transition-colors"
+        title="Documentation"
+      >
+        <TbFileDescription class="size-5" />
```
crowlbot, line 74: On screens < `sm`, this link is icon-only — `<span class="hidden sm:inline">Docs</span>` removes the visible label. The fallback is `title="Documentation"`, which gives a desktop tooltip but is not reliably announced by screen readers (and not surfaced on touch at all).

```tsx
<a
  href={`/@${pkg.scope}/${pkg.name}/doc`}
  aria-label="Documentation"
  title="Documentation"
  class="..."
>
```

Keeps the tooltip for mouse users and gives assistive tech a name.

frontend/routes/package/(_islands)/DownloadChart.tsx
```diff
@@ -93,12 +93,15 @@ const getChartOptions = (
 });
 
 export function DownloadChart(props: Props) {
+  const hasData = props.downloads.some((v) => v.downloads.length > 0);
```
crowlbot: `hasData` is computed off the raw `props.downloads`, but `DownloadWidget.tsx` and `PackageHeader.tsx` both drop the last (incomplete) week via `.slice(0, -1)` before deciding what to render. So a package whose *only* downloads are in the current incomplete week:

- `DownloadChart` here ⇒ `hasData === true` ⇒ renders the chart (with one bar)
- `DownloadWidget` after slice ⇒ empty `data` ⇒ silently renders nothing useful (`min` stays `Infinity`)
- `PackageHeader` after slice ⇒ `weeklyDownloads === null`

Probably want a single helper like `hasCompletedWeekData(downloads)` shared by all three so the chart, the widget, and the header always agree.

frontend/routes/package/(_islands)/DownloadWidget.tsx
```diff
@@ -21,7 +21,9 @@ export function DownloadWidget(props: Props) {
   const chartRef = useRef<HTMLDivElement>(null);
 
   const xValues = collectX(props.downloads, AGGREGATION_PERIOD);
-  const data = normalize(props.downloads, xValues, AGGREGATION_PERIOD);
+  // Drop the last (current, incomplete) week so the chart only shows full weeks
+  const data = normalize(props.downloads, xValues, AGGREGATION_PERIOD)
+    .slice(0, -1);
```
crowlbot, line 28: No length guard after `.slice(0, -1)`. If `data` is empty (package with downloads only in the current week), the loop below leaves `min = Infinity` and `max = 0`, and the SVG path math downstream will produce a degenerate result. Easiest fix:

```ts
if (data.length === 0) {
  return <div class="text-xs text-tertiary">No data yet</div>;
}
```

at the top, after the slice.

frontend/routes/package/(_components)/PackageHeader.tsx
```diff
@@ -37,11 +37,11 @@ export function PackageHeader({
     (pkg.latestVersion === null ||
       greaterThan(selectedVersionSemver, parse(pkg.latestVersion)));
 
-  // Calculate weekly downloads for mobile display
+  // Calculate weekly downloads for mobile display (drop incomplete current week)
   let weeklyDownloads: number | null = null;
   if (downloads && downloads.total.length > 1) {
     const xValues = collectX(downloads.total, "weekly");
```
crowlbot, line 49: Worth a sanity-check: before this change, a brand-new package with only the current week's data showed *some* `weeklyDownloads` number. After the slice, it will show `null` (which I'm guessing renders as a placeholder). That's the right thing to do — we don't have a complete week to report — just confirming the design intent matches what the rest of #1130 wants.

frontend/routes/package/doc/all_symbols.tsx
```diff
@@ -30,6 +31,13 @@ export default define.page<typeof handler>(function AllSymbols(
         latestVersion={data.package.latestVersion}
       />
 
+      {noDefaultEntrypoint && (
+        <div class="mt-4 rounded border border-jsr-cyan-200 dark:border-jsr-cyan-800 bg-jsr-cyan-50 dark:bg-jsr-cyan-950 py-3 px-4 text-sm text-secondary">
```
crowlbot, line 35: Reading state from a query param means the URL is shareable / bookmarkable: someone who copies `?no_default_entrypoint` will keep seeing this banner even after the package adds a default entrypoint in a later version. Minor — most users won't notice — but consider either:

- determining "has a default entrypoint?" server-side from `data.package` and dropping the query param entirely, or
- making the redirect in `data.ts` only attach the param when *this version specifically* has no default entrypoint, accepting the shareable-URL artifact.

frontend/utils/data.ts
```diff
@@ -133,7 +133,7 @@ export async function packageDataWithDocs(
             headers: {
               Location: `/@${scope}/${pkg}${
                 version ? `@${version}` : ""
-              }/doc/all_symbols`,
+              }/doc/all_symbols?no_default_entrypoint`,
```
crowlbot, line 136: Pairs with the banner in `all_symbols.tsx`. The redirect already knows whether the package version has a default entrypoint, so a cleaner design is to set a server-side flag on the page data instead of round-tripping through a query string — but that's a bigger refactor. Fine to ship as-is if you accept the bookmarkable-URL quirk noted on the other file.

api/src/db/models.rs
```diff
@@ -907,13 +907,15 @@ pub type PackageWithGitHubRepoAndMeta =
 pub struct StatsPackage {
   pub scope: ScopeName,
   pub name: PackageName,
+  pub description: String,
```
crowlbot, line 910: `description: String` (not `Option<String>`) is correct here — `packages.description` is `NOT NULL DEFAULT ''` (`20230825084411_package_description.sql`). Just flagging because the sqlx output also lists it as `nullable: false`, so this is consistent. No change needed.

frontend/components/ListPanel.tsx
```diff
@@ -36,20 +37,27 @@ export function ListPanel(
...
+                  <span class="block group-hover:text-jsr-cyan-800 dark:group-hover:text-cyan-300 pr-4 flex-1 group-hover:underline truncate">
+                    {entry.value}
+                  </span>
+                  {entry.label && (
+                    <div class="chip bg-jsr-cyan-200 dark:bg-jsr-cyan-950 max-w-20 truncate">
+                      {entry.label}
+                    </div>
+                  )}
+                </div>
+                {entry.description && (
```
crowlbot, line 57: Good — `{entry.description && ...}` correctly skips empty-string descriptions (which is the DB default). Just be aware that `truncate` will cut at one line; for the homepage panels that's probably what you want, but a quick visual check on a description like `"A really really really really really really really really really really long description"` against the narrow panel widths is worth doing as part of the test plan.

crowlbot (approved):
All three blocking items from my review are addressed in 83a23c8:

1. **`deno fmt`**: CI green. ✅
2. **`aria-label` on mobile docs link**: `PackageHit.tsx` now has both `aria-label="Documentation"` and `title="Documentation"`. ✅
3. **Download chart/widget/header "no data" disagreement**: Shared `hasCompletedWeekData()` helper in `DownloadChart.tsx`, used consistently by `DownloadWidget`, `PackageHeader`, and `DownloadChart` itself. Empty-state guard added to `DownloadWidget` (returns `null`) and `DownloadChart` (renders placeholder message). The `.slice(0, -1)` to drop the incomplete current week is now applied consistently. ✅

Additional improvements I see:
- `break-words` on description paragraph — good for long unbroken strings
- Homepage panels now show package descriptions via the API — nice touch
- `chart?.destroy()` optional chaining prevents the error when `hasData` is false (the `useEffect` returns early without creating a chart)

Minor residual: the `actions` variant in `List.tsx` uses only `focus:ring-2` on the inner `<a>`, while the non-`actions` variant also applies `focus:bg-jsr-cyan-50`. Barely noticeable in practice since both branches share the same hover/ring style. Not blocking.

Ship it.

PR #1366, opened 2026-04-03
docs: improve documentation for slow types, publishing, packages, and badges

### LINKED ISSUE #290: missing unsupported-nested-javascript documentation
In a dry run I get

```
error[unsupported-nested-javascript]: referenced a JavaScript module without type declarations from a TypeScript module
 --> # …
  = hint: add a type declaration (d.ts) for the JavaScript module, or rewrite it to TypeScript

  info: JavaScript files with no corresponding declaration require type inference to be type checked
  info: fast check avoids type inference, so referencing a JavaScript file with no type declarations is not supported
  docs: https://jsr.io/go/slow-type-unsupported-nested-javascript
```

The link <https://jsr.io/go/slow-type-unsupported-nested-javascript> redirects to <https://jsr.io/docs/about-slow-types>, which doesn't have anything explicit about nested types.

Compare to the `missing-explicit-type` error, which links to <https://jsr.io/go/slow-type-missing-explicit-type>, which redirects to the relevant <https://jsr.io/docs/about-slow-types#explicit-types>

--

Tangent: if anyone reads this has links or recommendations for resolving the `unsupported-nested-javascript` I'd gladly take them. Context: I'm importing [`tailwindcss/plugin.js`](https://github.com/tailwindlabs/tailwindcss/blob/3.4/plugin.js), which is typed in [`tailwindcss/plugin.d.ts`](https://github.com/tailwindlabs/tailwindcss/blob/3.4/plugin.d.ts), with `import plugin from 'tailwindcss/plugin.js'`.

### LINKED ISSUE #370: d.ts files are getting ignored when entrypoint is a js file
which leads to jsr resorting to trying to infer types trivially, and as a result - failure to use the library altogether.

## mcve

deno.json
```json
{
    "name": "@foo/bar",
    "version": "1.3.5",
    "exports": "./mod.js"
}
```

mod.js
```js
const exports = {};

exports.ns = {};
(function(ns) {
    ns.bar = "egg";
})(exports.ns);

export const ns = exports.ns;
```

mod.d.ts
```ts
export declare namespace ns {
    const bar = "egg";
}
```

> the above is similar to commonjs output from tsc for typescript namespaces

## expected result

jsr uses d.ts files for typings and doesn't try to infer them on its own

## actual result

jsr warns about "slow types" when publishing and advises to add d.ts files (which are already there):
```log
warning[unsupported-javascript-entrypoint]: used a JavaScript module without type declarations as an entrypoint
 --> /path/to/mod.js
  = hint: add a type declaration (d.ts) for the JavaScript module, or rewrite it to TypeScript

  info: JavaScript files with no corresponding declaration require type inference to be type checked
  info: fast check avoids type inference, so JavaScript entrypoints should be avoided
  docs: https://jsr.io/go/slow-type-unsupported-javascript-entrypoint
```
> btw, linked docs might be out of date? there's nothing mentioned about js entrypoints

## additional context

due to this bug, `export const ns` from the mcve, whose type can't be trivially inferred, results being unusable at all:

```js
// some other package
import { ns } from '@foo/bar'

console.log(ns.bar) // errors here, ns is not a namespace
```

### LINKED ISSUE #435: Document how to publish from other CI providers
For example GitLab CI or Sourcehut. This requires #393.

### LINKED ISSUE #509: no-slow-types - Add documentation on unsupported-expando-property
* https://github.com/denoland/deno_graph/pull/428/files#diff-bfee9a1b96f39956d569df6c610e0e8b4eb5fb38ea127a03cd0abc38dea7c7e3R269
* https://github.com/denoland/deno_graph/pull/428#discussion_r1597140066

Might not be worth doing though because this is just a temporary error until we can improve it.

### LINKED ISSUE #639: Unpublished packages should not return a 404
The API should be able to differentiate between a missing package and an empty package.

If you curl the API on a created package but that has no release yet, it shouldn't return a 404:
```text
404 - Not Found
```
But rather a JSON similar to this one, just with no versions and maybe latest set to null
```json
{
  "scope": "scope",
  "name": "package",
  "latest": null,
  "versions": {}
}
```

Also on a related note, using the version badge with an unreleased package should also be updated accordingly to inform the package exist but isn't released yet

### LINKED ISSUE #682: Document how to create a pre-release?
JSR automatically distinguishes SemVer-compliant pre-release versions from normal versions and treats them like yanked versions, but this specification may not always be obvious to new users.

Since this does not currently seem to be mentioned in the documentation, why not add this?

### LINKED ISSUE #767: docs: the "publishing packages" page has no example 'exports' field with object syntax
## What I see 

Every Example on the page https://jsr.io/docs/publishing-packages, 
and especially in the section [package-config-file](https://jsr.io/docs/publishing-packages#package-config-file) 
only has examples like 

```
  "exports": "./mod.ts"
```

## What I want 

An example, which shows how to use "exports" with object syntax. 
Especially I want to know: 
How to write the main import in object syntax? 
Do I write it like this? 

```
"exports": {
  "/": "./mod.ts"
  "/extra": "./extra/mod.ts"
}
```

Or like this? 

```
"exports": {
  ".": "./mod.ts"
  "extra": "./extra/mod.ts"
}
```

Or something completely different?

### LINKED ISSUE #1063: Diagnostic "ExportNotFound" has no documented section via https://jsr.io/go/slow-type-export-not-found
While encountering an "ExportNotFound" error (which I fixed since I had a typo in a path within deno.json's exports field), it printed out the following message:

```
  = hint: this error is the result of a bug in Deno and you don't be seeing it - please open an issue if one doesn't exist

  info: this error is the result of a bug in Deno and you don't be seeing it - please open an issue if one doesn't exist
  docs: https://jsr.io/go/slow-type-export-not-found
```

The problem I found is that there's no documented section on https://jsr.io/go/slow-type-export-not-found.

I wasn't sure if I should've opened an issue in https://github.com/denoland/deno_graph or here since the diagnostic report is written there, but I figured I'd open it since this is where the documentation lives. 

I think both repositories would need a PR (which I'd be happy to submit for), since the [`ExportNotFound` message](https://github.com/denoland/deno_graph/blob/e7fabf923f5d2dc4da9ff824afbc7a5a1bdf017d/src/fast_check/mod.rs#L382) is not completely accurate, as it should probably say something like:
```
this error may be the result of a typo or a bug in Deno - if you think this is the case, please open an issue
```

### LINKED ISSUE #1202: Overview tab seems to prefer JSDoc over README
I stumbled over this issue when looking into https://github.com/paulmillr/qr. On [npm](https://www.npmjs.com/package/qr), the package's page just renders the README, which is what I (and presumably also the package's author) expected. On [jsr](https://jsr.io/@paulmillr/qr), however, the overview tab is completely useless. It seems to be generated from the [JSDoc at the top of the index.ts file](https://github.com/paulmillr/qr/blob/7afbebab1187af5176a541a070e5db8f2f008edd/src/index.ts#L18-L33).

Personally, I would consider this a bug on JSR's end as I would expect Overview to basically be the readme and only the Docs tab to be generated from JSDoc. I guess jsr could also add a config option to have the README be preferred over JSDoc if keeping this as the default is important to you. I wasn't able to find such an option after some brief searching, so I decided to open an issue here instead of in the QR repo.

crowlKats:
## Summary

- **Slow types docs** (#1063, #290, #370, #509): Add missing documentation sections for `ExportNotFound`, `unsupported-nested-javascript`, `unsupported-javascript-entrypoint` (.d.ts clarification), and `unsupported-expando-property` diagnostics, with corresponding `/go/` redirects
- **Exports object syntax** (#767): Add examples showing the object syntax for the `exports` field in `publishing-packages.md`
- **Overview tab clarification** (#1202): Clarify in both `writing-docs.md` and `packages.md` that `@module` JSDoc comments on the default entrypoint take precedence over README on the Overview tab
- **Other CI providers** (#435): Add documentation for publishing from GitLab CI and other CI providers using personal access tokens
- **Pre-release versions** (#682): Document how SemVer pre-release versions work on JSR (excluded from latest, excluded from semver ranges)
- **Unpublished packages** (#639): fix badge to show "no versions" with gray color instead of null

Closes #1063
Closes #290
Closes #370
Closes #767
Closes #1202
Closes #435
Closes #639
Closes #682
Closes #509

## Test plan

- [x] Verify `/go/slow-type-unsupported-nested-javascript` redirects to `#nested-javascript-modules`
- [x] Verify `/go/slow-type-export-not-found` redirects to `#export-not-found`
- [x] Verify `/go/slow-type-unsupported-expando-property` redirects to `#expando-properties`
- [x] Verify badge shows "no versions" for packages with no published versions


frontend/docs/packages.md
```diff
@@ -179,11 +215,17 @@ locally. This will generate HTML files with very similar looking documentation
...
+the top of the file that includes the `@module` tag. If the default entrypoint
+has a module doc, **the README will not be shown on the Overview tab** — the
+module doc takes precedence. If the package does not have a default entrypoint,
+or the default entrypoint does not have a module doc, then the "Overview" tab
+will show the README of the package instead. If no README is present, then the
 "Overview" tab will only show the package outline in the sidebar.
 
+If you want the README to be displayed on the Overview tab, ensure that your
+default entrypoint does not have a `@module` JSDoc comment.
+
```
KnorpelSenf: I think this can be changed in the package settings, too. Should that be mentioned here?

crowlKats: right

frontend/docs/writing-docs.md
```diff
@@ -171,8 +171,13 @@ You can also include examples in module documentation:
...
 
-If a default entrypoint has a module documentation, it takes precedence over the
-README file in the "Overview" tab of the package page.
+If the default entrypoint (the `.` export) has a module doc comment (a JSDoc
+comment with the `@module` tag), that documentation takes precedence over the
+README file in the "Overview" tab of the package page. This means that if your
+default entrypoint has a `@module` doc comment, the README will **not** be shown
+on the Overview tab — only the module doc will be displayed. If you want the
+README to appear on the Overview tab instead, remove the `@module` tag from the
+default entrypoint's doc comment.
```
KnorpelSenf: same here then

KnorpelSenf (approved):
LGTM otherwise, great changes

bartlomieju (approved):

PR #1367, opened 2026-04-03
fix: improve package score accuracy for docs and symbols

### LINKED ISSUE #1104: The score tab does not indicate which modules lack docs
In the score tab you can get dinged for not having module docs on all entrypoints:
Image

However, I believe that I have put module docs on all entrypoints and modules yet this is still showing an error.

It would be really nice if there was a visual indication of which modules were causing this check to fail still.

Also it would be nice to indicate the time stamp of the last time it checked and if its not a real-time check per publication, provide a button that could be pressed to enqueue a recheck.

### LINKED ISSUE #1155: Score for "Has docs for most symbols" seems to include JSON exports, which can't be documented
The score for "Has docs for most symbols" seems to include JSON exports, which can't be documented.

For example, https://jsr.io/@li/lorem-babel/score lists 15% coverage. All publicly accessible TS exports have documentation, but there are a lot of JSON exports for the various locales, so I assume that's why the number is low.

### LINKED ISSUE #1356: Score uses documentation for overload implementation
"Has docs for most symbols" counts documentation for implementation exports of function overloads. The JSDoc for these neither show up in JSR nor they are meaningful for IDE assist.

For example, the following code would produce 66% coverage.

```ts
/** String variant. */
export function func(value: string): void;

/** Number variant. */
export function func(value: number): void;

export function func(value: string | number): void {
  if (typeof value === "string") console.log("string");
  else console.log("number");
}
```

How JSR presents this function: https://jsr.io/@tugrulates/test/doc/~/func

crowlKats:
## Summary

- **Skip overload implementation signatures** in "has docs for most symbols" score calculation — the implementation signature of function overloads (`has_body: true`) is not user-facing and its docs don't appear in generated documentation or IDE assist. Fixes #1356
- **Exclude JSON module exports** from both symbol documentation percentage and entrypoint module doc checks, since JSON files can't have JSDoc documentation. Fixes #1155
- **Show which entrypoints are missing module docs** in the score tab — when "has module docs in all entrypoints" is not fully satisfied, the specific entrypoint names (from the exports map) are listed so users know exactly what to fix. Fixes #1104

## Test plan

- [x] Added unit tests for overload implementation skipping (`percentage_docs_skips_overload_implementation`)
- [x] Added unit tests for JSON module exclusion (`percentage_docs_skips_json_modules`, `entrypoints_missing_docs_skips_json`)
- [x] Added unit tests for entrypoint name reporting (`entrypoints_missing_docs_returns_export_names`)
- [x] Added unit test for single function (no overloads) counting normally (`percentage_docs_counts_single_function_normally`)
- [x] All existing tests continue to pass
- [ ] Verify frontend renders entrypoint names correctly on score tab


api/src/analysis.rs
```diff
@@ -339,32 +345,50 @@ fn generate_score(
...
   has_readme: bool,
-) -> bool {
-  'modules: for (specifier, document) in documents_by_url {
-    // Skip WASM modules as their docs are auto-generated from binary
-    if specifier.path().ends_with(".wasm") {
+  exports: &ExportsMap,
+) -> Vec<String> {
+  // Build a reverse map from file URL path to export key
+  let url_to_export: HashMap<String, String> = exports
+    .iter()
```
crowlbot, line 358: Subtle bug: the reverse map's keys are built by direct string interpolation (`format!("file://{}", path)`), while `documents_by_url`'s keys come from `Url::parse(&format!("file://{}", path)).unwrap()` (see line 125 in `analyze_package_inner`). `Url::parse` percent-encodes characters like spaces — so an export `"./hello world": "./hello world.ts"` produces:

- `url_to_export` key: `"file:///hello world.ts"`
- `documents_by_url` key: `"file:///hello%20world.ts"`

The `.get(specifier.as_str())` misses and falls back to `specifier.path()`, so the UI shows `/hello%20world.ts` instead of `./hello world`. Easy fix:

```rust
let url = url::Url::parse(&format!("file://{}", path)).unwrap();
(url.into(), key.clone())
```

or iterate exports inside the for-loop and avoid the reverse map entirely (which sidesteps the question — exports is small, so O(n*m) is fine).

api/src/analysis.rs
```diff
@@ -339,32 +345,50 @@ fn generate_score(
...
+    .map(|(key, path)| {
+      let path = path.strip_prefix('.').unwrap();
+      (format!("file://{}", path), key.clone())
+    })
```
crowlbot, line 363: Same `strip_prefix('.').unwrap()` pattern as `analyze_package_inner` line 110. The publish path enforces that exports start with `.` upstream, but if a corrupted DB row somehow has an export like `"./.":"foo.ts"` (no leading-dot path), this would panic during analysis. Worth using `.unwrap_or(path.as_str())` defensively, since the cost of a panic in this code path is much higher than the cost of a slightly-wrong reverse-map key.

api/src/analysis.rs
```diff
@@ -377,15 +401,41 @@ fn percentage_of_symbols_with_docs(documents_by_url: &ParseOutput) -> f32 {
       continue;
     }
 
+    // Skip JSON modules as they can't have JSDoc documentation
```
crowlbot, line 407: JSON modules are now skipped here, which matches what `entrypoints_missing_module_doc` does at line 372. The pre-existing `.wasm` skip is in `entrypoints_missing_module_doc` only — should it apply here too? A WASM entrypoint's symbols (synthesized from binary exports) currently count toward the `percentage_documented_symbols` denominator as undocumented, which would pull the score down for any package with a WASM module. Same reasoning as JSON: WASM can't carry JSDoc.

api/src/analysis.rs
```diff
@@ -377,15 +401,41 @@ fn percentage_of_symbols_with_docs(documents_by_url: &ParseOutput) -> f32 {
...
+      // For function overloads, skip the implementation signature (has_body: true)
+      // as it is not user-facing and its docs don't appear in generated documentation.
+      let has_overloads = non_private_decls.len() > 1
+        && non_private_decls.iter().all(|decl| {
+          matches!(&decl.def, deno_doc::node::DeclarationDef::Function(_))
+        });
 
-          if !decl.js_doc.is_empty() {
-            documented_symbols += 1;
+      for decl in &non_private_decls {
```
crowlbot, line 431: The `has_overloads` heuristic requires *every* non-private declaration to be a `Function` variant. That correctly handles the common case (`function foo(x: string): T; function foo(x: number): T; function foo(x: any): T { /* body */ }`), but won't trigger when there's declaration-merging like `function foo() {}; namespace foo { ... }` — those will count both the function impl and the namespace as separate symbols, with the function impl undocumented. Probably the right call (they aren't really overloads), but worth a one-line comment explaining the choice next to the `&& non_private_decls.iter().all(...)` predicate.

frontend/routes/package/score.tsx
```diff
@@ -127,6 +127,17 @@ function ScoreInfo(props: {
             module doc
           </a>{" "}
           summarizing what is defined in that module.
+          {score.entrypointsWithoutDocs.length > 0 && (
+            <span>
+              {" "}Entrypoints missing module docs:{" "}
```
crowlbot: This is the `jsx-curly-braces` error blocking the frontend CI. `deno lint --fix` rewrites to:

```tsx
  Entrypoints missing module docs:{" "}
```

(or you can swap to literal text with the comma separator pattern already used below). Also a UX nit: for a package with many missing entrypoints, the inline comma list can get long — worth thinking about whether to truncate after, say, 5 with a `(+N more)` or move to a `<ul>` for accessibility. Not blocking, just a consideration as packages with 20+ entrypoints start using this.

api/src/db/models.rs
```diff
@@ -482,6 +482,8 @@ pub struct PackageVersionMeta {
   pub has_readme: bool,
   pub has_readme_examples: bool,
   pub all_entrypoints_docs: bool,
+  #[serde(default)]
```
crowlbot, line 485: `#[serde(default)]` is the right move for backward compatibility — existing `PackageVersionMeta` JSON in `package_versions.meta` (or wherever this is persisted) will deserialize cleanly with an empty `Vec`. Newly-published versions will write the populated field. No migration needed. ✅

crowlbot (commented):
Substantively this is in good shape — the three fixes are well-scoped, the tests cover the new behavior, and the `#[serde(default)]` keeps the DB-stored `PackageVersionMeta` JSON backward-compatible. Two CI failures and one subtle URL-canonicalization bug to fix before merge:

1. **`check` (cargo fmt) is red**: the five new test functions in `api/src/analysis.rs` have unformatted `vec![...]` arguments — `cargo fmt -p jsr-api` rewrites them and the job goes green.
2. **`frontend` (deno lint) is red**: `score.tsx:132` has `{" "}Entrypoints missing module docs:{" "}` which trips the `jsx-curly-braces` rule. `deno lint --fix` fixes it.
3. **The reverse-URL map in `entrypoints_missing_module_doc` doesn't go through `Url::parse`, but `documents_by_url`'s keys do**, so any export path with special characters (spaces, unicode, percent-escape-worthy bytes) won't match and you'll get the raw file path in the UI instead of the export key. See inline.

The substantive question is: do you want to also exclude `.wasm` modules from `percentage_of_symbols_with_docs`? You already excluded them in `entrypoints_missing_module_doc` (it pre-existed there) and JSON from both. WASM modules also can't have JSDoc, so the asymmetry feels accidental.

crowlbot (approved):
All the concerns from my review are addressed across c054496 and 50a077a. The code is thorough and well-tested:

1. **`cargo fmt` / `deno lint`**: CI fully green. ✅
2. **URL canonicalization bug**: `url_to_export` now builds keys via `Url::parse`, matching `documents_by_url`'s percent-encoded keys. No more mismatch on paths with spaces/special characters. ✅
3. **`strip_prefix('.').unwrap()` panic risk**: Changed to `.unwrap_or(path.as_str())` — defensive, no-cost fallback. ✅
4. **JSON modules skipped in `percentage_of_symbols_with_docs`**: Added alongside the existing WASM skip in `entrypoints_missing_module_doc`. ✅
5. **Function overloads**: The `has_overloads` heuristic correctly skips `has_body` implementation signatures. The comment explains why declaration-merging cases aren't treated as overloads. ✅
6. **`entrypoints_missing_module_doc` → returns export names**: Replaces the old boolean `all_entrypoints_have_module_doc`. Score page now lists which entrypoints are missing docs — much more actionable for package authors. ✅
7. **`#[serde(default)]`**: Backward-compatible deserialization for `entrypoints_without_docs`. ✅
8. **Tests**: Five new tests covering overload skipping, single-function counting, JSON module skipping, missing-entrypoint naming, and JSON entrypoint skipping. Good coverage. ✅

One minor open question from my original review: WASM modules are skipped in `entrypoints_missing_module_doc` but not in `percentage_of_symbols_with_docs`. Extremely rare in practice and not blocking — just noting for completeness.

Ship it.

Issue #1250, opened 2026-01-08
Doc generation doesn't include new signatures


alexanderharding:
I have a package with this symbol: https://jsr.io/@xan/observable-core/doc/~/ObservableConstructor

Though it's new signature has docs;
<img width="705" height="522" alt="Image" src="https://github.com/user-attachments/assets/616c4180-f325-4755-8ccb-229f9f5c7370" />

This is not reflected in the generated documentation;
<img width="609" height="375" alt="Image" src="https://github.com/user-attachments/assets/ee820842-c088-4b83-b8dc-29b1f6cc52ef" />

crowlKats:
Verified fixed: interface construct signatures now render with their JSDoc (description, `@param` docs and `@example`s) since denoland/deno_doc#836, included in deno_doc 0.204.0 which jsr's main now pins (also fixed in `deno doc` itself per denoland/deno#32099). I reproduced on a local jsr instance with an `ObservableConstructor`-style interface — the `new (...)` signature shows with its docs.

This will be visible on jsr.io after the next deploy; if your package still misses it then, please reopen. Closing.

Issue #1013, opened 2025-03-11
Symbols missing documentation from "implements"


panva:
The implemented interface (`implements [types.ProduceJWT]`) is not linked, there's no click through to the interface that a class implements.

<details>
<summary>Details</summary>

![Image](https://github.com/user-attachments/assets/409de5a9-942c-4e01-b658-1b9fe7df21fa)

</details>

Furthermore, none of the methods from said implemented interface have their documentation used

See e.g. https://jsr.io/@panva/jose/doc/jwt/sign/~/SignJWT#method_setaudience_0 which lacks documentation from the interface[^1][^2] despite typescript obviously picking up on those and using them in intellisense.

[^1]: https://jsr.io/@panva/jose/doc/~/ProduceJWT.setAudience
[^2]: https://jsr.io/@panva/jose/doc/~/ProduceJWT#methods_setaudience_2

Issue #957, opened 2025-02-12
Re-exported interfaces are missing in the auto-generated docs


nicolo-ribaudo:
In this package: https://jsr.io/@babel-test-6ae45912/types@0.0.7-8.0.0-alpha.15

- You can see that [`./lib/ast-types/generated/index.ts`](https://jsr.io/@babel-test-6ae45912/types/0.0.7-8.0.0-alpha.15/lib/ast-types/generated/index.ts) exports a lot of interfaces
- They are re-exported at [line 75 of `./lib/index.ts`](https://jsr.io/@babel-test-6ae45912/types/0.0.7-8.0.0-alpha.15/lib/index.ts#L75) (the entrypoint)
- None of those interfaces is listed at https://jsr.io/@babel-test-6ae45912/types@0.0.7-8.0.0-alpha.15/doc

Issue #919, opened 2025-02-02
Types imported through `import *` don't get linked


nicolo-ribaudo:
`Expression` at https://jsr.io/@babel-test-6ae45912/template@0.0.1-8.0.0-alpha.15/doc/~/expression

PR #1371, opened 2026-04-04
fix: sync frontend @deno/doc with api deno_doc version

### LINKED ISSUE #917: 404 autogenerated link
https://jsr.io/@babel-test-6ae45912/traverse@0.0.1-8.0.0-alpha.15/doc/~/Scope.Binding

Either that `_Binding` page should exist, or it shouldn't be a link.

### LINKED ISSUE #1092: "Import symbol" generates unrunnable imports when package name is a reserved word
For example: `@mycoolscope/enum` generates the import `import * as enum from "@mycoolscope/enum";`.

One possible workaround is to capitalise package names when the package name is a keyword. E.g. `import * as Enum from "@mycoolscope/enum";`. Either that, or people can just recognise when a package name is a reserved word and come up with their own workarounds.

crowlKats:
## Summary

The doc generation fixes originally delivered here via a local `deno_doc` patch have since shipped upstream: denoland/deno_doc#793 merged and was released in `deno_doc 0.199.0`, and main is already on `=0.201.0` for the Rust API. The `[patch.crates-io]` block, version pin, and pathless lock entry are gone after rebasing onto main.

What remained out of sync was the frontend: `api/Cargo.toml` says `# sync with frontend/deno.json`, but the frontend still pinned `@deno/doc@^0.194.0`. This PR bumps it to `^0.201.0` to match the API, so the frontend types line up with the doc JSON the API now produces:

- **Reserved word handling in import usage** (#1092)
- **Interface construct signatures** (#1250)
- **Doc inheritance from interfaces** (#1013)
- **Type parameter and import linking** (#922, #919, #865)
- **Partition deduplication fix** (#957)

Fixes #1092, #1250, #1013, #919, #957
Partially addresses #922, #865

## Test plan

- [x] `deno check` in `frontend/` passes with `@deno/doc@0.201.0` (no type changes needed)
- [x] `Cargo.toml`/`Cargo.lock` are untouched relative to main

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1187,9 +1187,7 @@ dependencies = [
 
 [[package]]
 name = "deno_doc"
-version = "0.195.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "dca3e6557d8ae3fd25140a2b4d18f4e0c3ac9ab604b4f121e96c4681e22b423c"
+version = "0.197.0"
 dependencies = [
  "anyhow",
  "cfg-if",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,3 +1,6 @@
 [workspace]
 members = ["api", "api/macros"]
 resolver = "2"
+
+[patch.crates-io]
+deno_doc = { path = "../deno_doc" }
--- a/api/Cargo.toml
+++ b/api/Cargo.toml
@@ -84,7 +84,7 @@ async-compression = { version = "0.4", features = ["futures-io", "gzip"] }
 deno_graph = "=0.107.0"
 deno_ast = { version = "0.53.0", features = ["view"] }
 # sync with frontend/deno.json
-deno_doc = { version = "=0.195.0", features = ["comrak"] }
+deno_doc = { version = "=0.197.0", features = ["comrak"] }
 deno_error = "0.7.0"
 comrak = { version = "0.29.0", default-features = false }
 ammonia = "4.0.0"
--- a/frontend/deno.json
+++ b/frontend/deno.json
@@ -14,7 +14,7 @@
     }
   },
   "imports": {
-    "@deno/doc": "jsr:@deno/doc@^0.194.0",
+    "@deno/doc": "jsr:@deno/doc@^0.201.0",
     "@fresh/plugin-tailwind": "jsr:@fresh/plugin-tailwind@^1.0.0",
     "@fresh/plugin-vite": "jsr:@fresh/plugin-vite@^1.1.2",
     "@tailwindcss/vite": "npm:@tailwindcss/vite@^4.3.0",
--- a/frontend/deno.lock
+++ b/frontend/deno.lock
@@ -2,7 +2,7 @@
   "version": "5",
   "specifiers": {
     "jsr:@deno/cache-dir@0.25": "0.25.0",
-    "jsr:@deno/doc@0.194": "0.194.1",
+    "jsr:@deno/doc@0.201": "0.201.0",
     "jsr:@deno/esbuild-plugin@^1.2.0": "1.2.1",
     "jsr:@deno/gfm@0.11": "0.11.0",
     "jsr:@deno/graph@0.100": "0.100.1",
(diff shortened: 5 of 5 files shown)


Cargo.toml
```diff
@@ -1,3 +1,6 @@
 [workspace]
 members = ["api", "api/macros"]
 resolver = "2"
+
+[patch.crates-io]
+deno_doc = { path = "../deno_doc" }
```
crowlbot, line 6: Drop the `[patch.crates-io]` block before merging — `../deno_doc` doesn't exist in CI, which is exactly what's failing the `check` and `test` jobs. The upstream changes are released as `deno_doc 0.199.0` (2026-05-02), so the patch isn't needed anymore.

api/Cargo.toml
```diff
@@ -84,7 +84,7 @@ async-compression = { version = "0.4", features = ["futures-io", "gzip"] }
 deno_graph = "=0.107.0"
 deno_ast = { version = "0.53.0", features = ["view"] }
 # sync with frontend/deno.json
-deno_doc = { version = "=0.195.0", features = ["comrak"] }
+deno_doc = { version = "=0.197.0", features = ["comrak"] }
```
crowlbot, line 87: This pins `=0.197.0`, which was published 2026-03-21 — before the upstream docs-fixes PR (denoland/deno_doc#793) merged on 2026-04-17. So the path patch was actually doing the heavy lifting; the version pin alone wouldn't contain any of the fixes this PR is supposed to deliver. After removing the patch block, bump this to `=0.199.0` (which both contains #793 and is a superset of `=0.198.0` that main is on, resolving the conflict).

Cargo.lock
```diff
@@ -1187,9 +1187,7 @@ dependencies = [
 
 [[package]]
 name = "deno_doc"
-version = "0.195.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "dca3e6557d8ae3fd25140a2b4d18f4e0c3ac9ab604b4f121e96c4681e22b423c"
+version = "0.197.0"
```
crowlbot, line 1190: This `deno_doc` lock entry has no `source` or `checksum` line — that's the tell that `cargo` resolved it through the local `[patch.crates-io]` rather than crates.io. Once `Cargo.toml` is cleaned up, `cargo update -p deno_doc` will regenerate this with the proper `source = "registry+..."` and `checksum = "..."` (and bump the version to 0.199.0). Worth committing the regenerated lock as the last step before pushing.

crowlbot (commented):
The substantive deno_doc changes here (reserved-word import handling, interface constructors, doc inheritance, etc.) look great — upstream PR #793 was merged on 2026-04-17 and shipped in `deno_doc 0.199.0` on 2026-05-02. But the form of this PR can't merge as-is:

1. **`[patch.crates-io]` with a relative path** (`../deno_doc`) breaks any build that doesn't have a sibling `deno_doc` checkout. That's why `check` and `test` are red — CI doesn't have one. This was the right thing while #793 was unreleased; now that 0.199.0 is on crates.io, the patch block should be deleted.
2. **The version pin (`=0.197.0`) is wrong** for what this PR actually wants. 0.197.0 was published 2026-03-21, before #793 merged — it doesn't contain any of these fixes. The path patch was masking that. The intended target is `=0.199.0`.
3. **Stale against main**: main moved to `deno_doc = "=0.198.0"` in #1382 (stack-overflow fix). That's the source of the `CONFLICTING` state, and bumping straight to `=0.199.0` resolves it cleanly (0.199.0 is a superset of 0.198.0 + #793).

Concretely, the fix on this branch is roughly:

```diff
--- a/Cargo.toml
+++ b/Cargo.toml
@@
 [workspace]
 members = ["api", "api/macros"]
 resolver = "2"
-
-[patch.crates-io]
-deno_doc = { path = "../deno_doc" }
```
```diff
--- a/api/Cargo.toml
+++ b/api/Cargo.toml
@@
-deno_doc = { version = "=0.195.0", features = ["comrak"] }
+deno_doc = { version = "=0.199.0", features = ["comrak"] }
```
…then `cargo update -p deno_doc` to regenerate the lock entry with a real source/checksum.

One sanity-check before pushing: re-run docs generation against a package that exercises each of the five fix areas (reserved-word imports like `enum`, interface `new()`, doc inheritance, type-parameter cross-module linking, partition dedup) to confirm 0.199.0 carries the same behavior the local-path build was showing. There's been one release on crates.io since #793 merged, so the diff should be minimal but worth eyeballing.

Also worth syncing `frontend/deno.json`'s deno_doc dep if it's tracked separately (the existing comment `# sync with frontend/deno.json` in `api/Cargo.toml` implies it is).

James-Ansley:
This appears to be fixed by #1382 and later commits/PRs

Cargo.toml
crowlKats: Done — rebased onto main, which is already on `deno_doc =0.201.0` (a superset of 0.199.0), so the patch block, the pin, and the pathless lock entry are all gone. The PR now only carries the frontend `@deno/doc` sync to `^0.201.0`.

api/Cargo.toml
crowlKats: Resolved by the rebase — main already bumped to `=0.201.0`, which contains denoland/deno_doc#793. This PR now bumps the frontend's `@deno/doc` to `^0.201.0` to honor the `# sync with frontend/deno.json` comment.

Cargo.lock
crowlKats: Resolved — `Cargo.lock` is untouched relative to main after the rebase; the registry-sourced `deno_doc 0.201.0` entry with its checksum is what's in the lock now.

littledivy (approved):

PR #1372, opened 2026-04-04
perf: optimize slow database queries and add covering indexes

crowlKats:
## Summary

Addresses several slow PostgreSQL queries identified from GCP Cloud SQL logs. The optimizations fall into three categories: **query restructuring**, **covering indexes**, and a **write-time rollup table**.
Increases max connections to 150 since we are always at 99/100.

### Query restructuring

- **Package list queries (`list_packages_by_scope`, `list_packages`)**: Replaced 3 correlated subqueries per row (version count, latest version, latest version meta) with `LEFT JOIN LATERAL` subqueries. For a result set of N packages, this reduces the subquery overhead from 3N separate index lookups to 2N lateral scans that can use a single covering index. The original `PACKAGE_SELECT_JOINED` fragment is kept unchanged for single-row queries (e.g. `get_package`) where correlated subqueries are fine.

- **Metadata generation (`PackageMetadata::create`)**: The metadata endpoint only needs `(version, is_yanked, created_at)` but was using the full `list_package_versions` query which JOINs with the `users` table and fetches all version columns. Added a lightweight `list_package_versions_for_metadata` function with a dedicated `PackageVersionForMetadata` struct that queries only the three needed columns — no user JOIN, no extra columns.

- **Package download counts (`get_package_downloads_24h`)**: Previously aggregated per-version rows at read time with `GROUP BY scope, package, time_bucket, kind` + `SUM(count)`. For a package with 100 versions over 90 days, this scans ~18,000 rows. Now reads from a pre-aggregated `package_download_counts_24h` rollup table (~180 rows, no GROUP BY/SUM needed).

### New rollup table

- **`package_download_counts_24h`**: Stores package-level (not version-level) download counts. Populated at write time during `insert_download_entries` via an `INSERT ... ON CONFLICT DO UPDATE` upsert. The migration backfills from existing `version_download_counts_24h` data.

### Index optimizations

| Index | Purpose |
|-------|---------|
| `idx_package_versions_latest` | Covering index on `(scope, name, version DESC) INCLUDE (meta)` with partial filter `WHERE is_yanked = false AND version NOT LIKE '%-%'`. Enables index-only scans for the lateral join that fetches latest version + meta. |
| `idx_packages_scope_archived_name` | Composite index on `(scope, is_archived, name)` for `list_packages_by_scope` ordering pattern. |
| `idx_package_version_deps_kind_name` | Full key columns `(dependency_kind, dependency_name, package_scope, package_name)` instead of INCLUDE — enables PostgreSQL to produce sorted output for the `COUNT(DISTINCT)` dependent query, avoiding hash-based dedup. |
| `idx_download_counts_24h_package` | Covering index on `(scope, package, time_bucket) INCLUDE (version, kind, count)` for per-version download queries. |

### Files changed

- `api/src/db/sql_fragments.rs` — New lateral join SQL fragment constants
- `api/src/db/database.rs` — Rewired list queries to lateral joins, added metadata query, added rollup write, simplified download read
- `api/src/db/models.rs` — New `PackageVersionForMetadata` struct
- `api/src/metadata.rs` — Uses lightweight metadata query
- `api/migrations/20260404000000_optimize_package_version_indexes.sql` — New indexes + rollup table + backfill

## Test plan

- [ ] Run `deno task db:prepare` to regenerate `.sqlx/` cache files
- [ ] Verify `cargo check` passes with regenerated cache
- [ ] Run migration against a test database and verify indexes are created
- [ ] Verify backfill query populates `package_download_counts_24h` correctly
- [ ] Load test `list_packages_by_scope` and `get_package_downloads_24h` endpoints
- [ ] Verify `EXPLAIN ANALYZE` shows index-only scans on the new covering indexes


miracatbot (commented):
One correctness concern from review:

The new `package_download_counts_24h` rollup is backfilled from `version_download_counts_24h`, but ongoing writes rebuild it from `version_download_counts_4h`. Since the 4h buckets are later pruned, this means the package-level daily rollup does not have a stable long-term source of truth. If historical data is ever replayed or an older bucket is inserted after some 4h rows for that day have already aged out, the upsert here can overwrite a previously correct package/day total with a value recomputed from incomplete surviving 4h data.

I think this should either:
- derive the package-level 24h rollup from `version_download_counts_24h`, or
- explicitly guarantee that older days are never touched once 4h source data may be incomplete.

A regression test around historical/backfilled data plus a late older bucket would also help.

bartlomieju (approved):

PR #1376, opened 2026-04-06
fix: fix first h1 with preceding content

andylovescode:
Previously, h1 elements would be styled differently if they were the first of their kind.

This causes an issue in the case that you have text before your heading.

This commit changes the condition to only apply if the h1 has no preceding content, which looks better in my opinion.

# Preview

| Before | After |
| --- | --- |
| image | image |


KnorpelSenf:
Very nice idea, but I guess the LLM review actually has a reasonable point here?

andylovescode:
That also applies to the previous behavior, and as far as I know, this issue has not been observed with the previous behavior in production.

KnorpelSenf:
Hm good point

KnorpelSenf (approved):

andylovescode:
Apologies on this taking so long, I didn't realize I needed to sign the CLA, it's signed now.

PR #1380, opened 2026-04-14
feat: managed challenge for non-module endpoints

crowlKats:
## Summary
- Adds a Cloudflare managed challenge for bot protection on non-module endpoints (frontend + API) while leaving the module-file path untouched, so package clients (`deno add`, `npm install`, etc.) are unaffected.
- The Worker uses the exact `isModuleFilePath()` / `canAccessModuleFile()` routing logic to decide who needs a challenge, then redirects to `/__challenge`, where a WAF rule serves the managed challenge. After solving, the `cf_clearance` cookie lets subsequent requests pass.
- Requests without a `CF-Ray` header (local dev, direct Cloud Run) skip the challenge so we don't create an infinite redirect loop outside of Cloudflare.

## Tradeoffs
- One extra round-trip per `cf_clearance` lifetime (~30 min) for unchallenged visitors. The alternative (WAF-only matching) wasn't viable: WAF `matches` (regex) is Enterprise-only, and approximations don't cover every module path variant.

## Test plan
- [ ] Verify first visit to `jsr.io/` triggers the managed challenge and then loads normally
- [ ] Verify module fetches (`/@scope/pkg/x.y.z/mod.ts`, `meta.json`, `x.y.z_meta.json`) bypass the challenge
- [ ] Verify `npm.jsr.io` and `api.jsr.io` are unaffected (they don't route through `handleRootRequest`)
- [ ] Verify bots (Googlebot) still reach the frontend without a challenge
- [ ] Verify local dev (`jsr.test`) no longer hits a redirect loop

### DIFF
--- a/lb/main.ts
+++ b/lb/main.ts
@@ -158,7 +158,15 @@ export async function handleRootRequest(
   const url = new URL(request.url);
   const path = url.pathname;
 
+  if (path === "/__challenge") {
+    const returnPath = url.searchParams.get("return") || "/";
+    const safeReturn = returnPath.startsWith("/") ? returnPath : "/";
+    return Response.redirect(new URL(safeReturn, url.origin).toString(), 302);
+  }
+
   if (isAPIRoute(path)) {
+    const challenge = requireChallenge(request);
+    if (challenge) return challenge;
     return await handleAPIRequest(request, env, false);
   } else if (isBot(request)) {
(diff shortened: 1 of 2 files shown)


crowlKats:
Closing — a managed challenge interstitial is the wrong UX for a package registry. npm/crates.io don't do this, and neither should we.

PR #1381, opened 2026-04-14
feat: rate-limit the frontend proxy

crowlKats:
## Summary
- Adds a per-IP rate limit only on the Cloud Run frontend path (`handleFrontendRoute`). Modules (R2), the API, and npm compat are untouched.
- Uses Cloudflare's native rate-limit binding: 60 requests per minute, keyed on `CF-Connecting-IP`. Invisible to real users; bites scrapers generating cache-miss load on Cloud Run.
- Falls open when the binding or `CF-Connecting-IP` isn't present (local dev, direct Cloud Run), so nothing regresses outside of Cloudflare.

## Notes
- Cloudflare's rate-limit binding is per-colo and best-effort; it's the right lightweight tool for this, not a precise global quota. Fine for frontend protection.
- `limit`/`period` are tunable in `terraform/lb.tf` if 60/min turns out to be too lax or too strict.

## Test plan
- [ ] Verify normal browsing of jsr.io works (stays well under the limit)
- [ ] Verify `deno add` / `npm install` from a JSR package is unaffected (module + npm paths)
- [ ] Verify `api.jsr.io` and `jsr.io/api/*` are unaffected
- [ ] Hammer the frontend from one IP and confirm 429s after ~60 req/min


AugustinMauroy (approved):
seem good

crowlbot (commented):
Nice, tight change — falls open in local dev, the binding is exactly the right tool for this. A few things worth thinking about before merging, mostly around the boundary between this gate and the existing routing/caching layers.

The main one I'd want resolved: requests that `handleRootRequest` already classified as `isBot=true` (Googlebot, Slack/Twitter/Discord link previews) go through this same gate at 60/min/IP. For Slack/Twitter that's probably fine — they're a small share of traffic — but a Googlebot deep-crawl from one of its IPs *can* burst above 60/min on large packages, and a 429 response to Googlebot would tank indexing for the affected pages. Worth deciding whether the existing `isBot` arg (already plumbed into `handleFrontendRoute`) should bypass or get a higher ceiling.

The other notable thing is that this runs before `cachedFetch`'s `caches.default` lookup, so cache *hits* count against the limit too. That's actually defensible — a scraper that only fetches cached pages still costs you egress and the rate-limit binding is cheap — but the PR description ("scrapers generating cache-miss load on Cloud Run") undersells what's actually happening. Worth either adjusting the description or moving the check below the cache lookup if you really only want to protect Cloud Run.

Rest is nits inline.

lb/main.ts
```diff
@@ -221,6 +221,9 @@ async function handleFrontendRoute(
   env: WorkerEnv,
   isBot: boolean,
 ): Promise<Response> {
+  const limited = await isRateLimited(request, env);
```
crowlbot: `handleFrontendRoute` already receives `isBot: boolean` (set true for Googlebot/Slack/Twitter/WhatsApp/Discord in `handleRootRequest`). Currently those are rate-limited at the same 60/min/IP as everyone else.

For Googlebot in particular this could be a problem: deep crawls of large packages can spike above 60/min from a single colo'd IP, and serving 429s to Googlebot means dropping those pages from the index until the next crawl. Slack/Twitter previews are bursty too (think a popular tweet linking a JSR package), though the impact there is just a missing unfurl.

A cheap option: `if (isBot) return null;` at the top of `isRateLimited`, or move the `isRateLimited` call into the non-`isBot` branches of `handleRootRequest` only. The scraper threat model here is browser/CLI scrapers spoofing UA, which `isBot` doesn't catch anyway.

lb/main.ts
```diff
@@ -232,6 +235,30 @@ async function handleFrontendRoute(
   return response;
 }
 
+/**
+ * Per-IP rate limit for the frontend only. Module files (R2), the API, and
+ * npm compat are never rate-limited here — only the Cloud Run frontend, which
+ * is the backend that scrapers actually generate cache-miss load on.
+ */
+async function isRateLimited(
```
crowlbot: Naming nit: `isRateLimited` reads as a predicate but the return type is `Response | null` (truthy `Response` ⇒ blocked, `null` ⇒ allowed). Something like `rateLimitGuard`, `checkRateLimit`, or `tryRateLimit429` would match the shape better and avoid a double-take at the call site (`const limited = await isRateLimited(...)` — `limited` ends up being a `Response`, which is mildly confusing).

lb/main.ts
```diff
@@ -232,6 +235,30 @@ async function handleFrontendRoute(
...
+  request: Request,
+  env: WorkerEnv,
+): Promise<Response | null> {
+  const limiter = env.FRONTEND_RATELIMIT;
+  if (!limiter) return null;
+  const ip = request.headers.get("CF-Connecting-IP");
+  if (!ip) return null;
+  const { success } = await limiter.limit({ key: ip });
+  if (success) return null;
+  return new Response("Too Many Requests", {
```
crowlbot: Every other response path in this worker runs through `setSecurityHeaders(response, FRONTEND)` (line 229 for the proxy path, similar for API/NPM/MODULES). The 429 skips it, which means no HSTS, CSP, or X-Frame-Options on the rate-limit response. The body is `text/plain`, so CSP/XFO are largely moot — but losing HSTS on a 429 is a small consistency wart and trivial to fix:

```ts
const res = new Response("Too Many Requests", { status: 429, headers: { "Content-Type": "text/plain", "Retry-After": "60" } });
setSecurityHeaders(res, FRONTEND);
return res;
```

lb/main.ts
```diff
@@ -232,6 +235,30 @@ async function handleFrontendRoute(
...
+    status: 429,
+    headers: {
+      "Content-Type": "text/plain",
+      "Retry-After": "60",
```
crowlbot, line 257: `Retry-After: 60` is the worst-case wait — Cloudflare's `simple` rate-limit binding uses a fixed window of `period` seconds, so a client that hit the limit 5s into the current window only has to wait ~55s. Not wrong, just generous. Fine as-is unless you want to compute a tighter value, which the binding doesn't expose anyway.

terraform/lb.tf
```diff
@@ -53,6 +53,17 @@ resource "cloudflare_workers_script" "jsr_lb" {
...
       text = google_cloud_run_v2_service.registry_frontend["us-central1"].uri
+      }, {
+      # Per-IP rate limit applied only on the frontend proxy path (see
+      # handleFrontendRoute in lb/main.ts). Keeps scrapers from generating
+      # cache-miss load on Cloud Run without touching modules, API, or npm.
+      type         = "ratelimit"
+      name         = "FRONTEND_RATELIMIT"
+      namespace_id = "1001"
+      simple = {
+        limit  = 60
```
crowlbot: Two unrelated bits in this diff:

1. `namespace_id = "1001"` is a magic string. Since this is the first rate-limit binding in the worker, any value works, but a short comment explaining what 1001 is (Cloudflare's namespace IDs aren't documented as having reserved values, but this one is sometimes used as a default) would save the next person a Stack Overflow trip.
2. The diff also includes a couple of `terraform fmt` re-alignments in unrelated blocks (the `enabled = true` realign at line 14, the `}, {` reindent for the existing R2 binding at line 32). Not a blocker — just noting in case you want to split them into a separate "fmt" commit for cleaner history.

lb/types.ts
```diff
@@ -24,4 +24,8 @@ export interface WorkerEnv {
   DOWNLOADS?: AnalyticsEngineDataset;
   NPM_BUCKET: PartialBucket;
   MODULES_BUCKET: PartialBucket;
+
+  // Optional: omitted in local dev. Applied only to frontend HTML proxying so
+  // scrapers can't generate cache-miss load on the Cloud Run frontend.
+  FRONTEND_RATELIMIT?: RateLimit;
```
crowlbot, line 31: The comment here says "applied only to frontend HTML proxying" — which is true in spirit, but `handleFrontendRoute` also serves non-HTML responses from Cloud Run (e.g. the `/api/` JSON paths that get routed here when `isAPIRoute(path)` is false but the path still resolves to something the frontend renders, or static assets that the frontend serves directly). Tightening the wording to something like "Applied only to the Cloud Run frontend route — not modules (R2), the API server, or npm compat" matches the actual gate.

bartlomieju (approved):

terraform/lb.tf
```diff
@@ -53,6 +53,19 @@ resource "cloudflare_workers_script" "jsr_lb" {
...
+      # Per-IP rate limit applied only on the frontend proxy path (see
+      # handleFrontendRoute in lb/main.ts). Keeps scrapers from generating
+      # cache-miss load on Cloud Run without touching modules, API, or npm.
+      # namespace_id is a per-account identifier for this rate-limit binding;
+      # any unused value works (no Cloudflare-reserved meaning for "1001").
+      type         = "ratelimit"
+      name         = "FRONTEND_RATELIMIT"
+      namespace_id = "1001"
+      simple = {
+        limit  = 60
```
crowlKats:
```suggestion
        limit  = 30
```

PR #1406, opened 2026-05-29
feat: run the API server as a Cloudflare Container

crowlKats:
## Summary

Moves the Rust API server off Google Cloud Run and runs it as a **Cloudflare Container** fronted by the existing `lb` worker. A fresh implementation of #1378 on top of `main` (which has since gained the frontend-worker migration).

Background **tasks** continue to run on Cloud Run (a sleeping container can't run scheduled work), and the Cloud Run API **serving** service is kept as a fallback for now.

## Why this isn't a straight cherry-pick of #1378, and why it's wrangler-deployed

#1378 put a `containers` block (`image`/`max_instances`) on `cloudflare_workers_script`. The pinned `cloudflare` provider can't express that — it can attach a Durable Object class to a worker version but **cannot set a container image, roll one out, or run a DO migration** within its per-deploy version model.

So the LB worker is **deployed with `wrangler deploy`**, and to avoid hand-maintaining a `wrangler.jsonc` that duplicates tfvars, **Terraform renders `lb/wrangler.json`** from values it already owns (`local_file` in `lb.tf`). CI runs `terraform apply` (writes the config) → `wrangler deploy` (image + DO migration + container rollout in one) → waits for the rollout → e2e. Secrets stay in Terraform and are delivered with `wrangler secret bulk` (`tools/lb-secret-bulk.ts`).

### API (`api/`)
- `gcp.rs`: new `MetadataStrategy::ServiceAccountKey` — signs an RS256 JWT and exchanges it for a GCP access token, since the container runs on Cloudflare with no metadata server. `METADATA_STRATEGY` now selects this.
- `db/database.rs`: optional `DbTls` (client cert/key + root CA) → `sslmode=verify-ca` with an mTLS client cert (see DB security below).
- `main.rs`: install a process-level **ring** rustls `CryptoProvider` at startup (sqlx's TLS verifier needs one; ring runs in the container's firecracker VM, aws-lc-rs does not).
- Removed Cloud Trace (`--cloud_trace`, `gcp_project_id`, the `opentelemetry-gcloud-trace` dep, the `CloudTrace` target) and the orphaned bigquery testdata.
- **`Cargo.toml`: pinned `rustls = "=0.23.23"`** — see the gotcha below.

### LB worker (`lb/`)
- `containers.ts`: `ApiContainer` (`defaultPort 8001`, `sleepAfter 5m`, `--api --tasks=false` entrypoint, env passthrough to the container).
- `main.ts`: API requests routed to the container via `getRandom(env.API_CONTAINER, API_CONTAINER_INSTANCES)` (30, kept in sync with `max_instances`), reusing `proxyToBackend` so caching/CORS/headers are preserved.
- `types.ts` / `local.ts`: `API_CONTAINER` binding + the API env vars; local dev proxies straight to the local API server (no container locally).

### Terraform
- `lb.tf`: renders `lb/wrangler.json` (worker name, routes, container image/instance type, DO binding + `v1` migration, R2/services/ratelimit bindings, non-secret vars) and exposes the worker secrets as outputs.
- `db.tf` + `cloud_run_api.tf`: the mTLS DB model (below).
- `providers.tf`: add the `local` provider.
- `tools/wait-container-rollout.ts`: polls the CF containers API and blocks until the rollout completes before the deploy is considered done.

### CI (`.github/workflows/ci.yml`)
- `docker-images` pushes the API image to the CF registry (`wrangler containers push`).
- Each deploy job: `terraform apply` → set worker secrets → `wrangler deploy` → wait for container rollout → e2e.

## DB connectivity & security (mTLS)

The container egresses over the public internet and its IP **can't be pinned** (no Zero Trust dedicated-egress add-on), so it reaches Cloud SQL over the **public IP**. Rather than rely on a network ACL, the DB is set to `ssl_mode = TRUSTED_CLIENT_CERTIFICATE_REQUIRED`: the public IP is left open but a **required client certificate (mTLS)** is the access boundary. The client key is a secret (worker secrets / Cloud Run env), so only cert holders can connect — verified that a no-cert connection is rejected.

- Container → DB: public IP, `verify-ca` (the instance-specific CA is pinned; hostname is skipped because the cert SAN is an internal `*.sql.goog` name that never matches the IP) + client cert.
- Cloud Run (serving + tasks) → DB: private VPC IP, same `verify-ca` + client cert (delivered via `api_envs`).
- Applies to **both** staging and prod.

### ⚠️ The `rustls = "=0.23.23"` pin

sqlx 0.8.6's `NoHostnameTlsVerifier` (used for `verify-ca` to skip the hostname) only swallows rustls' old `CertificateError::NotValidForName`. rustls **≥ 0.23.24** returns `NotValidForNameContext` instead, which sqlx 0.8.6 does not catch — so `verify-ca` aborts the handshake with a `bad_certificate` alert (seen in the Cloud SQL logs as `SSLV3_ALERT_BAD_CERTIFICATE`) and the container crashed on startup at `Database::connect`. Verified against the live DB: 0.23.23 connects, 0.23.24+ fails. The fix is already on sqlx `main`; **drop the pin once a sqlx > 0.8.6 ships it.**

## Deploy / cutover notes

- **One-time state surgery per environment.** The worker + routes move from Terraform ownership (`cloudflare_workers_script.jsr_lb` + `cloudflare_workers_route.*`) to wrangler. Before the first apply, remove them from state so Terraform doesn't delete the live worker — wrangler then adopts it by name:
  ```
  terraform state rm cloudflare_workers_script.jsr_lb \
    cloudflare_workers_route.jsr_root cloudflare_workers_route.jsr_api cloudflare_workers_route.jsr_npm
  ```
  Done on staging; **must be run against prod state before merging.**
- Container sizing: 1 vCPU / 3 GiB (Cloudflare requires ≥3 GiB per vCPU; only ~512 MiB is used), up to 30 instances.

## Status

- ✅ Staging: green CI, container serving, mTLS enforced and verified.
- Prod: config is ready (same model); needs the state-rm step above on land.

## Follow-ups

- Drop the `rustls` pin when sqlx > 0.8.6 is released.
- Tear down the Cloud Run API **serving** service once the container is proven in prod.
- Revisit DB pool sizing (30 instances × pool 4 = 120 vs `max_connections` 150) before prod traffic.
- Consider Cloudflare dedicated-egress IPs to replace the open public IP with a pinned allowlist.


crowlbot (commented):
Solid PR — the rebase on top of the post-frontend-migration `main` is clean, the version → wait-for-rollout → promote sequencing is the right model, and the surgical removal of Cloud Trace / gcemeta / opentelemetry-gcloud-trace (and their transitive `tonic 0.9` / `prost 0.13` duplicates) materially simplifies `Cargo.lock`. The split — serving on Cloudflare, tasks staying on Cloud Run — is the right call given containers sleep.

A few things worth confirming before merge, in roughly decreasing order of risk:

### 1. DO migrations on every new `cloudflare_worker_version` (terraform/lb.tf:78–81)

```hcl
migrations = {
  new_tag            = "v1"
  new_sqlite_classes = ["ApiContainer"]
}
```

`cloudflare_worker_version` `RequiresReplace` on most fields, so every code change re-creates this resource with the **same** `new_tag = "v1"` and the **same** `new_sqlite_classes = ["ApiContainer"]`. Per the [Durable Objects migrations docs](https://developers.cloudflare.com/durable-objects/reference/durable-objects-migrations/): *"Migration tags are treated like unique names and are used to determine which migrations have already been applied."* Behavior on second apply could be (a) tolerated no-op, (b) hard error from the API, or (c) silent drift. The frontend `cloudflare_worker_version` carries no `migrations` block so this path is genuinely new in this repo.

The safest patterns I'm aware of:

- Set `migrations` only when actually changing DO classes (subsequent versions omit the block or pass `null`), or
- Make the block conditional on a `migrations_version` variable that you bump manually when adding/renaming classes.

Worth a single staging deploy of two consecutive code-change apply cycles to confirm what actually happens.

### 2. Capacity ceiling from `getRandom(env.API_CONTAINER, 3)` + `sleepAfter = "5m"`

`lb/main.ts:81` hardcodes 3 containers and `lb/containers.ts:38` puts them to sleep after 5 min. With `--database_pool_size=4` per container, that's a **12-connection** steady-state ceiling on API traffic. Cloud Run today autoscales the serving service; this swaps autoscaling for a fixed fleet of 3.

Two concrete concerns:

- **Worst-case load**: if 4th-and-beyond concurrent request waves arrive before any container queues drain, latency tails grow. Have you sized 3 × pool=4 against current Cloud Run concurrency at peak?
- **Random LB ≠ least-loaded LB**: `getRandom` can hammer one container while another is idle. With only 3, the variance is meaningful. `getByName` with a sticky-ish key (e.g. hashed client IP, or path-prefix) at least spreads load deterministically.

Cloud Run remains as fallback during cutover so the rollout-level risk is bounded, but the steady-state ceiling deserves an explicit decision.

### 3. CI image-existence check splits GCP vs Cloudflare (.github/workflows/ci.yml:184–198)

```yaml
- name: Push api image to Cloudflare container registry
  if: github.event_name != 'push' || steps.check_existing.outputs.exists != 'true'
  run: wrangler containers push jsr-api:${{ github.sha }}
```

`check_existing` queries the **GCP** Artifact Registry. If a prior workflow run pushed to GCP but failed before reaching the new Cloudflare-push step (or pre-dated this PR), every subsequent run for that SHA skips the CF push too — but the deploy job still proceeds, and the container rollout will be stuck pulling a tag that doesn't exist in CF. The "deploy mysteriously hangs after a 'no-op' image step" failure mode is hard to spot.

Two safe fixes: (a) gate the CF push on its own existence check against the CF registry, or (b) always run `wrangler containers push` — image pushes by digest are no-ops if the layers are already present, so the cost is one HEAD round-trip.

### 4. `wrangler containers push` ordering on the first deploy

The image is pushed to Cloudflare in `docker-images` before any `terraform apply` has ever run. The `containers = [{ class_name }]` attachment in `cloudflare_worker_version` is what creates the container application on the first deploy. Two questions worth verifying with a clean staging:

- Does `wrangler containers push <tag>` push to the **account-level** registry (success without an application), or does it require the application to exist? If the latter, the first-ever deploy can't bootstrap.
- Once the application exists, how does it pick up the SHA tag? `tools/wait-container-rollout.ts` polls *something*, but nothing in this PR appears to tell the application "use image `jsr-api:<SHA>`". Is there an implicit `latest` convention, or does the rollout trigger come from the `worker_version` resource creation itself?

Both of these are flagged in the PR description, so this is more a "let's keep an eye on the first staging run" than a finding.

### 5. `METADATA_STRATEGY` as a polymorphic env var (api/src/gcp.rs:34–43)

```rust
_ => {
  if s.starts_with('{') {
    Ok(Self::ServiceAccountKey(s.to_owned()))
  } else {
    Err(anyhow::anyhow!("Invalid metadata strategy '{}'", s))
  }
}
```

Functional, but the env var name no longer matches what it carries on Cloudflare. Two consequences:

- The terraform-level binding name (`METADATA_STRATEGY`) reads like "instance_metadata / testing", but on Cloudflare it's a 2KB PEM-wrapped JSON blob. New reader-of-the-code overhead.
- `s.starts_with('{')` is a fragile sniff — if anyone ever sets `METADATA_STRATEGY=" {...}"` (leading whitespace from a CI editor or a terraform `<<EOT`), parsing dies in the deeper `serde_json::from_str` call with a panic via `expect("Failed to parse GCP service account key JSON")`.

Suggest splitting: keep `METADATA_STRATEGY ∈ {instance_metadata, service_account_key, testing}`, and put the key JSON in its own `GCP_SERVICE_ACCOUNT_KEY` env var read only when strategy == `service_account_key`. The TF wiring barely changes.

### 6. JWT signing on the Tokio executor (api/src/gcp.rs:177–181)

```rust
let encoding_key = jsonwebtoken::EncodingKey::from_rsa_pem(
  sa_key.private_key.as_bytes(),
)?;
let jwt = jsonwebtoken::encode(&header, &claims, &encoding_key)?;
```

Both calls are sync + CPU-bound (RSA sign). The cache normally hits, so this fires ~once/hour and is fine. But `from_rsa_pem` re-parses the PEM on every refresh — cheap, but unnecessary. Parsing the `EncodingKey` once at `Client::new` time and storing it next to `service_account_key` removes the per-refresh PEM parse and the panic from `expect` on the parsed JSON (you'd want to surface that error earlier anyway, ideally as a startup-time failure).

### 7. Concurrent token-fetch race (api/src/gcp.rs:157–171)

```rust
{
  let guard = self.access_token.lock().unwrap();
  if let Some((token, expires_at)) = guard.clone() && expires_at > ... { return Ok(token); }
}
// JWT + network call, lock released
let mut guard = self.access_token.lock().unwrap();
*guard = Some(...);
```

N concurrent callers with a stale cache all do the JWT signing + network round-trip, then race to write back. Pre-existing pattern in the `InstanceMetadata` branch above, so no regression — flagging it because this PR doubles the code paths that have this shape. A `tokio::sync::OnceCell` with a generation counter, or holding an `Arc<Mutex<Option<Future>>>` so latecomers await the in-flight refresh, dedupes cleanly when you next touch this.

### 8. `lb/containers.ts:53` — `apiEnvVars(this.env)` at field-initializer time

```ts
export class ApiContainer extends Container<WorkerEnv> {
  override envVars = apiEnvVars(this.env);
}
```

This works because `Container`'s constructor sets `this.env` before the subclass's field initializers run — but that's a `@cloudflare/containers` convention that isn't obvious from the call site. A 1-line comment ("`this.env` is populated by `Container`'s constructor before this initializer runs") saves a future debugging trip if the lib ever swaps to a lazy-env pattern.

### 9. `lb/local.ts:240` — `as unknown as WorkerEnv` hides missing API fields

```ts
return main.fetch(req, {
  FRONTEND: frontendShim,
  MODULES_BUCKET: ...,
  // ... no DATABASE_URL, METADATA_STRATEGY, S3_*, etc.
} as unknown as WorkerEnv);
```

This is safe in practice because `proxyToLocalApi` intercepts every API request before `main.fetch` is called, so `env.API_CONTAINER` and the API container env vars are never read on the local path. But the assertion silences the compiler exactly where the proof of safety lives. Two cleaner options:

- Split `WorkerEnv` into `LbEnv` (always required) + `ApiContainerEnv` (forwarded to the container), and have `main.fetch` accept `LbEnv & Partial<ApiContainerEnv>` so local only needs `LbEnv`.
- Leave a comment at the cast site naming the invariant ("`proxyToLocalApi` short-circuits all API routes, so the container-only fields are unreachable here").

### 10. Dead `google_project_iam_member` resources (follow-up cleanup)

`terraform/cloud_run_api.tf` still declares `google_project_iam_member.api_cloud_trace` (and `cloud_run_frontend.tf` the frontend equivalent), but with `--cloud_trace` gone the IAM grant is dead. Harmless, just a follow-up sweep.

### 11. `wait-container-rollout.ts` — minor

- No SIGTERM handler, so a cancelled CI run keeps polling until `--timeout`. Cheap fix: `Deno.addSignalListener("SIGTERM", () => Deno.exit(143))`.
- The response-shape fallback chain (`rollout?.status` → `active_rollout_id` → "no active rollout") is solid given the API uncertainty the PR description calls out. Worth a one-line comment that any 4xx from the `applications/{id}` endpoint means the application hasn't been created yet, so the first poll on a brand-new deploy may need to swallow a 404 (depends on whether terraform's apply blocks until creation is observable).

---

Nothing here is a hard block — (1), (2), and (3)/(4) are the ones I'd actually want answered before clicking merge. Happy to re-review once the first staging deploy validates the rollout wiring.

crowlbot (approved):
All findings from the prior review are addressed across 24217e3 and fb803e0. Quick map:

- **(1) DO migrations**: `var.migrations_version` drives `new_tag`; deploys re-send the same tag (Cloudflare no-ops), and the variable is bumped manually only when DO classes change. The first-time-bootstrap note (binding-before-migration ordering, error 100123, the comment-out-then-restore dance) is the right level of explicit. ✅
- **(2) Capacity ceiling**: `API_CONTAINER_INSTANCES = 3` is now a named constant with the trade-off (fixed fleet × pool=4, Cloud SQL budget, sleepAfter) documented at the definition site. Treating it as a deliberate cap is fine. ✅
- **(3) CI image-push gating**: unconditional push, `wrangler containers push`'s idempotence carries the cost story. Also good that `buildx` / `Cargo.lock copy` are now unconditional — the prior gating left them as a footgun if any later step needed them on the reuse path. ✅
- **(4) Bootstrap ordering**: covered by the same lb.tf comment. ✅
- **(5) METADATA_STRATEGY**: split into a clean three-variant enum plus `GCP_SERVICE_ACCOUNT_KEY`, wired through clap, container envVars, types.ts, and lb.tf. ✅
- **(6) JWT signing**: `ServiceAccount` carries a pre-parsed `EncodingKey`, parsing moved to `Client::new` returning `Result`, no per-refresh PEM parse, no panic on malformed JSON. Also nice catch with the `.trim()` on the env value. ✅
- **(8) `this.env` field-init**, **(9) local.ts cast**, **(10) dead IAM grant**, **(11) SIGTERM + retry-on-not-found**: all addressed. ✅

(7) concurrent-token-fetch race is intentionally untouched — it's a pre-existing pattern, no regression in this PR.

One residual to keep in mind, not blocking:

- The first-time-bootstrap procedure (comment out the binding + containers block, apply, restore, apply again) is documented in code but lives in a Terraform comment, which is easy to overlook during the actual cutover. If you don't already have an internal runbook for this rollout, copying that paragraph into the change-management ticket / staging-deploy plan would save a future operator from rediscovering error 100123 at 2am.

LGTM.

crowlbot (changes requested):
Re-reviewing after 6dc2165 — the wrangler-deploy refactor is a substantial pivot from the version I approved, and a couple of concrete issues popped up that I'd like resolved before this lands. The motivation is right (the cloudflare TF provider really can't drive the container application), and the "Terraform owns config values + renders wrangler.json + exposes secrets as outputs" split is a clean compromise.

In rough priority order:

### 1. `max_instances = 30` vs `getRandom(env.API_CONTAINER, 3)` mismatch

`terraform/lb.tf:140`:
```hcl
containers = [
  {
    class_name    = "ApiContainer"
    image         = "..."
    max_instances = 30
  }
]
```

But `lb/main.ts:81` still has:
```ts
const API_CONTAINER_INSTANCES = 3;
// ...
const container = await getRandom<ApiContainer>(env.API_CONTAINER, API_CONTAINER_INSTANCES);
```

`getRandom(ns, N)` picks one of `N` stable DO IDs by hashing. Each DO ID has at most one container instance, so the effective ceiling is `min(getRandom-N, max_instances) = 3` — the `30` is dead.

Either drop `max_instances` to 3 (matches steady-state capacity, no surprise headroom that won't be used) or bump `API_CONTAINER_INSTANCES` to match the new max — but if you do the latter, the long comment in `lb/main.ts:73-80` about Cloud-SQL connection budget needs updating, because 30 × pool=4 = 120 DB conns is well beyond the current planning.

This is the one item I'd treat as actually blocking, since it's a real inconsistency that masks intent.

### 2. CI deploy order: `wrangler deploy` runs before `wrangler secret bulk`

In `.github/workflows/ci.yml`:
```
- Deploy LB worker             # wrangler deploy
- Set LB worker secrets        # wrangler secret bulk
- Wait for container rollout
```

`wrangler deploy` creates/updates the worker AND triggers the container rollout immediately. At that moment, no secrets are set yet — the container boots with empty `DATABASE_URL`, `GCP_SERVICE_ACCOUNT_KEY`, `S3_*`, etc., and the Rust process either panics at startup (`expect("failed to initialize GCP client")`) or fails a DB connect.

The deploy comment says:
> the rollout becomes healthy once set

— but that assumes the rollout-status API treats the in-between-failed-boots as "still rolling" rather than "failed". If Cloudflare ever flips the application to `reverted` after N failed boots, the deploy goes red and `wait-container-rollout.ts` throws.

Two safer shapes:
- **Set secrets first, deploy second.** Requires the worker to already exist for `secret bulk` to land; would need a one-time bootstrap (`wrangler deploy --dry-run=false` once with the container config absent), but after that every subsequent run has secrets in place when the new image is rolled out.
- **Embed a startup retry loop in the container** — `Client::new` and the DB pool builder retry with backoff for the first ~60s, so a missing-secret window is tolerated. Slightly uglier but doesn't change the deploy ordering.

Either way, a staging run will reveal whether the current order works in practice, but it deserves a deliberate decision rather than "hope rollout-status absorbs the gap."

### 3. `compatibility_date = "2025-01-01"` is older than the frontend's `2026-05-19`

Looking at `cloudflare_frontend.tf:42`, the frontend worker is pinned to `2026-05-19`. The LB worker here lands at `2025-01-01` — roughly 17 months behind. Cloudflare Containers + Durable Objects + recent Workers API surface have all gotten bug fixes after Jan 2025. Was this date copy-pasted from somewhere, or are there compat-flag concerns blocking a newer date? Aligning with the frontend (or the latest stable) avoids hitting "fixed in compat 2026-xx-yy" issues during the rollout.

### 4. `observability.logs.enabled` is implicitly defaulted

Old config:
```hcl
logs = {
  enabled         = true
  invocation_logs = true
  persist         = true
}
```

New config:
```hcl
observability = {
  enabled            = true
  head_sampling_rate = 0.01
  logs               = { invocation_logs = true }
}
```

`observability.enabled = true` enables tracing; `observability.logs.enabled` is a separate toggle. Without an explicit `logs.enabled = true`, wrangler's default may or may not match the prior behavior, and `persist = true` is gone too. If you didn't intend to drop log persistence, spell it out:

```hcl
logs = {
  enabled         = true
  invocation_logs = true
  persist         = true
}
```

### 5. `routes[].zone_name` vs the prior `zone_id`

The old Terraform routes used `var.cloudflare_zone_id`; the new wrangler routes use `zone_name = var.domain_name`. That works *if* `var.domain_name` is literally the apex zone (`jsr.io`) — but `npm_domain` (e.g. `npm.jsr.io`) lives in the same zone, so the third route's `zone_name = var.domain_name` is only right because the apex domain *is* the zone name. The mapping breaks the moment someone introduces a route whose pattern is in a different zone (e.g. a `*.deno-registry-staging.net` setup where the apex isn't `var.domain_name`). Using `zone_id = var.cloudflare_zone_id` is what you already have and is less error-prone.

### 6. `local_file.lb_wrangler_config` writes into the repo

`local_file` produces `lb/wrangler.json` on every apply. With this in `.gitignore` and CI being the only operator, fine — but if a developer runs `terraform apply` against staging from a workstation (e.g. to inspect a plan), they'll generate a config pinned to whatever `api_image_tag` they passed (or didn't pass — `var.api_image_tag` has no default now). A `terraform plan` will demand a value for `api_image_tag` and otherwise fail loudly, which is OK as a footgun guard. Worth a one-line comment at `local_file.lb_wrangler_config` saying "intended to be rendered from CI; running `terraform apply` locally will overwrite the file."

### 7. Stale comment in `lb/main.ts`

```ts
// Re-export the container class so the Worker runtime (and the
// `cloudflare_worker_version` Durable Object binding) can discover it.
export { ApiContainer };
```

`cloudflare_worker_version` doesn't exist anymore in this PR's terraform — the binding now lives in the wrangler.json that terraform renders. Update the comment to refer to `lb/wrangler.json` (or to "the worker's DO binding") so a future reader doesn't grep for the resource and come up empty.

### 8. `tools/lb-secret-bulk.ts` — small things

- `terraform -chdir=terraform output -raw` assumes the state is already initialized in CI. That's true today (the prior `terraform init` step seeds it), but a one-line comment at the `tfOutput` definition documenting the precondition saves a future reader some debugging when running this outside CI.
- The script uses `await tfOutput(...)` inline in the `secrets` object literal, which means failures throw at object construction and the temp file is never created (so cleanup isn't needed) — that's fine, just worth noting that the `try/finally` block's safety relies on the awaits being inside it. Currently they're outside; if any `await tfOutput(...)` throws, the process exits before the file exists. Today that's OK; if you ever reorder, the leak window opens up. Minor.

### 9. Bootstrap on a brand-new environment

The previous version had the explicit "comment out binding + container block, apply, restore, apply again" dance for the error 100123 binding-before-migration issue. Wrangler may or may not handle this differently — `wrangler deploy` with a `migrations` block and a `durable_objects.bindings` entry referencing the new class in the same call could hit the same API path. Worth a single bootstrap test on a fresh CF account/test environment, or at minimum noting in this PR's description that the bootstrap behavior is "to be verified on first deploy."

### 10. CI is currently red on `docker-images` for the same `CLOUDFLARE_API_TOKEN` reason flagged in the original PR description

```
ERROR: In a non-interactive environment, it's necessary to set a CLOUDFLARE_API_TOKEN environment variable for wrangler to work.
```

The `docker-images` job has no `environment:`, so `${{ secrets.CLOUDFLARE_API_TOKEN }}` evaluates to empty. The "⚠️ To confirm before merge" bullet in the PR description called this out — it now needs to actually be resolved (either give the job an `environment:` or use a repo-level secret rather than an environment-level one).

---

(1)–(2) are the ones I'd want fixed before re-approving; (3)–(6) are real but lower-priority; (7)–(10) are quick cleanups / clarifications. Treating this as a fresh review since the refactor moved the deploy story enough that my earlier approval doesn't carry forward.

crowlbot (approved):
All the substantive items from the previous round are addressed in ac8f96d:

- **(1)** `max_instances = 3` matches `API_CONTAINER_INSTANCES` with the in-place comment locking the two together; ✅
- **(2)** the deploy-order comment is a clear and convincing explanation — secrets persist across deploys, so steady-state has no missing-secret window, and `secret bulk` is rotation-only. The first-deploy gap is called out explicitly; ✅
- **(3)** `compatibility_date = "2026-05-19"` with a "keep in sync with cloudflare_frontend.tf" cross-reference; ✅
- **(4)** `observability.logs.enabled = true` restored. (Wrangler config doesn't surface `persist` the way the cloudflare TF resource did — that field was a provider-side knob, not a wrangler one — so dropping it here is correct, not a regression);  ✅
- **(5)** routes back to `zone_id = var.cloudflare_zone_id`; ✅
- **(6)** the "CI-rendered; bare local apply will overwrite the file" note at the top of `local_file.lb_wrangler_config`; ✅
- **(7)** stale `cloudflare_worker_version` comment in `lb/main.ts` fixed to point at the generated wrangler.json; ✅
- **(8)** `tfOutput` precondition documented; ✅

Plus the orthogonal-but-correct addition of `CLOUDFLARE_ZONE_ID` flowing through to the API container's env (consistent with how the Cloud Run service is already configured for the new cache-purge path in #1410).

A couple of small leftovers, none of them blocking:

- **PR description is stale.** The "Why this isn't a straight cherry-pick of #1378" and "Changes → Terraform → LB worker migrated to the version/deployment trio" sections still describe the pre-refactor architecture, and the ac8f96d CI comment references "the PR notes on first-environment bootstrap" — but those notes don't exist in the body. Worth a quick edit so a future bisecter or operator reading this PR understands the deployment model they're inheriting.
- **`docker-images` CI is still red** on the original `CLOUDFLARE_API_TOKEN` / `CLOUDFLARE_ACCOUNT_ID` issue (already flagged in your "To confirm before merge" list). The fix is repo-settings-side, not code: either give the job an `environment:`, or move those two values from environment-scoped secrets to repo-level. Worth pinning down before the first staging cutover, since without it the deploy jobs will always `skipping`.

LGTM otherwise — re-approving.

crowlbot (approved):
The CI fix is exactly the right shape: `wrangler containers push` now reads from `secrets.CLOUDFLARE_DEPLOY_API_TOKEN` (a repo-level token), while the deploy jobs keep their environment-scoped `CLOUDFLARE_API_TOKEN`. `docker-images` is green at last. (The 4 intermediate iteration commits make the bisect-noise tradeoff in favor of clarity, which seems right here.)

Approval re-applied — the net code delta since 02:28Z is one line, and it's the line I asked for.

One thing to keep an eye on, but not blocking: the `staging` deploy now actually runs (since docker-images is green) and the e2e job is red — every HTML response from `https://deno-registry-staging.net/...` is a 500. Two plausible causes worth ruling out before merge:

1. **First-environment bootstrap.** This is the very first staging deploy through the new wrangler path, so the LB worker is being created brand-new and the API container's first boot lands before `wrangler secret bulk` has uploaded `DATABASE_URL` / `GCP_SERVICE_ACCOUNT_KEY` / `S3_*`. The Rust `Client::new` panics on a missing GCP key (`expect("failed to initialize GCP client")`), the container crashes, the rollout-status API may still report "completed" if it considers the new revision live, and the LB then hands traffic to a container that's looping on startup → 500s. A rerun of the staging job after secrets are persisted should clear this — that's the exact scenario the comment block in ci.yml describes.

2. **`max_instances = 3` + `getRandom(env.API_CONTAINER, 3)` not actually producing a healthy fleet on first deploy.** Less likely (the wiring matches your last fix), but worth confirming the three DO instance IDs are all reachable rather than only the one that booted successfully.

If a quick CI re-run on the staging branch returns green, (1) is confirmed and you're good to go. If it stays red, the rollout-status polling in `wait-container-rollout.ts` may need to also check application *health* (instances on target version + non-zero healthy count) rather than just rollout completion — a healthy-instance count of 0 should not be a "rollout complete" signal.

Either way: the code is in good shape. I'd treat the e2e red as the "real staging run to validate the wiring" item already flagged in the PR description rather than a blocker on the diff itself.

crowlbot (approved):
Good detective work on the staging shakeout — all four commits read as real findings, not speculation:

- **05cbc08 + 30ebf0a**: custom `instance_type` (1 vCPU / 3 GiB / 2 GiB disk) replaces the 256 MiB lite default that OOM'd on boot. The 3 GiB-per-vCPU floor is documented at the call site so the next person doesn't second-guess it. ✅
- **2b6638c**: `wrangler secret bulk` before `wrangler deploy` — exactly the inversion my prior re-review was worried about. The brand-new-environment bootstrap note (deploy once to create the worker, then secrets, then redeploys are normal) is the right caveat. ✅
- **10e3bca**: the right shape of fix for the connectivity gap given Cloud Run was on the VPC connector. Using `data "http"` to read live `cloudflare.com/ips-v4`, staging-only, with a "client certificate is next" promise. ✅

Re-approving. A few things I'd want answered in a follow-up — none blocking this PR, but worth tracking explicitly:

### 1. Cloud SQL connection-budget math (the comment that disappeared)

The old `API_CONTAINER_INSTANCES` comment carried a warning: *"Sizing must stay under the Cloud SQL connection budget shared with the Cloud Run tasks service."* At `max_instances = 30 × --database_pool_size = 4`, the steady-state ceiling is now **120 connections** from the serving path alone, on top of whatever the Cloud Run **tasks** service holds open. Default `max_connections` for PG15 on Cloud SQL is tier-dependent (often 100-200 for smaller tiers); the new comment dropped the budget-warning. Worth a sanity check that:

- `db-custom-N-...` tier currently in use has headroom for `30 × 4 + (tasks footprint) + (admin/superuser reserve, typically 3-15)`
- and a follow-up to either tune `--database_pool_size` downward or move to PgBouncer/Cloud SQL connection pooling before this lands in prod.

### 2. `data "http" "cloudflare_ipv4"` introduces a plan-time dependency on cloudflare.com

`terraform plan` now reaches out to `cloudflare.com/ips-v4` on every run. If that endpoint is unavailable (rare but real — e.g. global Cloudflare incident), plans fail. Two cheap hardenings:

- Cache the response into `terraform.tfvars` (or a tracked `cloudflare_ipv4.txt`) and let CI refresh it weekly via a scheduled workflow — `data "http"` becomes a fallback, not the source of truth.
- Or pin the response with a checksum to detect drift between plans on the same SHA.

Not urgent for staging, but worth deciding before prod where a plan failure during an incident would compound the outage.

### 3. The Cloudflare egress allowlist is shared across all CF Workers customers

A "client certificate is the next hardening step" is the right promise — just want to make sure it lands before the prod toggle flips. Anyone routing through Cloudflare's network can hit the staging DB IP from those allowed ranges; only the DB password protects the door right now. Putting this in the prod cutover checklist (not as a code blocker) is enough.

### 4. Brand-new-environment bootstrap should make it into the PR description

The CI comment now documents the secrets-before-deploy ordering and acknowledges the one-time "deploy → secrets → redeploy" bootstrap. That note belongs in the PR body too, so the operator doing the prod cutover doesn't have to read commit history to discover it.

### 5. Staging e2e is still red, but the failures look unrelated to the API container

Looking at the failing run, the API-container path is healthy: every "should return **plain text** 200" assertion against `/@luca/flag/...meta.json` PASSES (those are the LB → API container proxy paths). The failures are all "should return **html** 200" — i.e. the LB → frontend-service-binding path returning something other than 200 for HTML-rendered routes.

The wrangler-rendered `services = [{ binding = "FRONTEND", service = "${var.gcp_project}-jsr-frontend" }]` looks right, and the deployed frontend worker still has that name (per cloudflare_frontend.tf). I'd guess one of:

- Staging's frontend worker hasn't been re-deployed after the LB rebuild and is referencing a stale LB binding/version (the frontend reaches back to the LB via the `LB` service binding for API requests — see `frontend/utils/api.ts`).
- The container is healthy but slow to respond on the first request after a long sleep, and the frontend's API fetch is timing out → the frontend renders a 5xx page.

If it's the first one, a single frontend redeploy clears it. If it's the second, the frontend's API-fetch timeout deserves a look — but neither is on this PR.

---

LGTM otherwise — the code-level changes are clean and the staging findings are exactly the right shape.

crowlbot (approved):
The trajectory has fully stabilized — and the staging-shakeout commits ended up being the most informative part of the PR. Quick rundown of what I see:

- **c59557a** (`rustls 0.23.23 pin`): really good debugging — the chain `Cloud SQL public IP → server cert SAN is `*.sql.goog` not the IP → sqlx must use `verify-ca` → sqlx 0.8.6's NoHostnameTlsVerifier only catches the old `CertificateError::NotValidForName` while rustls 0.23.24+ returns `NotValidForNameContext` → handshake aborts with `SSLV3_ALERT_BAD_CERTIFICATE`` is exactly the kind of finding that's nearly impossible to spot without a real run. The pin is documented and includes a "drop the pin once a sqlx > 0.8.6 lands" exit ramp; bisecting it later won't need archaeology. ✅
- **01d5bf6 + e08479e**: phase 2 lands, the cert is now required on both environments, Cloud Run gets the cert too (so its private-VPC path keeps working after the flip). Prod's DB is now `0.0.0.0/0` + `TRUSTED_CLIENT_CERTIFICATE_REQUIRED` — a deliberate trade: the cert is the access boundary instead of a network ACL, because CF Container egress isn't pinnable without the Zero Trust dedicated-egress add-on. With CI green end-to-end including staging e2e, this is in good shape.

Re-approving (5th pass — sorry, the iteration on this PR ate a lot of review states).

A handful of operational items I'd flag for the runbook / a follow-up PR, none of them blocking this one:

### 1. Prod DB threat model shift

Previously the DB was VPC-private; now it's reachable from anywhere on the internet, with mTLS doing the gating. Same control surface as a managed-Postgres provider (RDS / Supabase / Neon) — fine on its own, but the on-call runbook should reflect that the new attack-surface story is:

- TLS pre-handshake reachability of `db-custom-4-8192:5432` from anywhere
- Client-cert validation correctness in Cloud SQL's stack
- Cert/key safekeeping (Cloud Run env + Cloudflare worker secrets)

…rather than "the DB is private; you'd need a Cloud VPN to even reach it." Worth one Slack pin / runbook update so the next on-call knows.

### 2. Client-cert rotation

`google_sql_ssl_cert.api` has no rotation policy in this PR. If the cert key leaks (via env var dump, secret-scanner false-negative on a paste, etc.), the only revocation is `terraform taint google_sql_ssl_cert.api && terraform apply`, which forces every consumer to redeploy. Worth establishing:

- A documented rotation cadence (12–24 months) before the auto-cert expires
- A "rotate now" runbook that's been dry-run at least once in staging

For a "client cert is the boundary" model, this is the equivalent of password rotation hygiene.

### 3. `max_connections = 150` headroom check

With `max_instances = 30 × --database_pool_size = 4`, the serving path's steady-state ceiling is 120 connections. Adding `--database_pool_size = 1` for the **tasks** service (4 instances at most, in practice 1-2 active = 1-4 conns), plus admin / `pg_stat_activity` poll connections (3-15), prod is at roughly `120 + 5 + 10 ≈ 135 / 150 = 90%` saturation under steady-state. That's slim — a burst pushes over. Either:

- Bump `max_connections` to 200, or
- Drop `max_instances` to ~25 (still 4× current `--database_pool_size` headroom), or
- Move to PgBouncer / Cloud SQL's connection pooling so the pool size doesn't multiply by the instance count.

Not urgent for staging where load is low, but the first big prod traffic spike will surface this.

### 4. `let _ = rustls::crypto::ring::default_provider().install_default();`

The `let _` swallows the `Err(CryptoProvider)` returned if a provider is already installed. Today nothing else installs one before `main`, so this is fine. But: if some future dep starts installing aws-lc-rs in a `ctor`/static initializer, this would silently keep that provider — and you'd be back to "TLS handshakes fail in firecracker without an obvious cause." Either:

- `.expect("ring must be the rustls default provider")` to make a future regression noisy
- or leave a comment naming the invariant explicitly

Minor — the explicit `Cargo.toml` comment already does most of this work.

### 5. PR description is stale (still)

It still describes the original Terraform-managed `cloudflare_worker_version` / `cloudflare_workers_deployment` trio with the `containers = [{ class_name }]` attachment, the targeted-apply/promote/wait flow, and `METADATA_STRATEGY = json blob`. None of that's true in the merged state. Replacing the body with a short summary of where the PR landed (wrangler-deploy + TF-rendered wrangler.json, mTLS DB, secrets-before-deploy ordering, the bootstrap caveat) would save the next operator a `git log -p` archaeology session.

---

The CI failures over the last day were exactly the validation loop a PR of this scope needs, and the commits read as a clean debugging story rather than thrashing. LGTM.

crowlbot (commented):
Reviewing the new OTLP/HTTP traces work on top of the previously-approved Cloudflare-Containers migration (4 commits, +205/-39). The shape is good — `opentelemetry-otlp` is correctly switched to `http-proto` + `reqwest-client` (the Grafana Cloud gateway is HTTP-only and reqwest is the stack the rest of the crate uses), and the endpoint/headers split between non-secret wrangler var and bulk-pushed worker secret is clean.

A few things worth fixing before this lands:

### ⚠️ 1. Empty `OTLP_ENDPOINT` is silently treated as "tracing enabled"

`lb/containers.ts:apiEnvVars` always emits both keys with an empty-string fallback:

```ts
OTLP_ENDPOINT: env.OTLP_ENDPOINT ?? "",
OTLP_HEADERS: env.OTLP_HEADERS ?? "",
```

In `api/src/main.rs`:

```rust
let export_target = if let Some(endpoint) = config.otlp_endpoint {
  TracingExportTarget::Otlp { endpoint, headers: ... }
} else {
  TracingExportTarget::None
};
```

With clap-derived `env = "OTLP_ENDPOINT"`, a *set-but-empty* env yields `Some("")`, not `None`. So when telemetry isn't configured for the environment (the terraform `var.otlp_endpoint != ""` guards correctly skip wrangler vars / bulk secrets), the container still starts with `OTLP_ENDPOINT=""`, which routes into the `Otlp` arm and builds the HTTP exporter against `otlp_signal_endpoint("", "/v1/traces")` → `"/v1/traces"` (no scheme/host). That contradicts the `// Export is disabled when unset` contract in the config docstring and at best logs export errors on every batch.

Two clean fixes; I'd take both:

- **In `lb/containers.ts`**, conditionally include the keys rather than emitting empty:
  ```ts
  ...(env.OTLP_ENDPOINT ? { OTLP_ENDPOINT: env.OTLP_ENDPOINT } : {}),
  ...(env.OTLP_HEADERS ? { OTLP_HEADERS: env.OTLP_HEADERS } : {}),
  ```
- **In `api/src/main.rs`**, treat empty as unset:
  ```rust
  let export_target = match config.otlp_endpoint.as_deref() {
    Some(e) if !e.is_empty() => TracingExportTarget::Otlp {
      endpoint: e.to_string(),
      headers: crate::tracing::parse_otlp_headers(config.otlp_headers.as_deref()),
    },
    _ => TracingExportTarget::None,
  };
  ```

Cloud Run is fine — `local.otlp_envs` in `terraform/cloud_run_api.tf` skips the keys when empty. This only bites on the CF-Container path.

### ⚠️ 2. Terraform variable doc and Rust config doc disagree on URL form

`variables.tf` describes `otlp_endpoint` as a *full* signal URL:

```hcl
description = "OTLP/HTTP traces endpoint, full URL including the path (e.g. ...otlp/v1/traces)."
```

But `api/src/config.rs` describes it as the *base* (OTEL `OTEL_EXPORTER_OTLP_ENDPOINT` style) with `/v1/traces` appended automatically, and the `otlp_signal_endpoint` helper exists specifically to make the base form work — that's also what `de371fe`'s commit message says ("fix: append /v1/traces to the OTLP base endpoint"). Both forms work, but the terraform description directs operators to a form that's the opposite of the (recommended-and-named) base form that `OTEL_EXPORTER_OTLP_ENDPOINT` semantics imply.

Suggest matching the config docstring's wording: base URL preferred (e.g. `https://otlp-gateway-<zone>.grafana.net/otlp`), full signal URL also accepted.

### nits

**3. `otlp_signal_endpoint` partial-path edge case (`api/src/tracing.rs`).** The "already carries the signal path" check is `base.ends_with(signal_path)`. An operator who pastes `https://x/otlp/v1` (no `/traces`) — easy to do given Grafana's docs sometimes show `OTEL_EXPORTER_OTLP_TRACES_ENDPOINT` ending at `/v1/traces` and they trim one segment — produces `https://x/otlp/v1/v1/traces`, which 404s. One- or two-line comment in the docstring noting that only the bare base or the exact signal URL are recognized would save a future debugging session.

**4. Both the Cloud Run API service and the CF-Container API will export traces concurrently while Cloud Run remains as the fallback** (per the original commit message). The `OTLP_ENDPOINT` / `OTLP_HEADERS` are wired into both `cloud_run_api.tf` and the LB container env, so any cutover period will roughly double the volume Grafana sees. Not a bug — just worth confirming the Grafana Cloud quota / sampling is sized for that, and that traces from both paths are distinguishable (the `service.name = "api"` resource attr is the same; consider an additional resource attr like `deployment.environment` or `service.instance.id` set differently per path).

**5. `parse_otlp_headers` swallows malformed pairs silently.** A typo like `OTLP_HEADERS=Authorization Basic dXNlcjpwYXNz` (space instead of `=`) gives an empty header map and the export then fails 401 against Grafana. A warning log when the map ends up empty *and* the raw input was non-empty would make this faster to diagnose. Tiny.

**6. `update lockfile` commit (`fbf06f1`)** has no message body, only the headline; that's fine, but I'd typically squash it into the OTLP feat commit at merge time so the lockfile change isn't dangling in history.

---

The OTLP test suite for `otlp_signal_endpoint` / `parse_otlp_headers` is a nice touch — covers the right cases (trailing slash, already-has-path, `=` in base64 padding). Once #1 and #2 are in, this is ready from my side.

crowlbot (approved):
Reviewing the three sizing commits on top of the previously-approved migration (`d538393` scale to 12×standard-1, `3c9567d` DB-region pin, `6eaad65` max_instances 24). The direction is right and this is where the real Cloud Run → Containers cost win lands — going from 30×(1 vCPU/3 GiB custom) to 12×standard-1 (½ vCPU) is a ~5× cut in provisioned CPU at steady state. Staging CI is green. Approving, with three things to watch at prod cutover:

**1. ½ vCPU per container is the one number I'd watch.** The old comment had the API at ~90% of a full vCPU on Cloud Run. With the 12-way `getApiContainer` fan-out each container only sees ~1/12 of traffic, so aggregate 6 vCPU should be ample in steady state — but a single container can't burst past ½ vCPU, so a hot request (heavy `deno_doc`/`deno_graph` path that slips onto the serving container) could tail-latency-spike. Worth keeping an eye on container CPU-saturation / p99 right after cutover; the predefined ladder (basic ¼ → standard-1 ½ → standard-2 1) makes it a one-line bump if needed.

**2. The "~96 DB conns, well under 150" note is container-only.** Adding the rest of what shares the `max_connections=150` budget:
- tasks service: `max_instance_count=20 × --database_pool_size=1` = up to 20
- Cloud Run **serving fallback** (still present): `min 1 × pool 4` = 4 idle, but `max 30 × 4` = 120 if it ever scales

Realistic worst case is ~96 + 20 + 4 ≈ **120/150** — still safe, but the headroom is ~30, not ~54, and it's only safe *because the serving fallback stays drained* (lb no longer routes to it, so it sits at min=1). Might be worth wording the comment as "well under 150 given the serving service is drained," and it reinforces the existing follow-up to tear that service down — once it's gone the budget is comfortable again.

**3. `enam` for us-central1** is the right call given there's no central-NA hint; enam/wnam are both roughly equidistant and the latency delta is small next to the DB round-trip you're saving. Nice catch that the region-prefixed DO name (`enam-instance-N`) is what forces the DOs to be re-created with the hint in effect — without the rename the existing DOs would keep their old placement.

Minor: `getApiContainer` keeps `Math.random()` index selection so there's no session-affinity change vs `getRandom`, and warm-instance reuse via the stable per-index id is preserved. ✔ One sanity check — confirm the API image fits standard-1's 8 GB disk (the old custom type set `disk_mb=2048`), but that's almost certainly fine.

PR #1423, opened 2026-06-03
feat: export traces via OTLP/HTTP instead of gRPC and Cloud Trace

crowlKats:
## Summary

Extracts the OpenTelemetry changes from #1406 into a standalone PR so they can land independently of the Cloudflare Container migration.

Switches the trace exporter from **OTLP/gRPC (tonic)** and **Google Cloud Trace** to **OTLP/HTTP (protobuf) over reqwest**, the same HTTP stack the rest of the crate already uses. This is what the managed Grafana Cloud OTLP gateway speaks.

## Changes

- **`api/src/tracing.rs`**
  - `TracingExportTarget::Otlp` now carries `{ endpoint, headers }`; the gRPC exporter is replaced with the HTTP/protobuf exporter (`.http()` + `Protocol::HttpBinary` + `.with_headers(...)`).
  - `otlp_signal_endpoint()` appends the per-signal path (`/v1/traces`) to a base endpoint, OTEL `OTEL_EXPORTER_OTLP_ENDPOINT` style (the 0.12 HTTP exporter doesn't do this itself). Tolerates a trailing slash / an endpoint that already carries the path.
  - `parse_otlp_headers()` parses the `OTEL_EXPORTER_OTLP_HEADERS` `key=value,...` format, splitting on the first `=` only so base64 padding in a `Basic` auth header survives.
  - Removes the `CloudTrace` export target. Adds unit tests for both helpers.
- **`api/src/config.rs`**
  - `otlp_endpoint` is now the base OTLP/HTTP endpoint; new `otlp_headers` (`OTLP_HEADERS`) carries backend auth. Removes `--cloud_trace` and the now-unused `gcp_project_id`.
- **`api/src/main.rs`**: build the `Otlp { endpoint, headers }` target from config; drop the `cloud_trace` branch.
- **`api/Cargo.toml`**: `opentelemetry-otlp` now uses `default-features = false` + `http-proto` + `reqwest-client` (no `grpc-tonic`); removes `opentelemetry-gcloud-trace`.
- **`Cargo.lock`**: drops the tonic/prost/gcloud-trace dependency tree.

## Not included from #1406

Everything else — the Cloudflare Container migration, DB mTLS, the GCP service-account-key metadata strategy, the `rustls` pin, LB worker, Terraform, and CI changes — stays in #1406.

## Testing

- `cargo build` ✅
- `cargo test --bin registry_api tracing::` — 5/5 pass ✅
- `cargo fmt --check` ✅


crowlbot (approved):
Nice to see the OTLP work split out so it can land independently of the Container migration. I'd reviewed this same code on #1406, and the feedback from there is all addressed here:

- `parse_otlp_headers` splits on the **first** `=` only (`split_once`), so a `Authorization=Basic <base64==>` value keeps its padding — and there's now a test pinning exactly that.
- `otlp_signal_endpoint` tolerates both a trailing slash and an endpoint that already carries `/v1/traces`, also tested.
- `otlp_headers` is redacted (`***`) in the `Debug` impl, so the `println!("{config:?}")` at startup no longer risks leaking the backend auth.

I also checked that the `CloudTrace` / `gcp_project_id` removal is complete — no other references to either remain in `api/`, and CI (check/test/frontend/worker) is green. The `.http()` + `Protocol::HttpBinary` + `reqwest-client` switch is the right call for the Grafana Cloud OTLP gateway.

One non-blocking nit (carryover from my #1406 note):

**Empty `OTLP_ENDPOINT` is treated as "enabled".** clap parses a *present-but-empty* env var as `Some("")`, so `OTLP_ENDPOINT=""` takes the `if let Some(endpoint)` branch rather than falling through to `None`. `otlp_signal_endpoint("", "/v1/traces")` then yields a schemeless `"/v1/traces"`, and `install_batch(...).unwrap()` will panic the process at boot. On #1406 the wrangler env emitted `OTLP_ENDPOINT: env.OTLP_ENDPOINT ?? ""`, which is exactly the shape that would trip this — and that plumbing isn't in this PR, so it's worth hardening the Rust side regardless:

```rust
let export_target = if let Some(endpoint) =
  config.otlp_endpoint.filter(|s| !s.trim().is_empty())
{
  TracingExportTarget::Otlp { endpoint, headers: ... }
} else {
  TracingExportTarget::None
};
```

That makes an empty value behave like unset (export disabled) instead of crashing at startup.

LGTM otherwise — approving.

nathanwhit (approved):

PR #1424, opened 2026-06-03
feat: export Worker traces and logs via native Cloudflare Observability

crowlKats:
## Summary

Exports the Cloudflare Workers' **traces and logs** to an external OTLP backend (e.g. Grafana Cloud) using Cloudflare's **native Workers Observability** export — not a custom in-Worker OTLP exporter.

Cloudflare exports both automatic request **traces** and **logs** (console output + invocation logs) to a "destination" you configure with an OTLP endpoint + auth headers. The Worker config just references that destination by name; no exporter code runs in the Worker.

## Changes

- **`terraform/variables.tf`**: add `otlp_traces_destination` and `otlp_logs_destination` — the names of the dashboard-managed destinations. Empty (default) keeps export disabled, with Cloudflare still storing traces/logs in its own dashboard.
- **`terraform/cloudflare_frontend.tf`** and **`terraform/lb.tf`**: add a `traces` block and `logs.destinations` to each worker's `observability` block, pointing at the named destinations when the vars are set.
- **`.github/workflows/ci.yml`**: pass `TF_VAR_otlp_traces_destination` / `TF_VAR_otlp_logs_destination` (from non-secret GitHub Actions `vars`) on both staging and prod applies.

Covers **both** workers (frontend + lb) — native tracing needs no per-worker code.

## Required setup (one-time, per Cloudflare account)

The OTLP endpoint + auth headers ("the provided values") live in a **dashboard destination**, because the `cloudflare_workers_observability_destination` terraform resource is advertised but **not yet implemented** in the pinned provider (5.19.1) — see [cloudflare/terraform-provider-cloudflare#7127](https://github.com/cloudflare/terraform-provider-cloudflare/issues/7127).

1. In the Cloudflare dashboard → Workers & Pages → Observability → Telemetry, **add a destination** (type Traces, and one for Logs) with the OTLP endpoint URL and auth header(s) (e.g. `Authorization: Basic <base64>` for Grafana Cloud).
2. Set the GitHub Actions repo/environment **variables** `OTLP_TRACES_DESTINATION` and `OTLP_LOGS_DESTINATION` to those destination names.

Until both are done, behaviour is unchanged (Cloudflare dashboard storage only).

## Notes

- The frontend's existing hand-rolled tracer (`frontend/utils/tracing.ts`) is left as-is and stays dormant (it only exports if `OTLP_ENDPOINT` is set, which it isn't). With native export now handling traces, that module is a candidate for removal in a follow-up.

## Testing

- `terraform validate` + `terraform fmt` ✅ (provider 5.19.1 accepts the `observability.traces` / `destinations` schema on both `cloudflare_worker` and `cloudflare_workers_script`)


crowlbot (commented):
Companion to #1423 — and I like that this leans on Cloudflare's native Workers Observability export rather than an in-Worker exporter, so there's no per-worker exporter code to maintain. Keeping the endpoint+auth in a dashboard destination referenced by name also means the CI plumbing only needs non-secret `vars`, not `secrets` — the right call. CI is green and the change is a no-op until the destinations + repo vars are set up.

One behavioral thing I'd like to confirm, plus a couple of small notes:

### Disabled-state asymmetry between `logs` and `traces`

For **logs** you keep the block enabled and only toggle the export target:

```hcl
logs {
  enabled = true
  ...
  destinations = var.otlp_logs_destination != "" ? [var.otlp_logs_destination] : null
}
```

so when the var is empty, logs still record to the CF dashboard.

For **traces** you instead null the *entire* attribute:

```hcl
traces = var.otlp_traces_destination != "" ? { enabled = true, ... } : null
```

so in the disabled state there's no `enabled = true` for traces at all. The description says "Cloudflare still stores traces/logs in its own dashboard" when the destination is empty — but nulling the whole `traces` object looks like it would leave trace collection unconfigured (off / provider-default) rather than dashboard-on / export-off, which is what the logs branch gives you.

Is disabling trace collection entirely (including the dashboard) when there's no external destination intentional? If you want traces to mirror logs (always collected to the dashboard, external export toggled), it'd be:

```hcl
traces = {
  enabled            = true
  head_sampling_rate = 0.01
  persist            = true
  destinations       = var.otlp_traces_destination != "" ? [var.otlp_traces_destination] : null
}
```

Either is defensible — just want the disabled-state behavior to match the doc.

### Minor
- `head_sampling_rate = 0.01` (1%) is set in both workers' `traces` — probably intentional for the traffic level, just worth a sanity check that 1% yields enough trace volume to be useful.
- Pre-existing, not this PR: `lb.tf`'s **`logs`** block carries a `head_sampling_rate`, which reads like a traces-only concept and may be a no-op there — worth a glance while you're in the area.
- As you note, `frontend/utils/tracing.ts` becomes dead once native export covers traces — good follow-up candidate.

`terraform validate` passing is reassuring, but it only confirms schema shape, not that the CF API accepts `observability.traces` / `logs.destinations` on both `cloudflare_worker` and `cloudflare_workers_script` at apply — so worth watching the first staging apply.

Happy to approve once the traces disabled-state intent is confirmed.

nathanwhit (approved):

PR #1438, opened 2026-06-08
fix: dont use prepared statements in workers-rs

crowlKats:
### PR Checklist

- [ ] The PR title follows
- [ ] Is this closing an open issue? If so, link it, else include a proper
      description of the changes and reason behind them.
- [ ] Does the PR have changes to the frontend? If so, include screenshots or a
      recording of the changes.
      <br/>If it affect colors, please include screenshots/recording in both
      light and dark mode.
- [ ] Does the PR have changes to the backend? If so, make sure tests are added.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -272,7 +272,7 @@ jobs:
       - name: Build API Cloudflare Worker
         run: |
           cargo install worker-build --locked
-          worker-build --release
+          worker-build --release --panic-unwind
         working-directory: workers-rs
 
       - name: terraform plan
@@ -356,7 +356,7 @@ jobs:
       - name: Build API Cloudflare Worker
         run: |
           cargo install worker-build --locked
-          worker-build --release
+          worker-build --release --panic-unwind
         working-directory: workers-rs
 
       - name: terraform plan
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -272,7 +272,11 @@ jobs:
       - name: Build API Cloudflare Worker
         run: |
           cargo install worker-build --locked
-          worker-build --release --panic-unwind
+          # NOTE: `--panic-unwind` is omitted: wasm-bindgen 0.2.122 emits
+          # UnwindSafe bounds under `-Cpanic=unwind` that wasm-streams 0.5.0
+          # (a `worker` dep) does not satisfy, so it fails to compile. Panics
+          # are surfaced via console_error_panic_hook instead (see src/lib.rs).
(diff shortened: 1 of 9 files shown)


crowlbot (commented):
The `simple_query` migration is the right fix for the Hyperdrive connection-pooling desync (fresh client per request restarting prepared-statement counters while the backend connection is reused). The explanation in the inline comment is accurate and the `rows()` / `text()` helpers make the text-protocol trade-off explicit. `console_error_panic_hook` and the `invocation_logs = true` changes are straightforward.

One issue to fix before merging:

**`wrangler.toml` / CI inconsistency with `--panic-unwind`**

The CI steps explicitly run `worker-build --release` (no `--panic-unwind`) and add a comment explaining why — wasm-bindgen 0.2.122 emits `UnwindSafe` bounds under `-Cpanic=unwind` that wasm-streams 0.5.0 doesn't satisfy, so it fails to compile. But `wrangler.toml` is changed to `worker-build --release --panic-unwind`, which means anyone running `wrangler build` or `wrangler deploy` locally would hit exactly the compilation failure that CI avoids. The `wrangler.toml` change should drop `--panic-unwind` to stay consistent with CI (leaving it as `worker-build --release`), unless the underlying wasm-bindgen/wasm-streams compatibility issue is also being resolved here.

Also: the PR title "debug cloudflare" doesn't follow conventional commits format ("fix: ..." or "chore: ..."), which is why the Validate PR title CI check is failing.

crowlbot (changes requested):
The `simple_query` migration is correct and well-documented — the comment block explaining the Hyperdrive/prepared-statement wire-desync is exactly what a future reader needs. `console_error_panic_hook` is a good addition for observability, and the `invocation_logs` enablement in Terraform is sensible.

One thing still needs fixing before merge:

**`wrangler.toml` adds `--panic-unwind` but that flag fails to compile with the current dependency set.**

```diff
 [build]
-command = "worker-build --release"
+command = "worker-build --release --panic-unwind"
```

The CI workflow explicitly omits `--panic-unwind` and explains why:

> wasm-bindgen 0.2.122 emits UnwindSafe bounds under `-Cpanic=unwind` that wasm-streams 0.5.0 (a `worker` dep) does not satisfy, so it fails to compile.

So anyone who runs `wrangler build` or `wrangler dev` locally will hit a compile error, because `wrangler` uses the `[build]` command from `wrangler.toml`. The `console_error_panic_hook` already gives you panic visibility without needing `--panic-unwind`, so please drop the flag from `wrangler.toml` too.

PR #1444, opened 2026-06-09
fix: cut docs/diff origin render cost (noindex + per-IP cache-miss rate limit)

crowlKats:
## Why

docs/diff/source origin load is dominated by **crawler cardinality**, confirmed in the live data: the lb cache key is the full URL, and these pages fan out across every `symbol × entrypoint × version` (and for diff, every `old→new` version pair). So each *distinct* URL is a cold docs render at the origin that caching fundamentally can't collapse — and as the origin gets healthier, crawlers complete more of the walk, pushing load onto this uncacheable tail. Cutting the crawl is the lever.

This PR attacks it from two complementary angles: stop *inviting* the crawl (noindex), and *cap* it when it happens anyway (per-IP rate limit).

### 1. Unconditional `noindex` on high-cardinality, zero-SEO-value pages

- **Per-symbol doc pages** (`.../doc/.../~/:symbol`) — previously `noindex` only for pinned `@version` URLs, so the **latest** symbol pages (exactly what crawlers walk) stayed indexable.
- **All diff pages** — they key on `oldVersion`/`newVersion`, never `:version`, so the existing `ctx.params.version` guard was **dead code** and *every* diff page was indexable (N² version-pairs × symbols, no SEO value).

Canonical, low-cardinality pages stay indexable: the package index (`/@scope/pkg`), the docs landing/per-entrypoint pages, and `all_symbols`.

### 2. Stricter per-IP rate limit on docs/diff/source pages

A new `DOCS_RATELIMIT` (60 req / 60s) at the lb, applied **only** to the doc, diff, and source package pages via `isDocsDiffSourceRoute` in `lb/main.ts`. Keyed on `CF-Connecting-IP` — the real client IP at the Cloudflare edge — and stacked on top of the existing `FRONTEND_RATELIMIT` (240/60s), which stays as the broad frontend backstop.

This works where rate limiting was previously thought infeasible: the *API server* only sees the frontend's identity on SSR-driven renders, but the **edge lb sees the real client IP on every route**, so the gate sits in front of the SSR render that drives the origin load.

## Notes

- `noindex` reduces crawl *over time* (Google stops re-crawling and following into noindexed URLs); it's not instant, since a crawler must fetch a page once to see the header. The rate limit caps load immediately, including for clients that ignore `robots`/`noindex`.
- The limit is per-IP, so it fully bites a single/few-IP scraper; a widely distributed crawl needs the noindex lever (and the render-cost work) to do the heavy lifting.
- Scoped to the **pages** only. The backing `/api/.../docs` and `/api/.../diff` endpoints remain directly reachable without this limit (`isAPIRoute` routes them to `handleAPIRequest`, which has no guard) — capping the pages caps the SSR-driven renders, but a direct-API crawl is a separate vector left for follow-up.

### DIFF
--- a/frontend/routes/_middleware.ts
+++ b/frontend/routes/_middleware.ts
@@ -36,12 +36,17 @@ const auth = define.middleware(async (ctx) => {
     !pathname.startsWith("/api") &&
     !ctx.url.searchParams.has("__frsh_c");
   const { token, sudo } = getCookies(ctx.req.headers);
+  // Forward the originating client's user-agent so the api server records it
+  // on its HTTP span instead of an empty `http.user_agent`. Identify ourselves
+  // when the inbound request carries none.
+  const userAgent = ctx.req.headers.get("user-agent") ?? "jsr-frontend";
   if (interactive) {
     ctx.state.sudo = sudo === "1";
     ctx.state.api = new API(API_ROOT, {
       token,
       sudo: ctx.state.sudo,
       span: ctx.state.span,
+      userAgent,
     });
     if (ctx.state.api.hasToken()) {
       ctx.state.userPromise = (async () => {
@@ -53,6 +58,7 @@ const auth = define.middleware(async (ctx) => {
           ctx.state.api = new API(API_ROOT, {
             span: ctx.state.span,
             token: null,
+            userAgent,
           });
           const redirectTarget = `${ctx.url.pathname}${ctx.url.search}`;
           const loginUrl = `/login?redirect=${
--- a/frontend/routes/package/diff/[file].tsx
+++ b/frontend/routes/package/diff/[file].tsx
@@ -119,7 +119,10 @@ export const handler = define.handlers({
         versions,
         docsReq,
       },
-      headers: { ...(ctx.params.version ? { "X-Robots-Tag": "noindex" } : {}) },
+      // Diff pages key on old/new version (never `version`), so this branch was
+      // dead and every diff page was indexable — N² version pairs × entrypoints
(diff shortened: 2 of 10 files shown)


crowlbot (approved):
Solid, well-reasoned fix for a real problem. The root-cause analysis (crawlers walking the full `symbol × version` cardinality causing cold-render load that caching can't collapse) is convincing, and both levers — stop inviting the crawl, cap it when it happens anyway — are the right tools.

**noindex changes**

The dead-code catch on diff pages is a good find: `ctx.params.version` is never set on diff routes, so every diff page was indexable. Fixing all three diff variants unconditionally is correct.

Per-symbol doc pages (`.../doc/~/Symbol`) being unconditionally noindexed is the right call. Canonical low-cardinality pages (package index, docs landing, entrypoints, all_symbols) staying indexable preserves the SEO value that actually matters.

**Rate limit**

30 req/60s for `DOCS_RATELIMIT` stacked on top of the 240/60s `FRONTEND_RATELIMIT` is a sensible knob. The regex pattern covers all three route shapes correctly and the tests are thorough — the negative cases (score, versions, dependencies, API paths, /packages) are especially important to have explicit coverage for.

One minor observation: the PR bundles in the `userAgent` forwarding change (`_middleware.ts` / `api.ts`). That change is correct on its own (forwarding the originating UA so the api server records it on SSR spans), but it's unrelated to crawler load. Not a blocker, just worth noting for reviewers scanning for scope.

All CI checks pass. Good to go.

igorbdl (approved):

Issue #1448, opened 2026-06-10
Publish hangs indefinitely on GHA due to stale transaction locks after cancelled runs


uriva:
We are experiencing an issue publishing our package [`@alice-and-bot/core`](https://jsr.io/@alice-and-bot/core) (repository: `uriva/alice-and-bot`) to JSR from GitHub Actions.

### The Problem:
1. When publishing with provenance enabled, the publish process frequently hangs indefinitely during the Sigstore provenance signing phase (likely waiting on Rekor/Fulcio public transparency log rate-limits/throttling on shared GHA IPs).
2. If the user cancels the hanging GitHub Actions run, the runner is terminated, but **JSR's backend does not release the package transaction lock immediately**.
3. Because JSR keeps the package locked on their servers, **any subsequent publish attempts for new versions hang indefinitely on JSR's side (waiting for the lock)**, even if we attempt to publish with `--no-provenance` or change the publish command.
4. The `deno publish --dry-run` completes successfully and instantly in **104ms**, proving there are no compiler, resolver, or circular dependency loops in the codebase itself.
5. In addition, when runs are cancelled midway during the hang, JSR's backend appears to have registered the version in the database (creating the npm-compat mirror version under `@jsr/alice-and-bot__core`) but **fails to finalize the JSR native metadata (`meta.json` is not updated)**, resulting in new versions (e.g. `0.0.323`) existing on JSR's npm mirror but being completely invisible on JSR's native registry.

### Expected Behavior:
- JSR's backend should detect client disconnects/cancellations and immediately release the transaction lock on the package.
- Sigstore/OIDC signing operations should have a reasonable client-side and server-side timeout to fail early instead of hanging forever.

SisyphusZheng:
agree!!!

uriva:
my ai says: JSR CDN Cache Lag : JSR's Cloudflare CDN was caching meta.json with a 24-hour age, preventing Deno from finding the newly published @alice-and-bot/core@0.0.334. We bypassed this by mapping @alice-and-bot/core and ai-utils directly to their direct HTTPS endpoints in deno.json.

uriva:
also possibly related 

<img width="1266" height="302" alt="Image" src="https://github.com/user-attachments/assets/ada70d1c-3021-4bf9-b556-b2c23e55632f" />

in jsr you can't jump to latest version

uriva:
[...]

PR #1449, opened 2026-06-10
fix: self-heal publishing tasks stranded in processing/processed

crowlKats:
## Problem

Several recent reports (`@sys/tools@0.0.450`, `@alice-and-bot/core@0.0.323` in #1448, and "jump to latest" looping on its own version) are the same failure: a **publishing task stranded in a non-terminal state, leaving the package-level `meta.json` — the index Deno's resolver reads — never regenerated.**

The publish state machine has two phases:
- `Pending → Processed`: writes the immutable per-version `_meta.json` and atomically commits the version row + npm tarball.
- `Processed → Success`: regenerates `meta.json` from the DB, uploads it, and purges the CDN.

If the queue worker dies between them (Cloud Run timeout, cancelled CI run during Sigstore provenance signing, transient S3/Cloudflare error), the task strands:
- **`processed`** → version row + npm mirror exist, the version page renders, `_meta.json` returns 200 — but `meta.json` still lacks the version, so the resolver fails with `Could not find version ... that matches '<v>'`.
- **`processing`** → nothing recovers it (`publish_task` just errors `already processing`), and the `status != 'failure'` guard in `create_publishing_task` blocks re-publishing that exact version. This is the "stale transaction lock" described in #1448.

A secondary contributor: the manifest `Cache-Control` used `s-maxage=86400`, so a *failed* best-effort CDN purge could keep a correctly-regenerated `meta.json` invisible at the edge for up to 24h.

## Fix

1. **Reaper** — new `POST /tasks/requeue_stuck_publishing_tasks`, scheduled every 5 min via Cloud Scheduler. It finds tasks stuck in `processing`/`processed` for >30 min, resets stale `processing` tasks to `pending` (safe — their version row never committed, the finalize txn is atomic), and re-buffers everything through the publish queue, which drives `publish_task`'s state machine to completion (regenerating `meta.json`). This is the automated counterpart to the existing manual admin requeue endpoint.
2. **Cache-control** — `s-maxage` on `CACHE_CONTROL_MANIFEST` lowered `86400 → 60`, so a failed purge self-heals in ~60s via `stale-while-revalidate` with no added resolver latency.

## Notes

- Existing stranded packages still need a one-off manual `POST /api/admin/publishing_tasks/:id/requeue` until this deploys + schedules.
- The client/server-side asks in #1448 (timeouts on Sigstore signing, releasing the lock on client disconnect) are real but separate from this server-side stranding fix.

## Test plan

- New DB test `list_stale_publishing_tasks` covering the status filter and time threshold.
- `cargo clippy --all-targets`, `cargo fmt --check`, and `SQLX_OFFLINE=true cargo check --all-targets` all clean; sqlx offline cache regenerated.

### DIFF
--- a/api/.sqlx/query-c3c5af9e0e7091e64cb202578b02fa87bcc6e38b29644b69e79afad0e91cae98.json
+++ b/api/.sqlx/query-c3c5af9e0e7091e64cb202578b02fa87bcc6e38b29644b69e79afad0e91cae98.json
@@ -1,0 +1,41 @@
+{
+  "db_name": "PostgreSQL",
+  "query": "SELECT id, status as \"status: PublishingTaskStatus\"\n      FROM publishing_tasks\n      WHERE status IN ('processing', 'processed')\n        AND updated_at < now() - ($1::bigint * interval '1 second')\n      ORDER BY updated_at ASC\n      LIMIT 1000",
+  "describe": {
+    "columns": [
+      {
+        "ordinal": 0,
+        "name": "id",
+        "type_info": "Uuid"
+      },
+      {
+        "ordinal": 1,
+        "name": "status: PublishingTaskStatus",
+        "type_info": {
+          "Custom": {
+            "name": "task_status",
+            "kind": {
+              "Enum": [
+                "pending",
+                "processing",
+                "processed",
+                "success",
+                "failure"
+              ]
+            }
+          }
+        }
+      }
+    ],
+    "parameters": {
+      "Left": [
+        "Int8"
+      ]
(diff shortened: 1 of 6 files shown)


bartlomieju (approved):

mandarini:
This PR did not fully fix [this issue](https://github.com/jsr-io/jsr/issues/1448). :( I will try to put in a suggested patch.

mandarini:
Gave it a shot here: https://github.com/jsr-io/jsr/pull/1452 incomplete though

PR #1451, opened 2026-06-11
fix(search): fall back when package index misses results

SisyphusZheng:
## Summary

Improves package search resilience when the external Orama index drifts from registry state.

Changes:

- Falls back to the registry API/DB package search when Orama search throws.
- Falls back to the registry API/DB package search when a non-empty search term returns zero Orama hits.
- Fixes the package reindex filter to exclude packages with no versions, archived packages, and internal packages using `&&` instead of `||`.

## Why

Scope pages and package search use different data sources. A scope page can show packages from registry/API state while `/packages?search=...` returns no results if the Orama package index is stale or missing documents. Falling back for non-empty searches keeps package discovery useful even when the external index is temporarily inconsistent.

Concrete reproduction observed in production:

```text
https://jsr.io/@openelement
  -> lists packages

https://jsr.io/packages?search=%40openelement
https://jsr.io/packages?search=openelement
  -> no results from package search
```

## Verification

- `deno fmt --check frontend/routes/packages.tsx tools/orama_packages_reindex.ts` passed.
- `cd frontend && deno check --allow-import=googleapis.deno.dev,deno.land,jsr.io main.ts routes/**/*.tsx routes/**/*.ts` passed.
- `deno check --allow-import=googleapis.deno.dev,deno.land,jsr.io tools/orama_packages_reindex.ts` passed.

Notes:

- `deno task lint:frontend` currently fails on existing `no-process-global` findings outside and including this route's pre-existing `process.env` usage.
- `deno task lint:tools` currently fails on existing unrelated issues in `tools/db-switch.ts` and `tools/dev.ts`.

### DIFF
--- a/frontend/routes/packages.tsx
+++ b/frontend/routes/packages.tsx
@@ -1,5 +1,5 @@
 // Copyright 2024 the JSR authors. All rights reserved. MIT license.
-import { define } from "../util.ts";
+import { define, type State } from "../util.ts";
 import { OramaCloud } from "@orama/core";
 import type { List, Package } from "../utils/api_types.ts";
 import { assertOk, path } from "../utils/api.ts";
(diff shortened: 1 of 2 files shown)


crowlKats:
Closing as we have switched away from Orama

PR #1453, opened 2026-06-11
fix: hung publishes & frozen package metadata from lb cache key/poll caching

igorbdl:
This PR together with #1449 fixes the hang and stale/invisible metadata symptoms described in #1448


crowlKats (approved):

philcockfield:
Thanks for doing this @igorbdl 🙇🏻‍♀️👏

PR #1460, opened 2026-06-15
feat: sample otel traces and logs at 5%

crowlKats:
## What

Samples OpenTelemetry traces and their logs at 5% to cut OTLP export volume/cost. Affects **export only** — stdout logs are unchanged.

### Trace sampling
Adds a `ParentBased(TraceIdRatioBased(0.05))` sampler to the `TracerProvider`. Parent-based so child spans inherit the root's decision — each sampled trace stays whole rather than losing spans mid-trace, and an upstream sampling decision propagated via tracecontext is honored.

### Log sampling
The bridged logs (`/v1/logs`) weren't trace-sampled. A `LogSampler` filter is attached to the `OpenTelemetryTracingBridge` layer:

- **WARN/ERROR logs are always exported**, regardless of sampling — these are the lines you want during an incident, and head sampling can't keep error traces (the decision is made before the request runs). Keeping the log even when its trace was dropped means the message, fields, and trace_id survive even if the spans don't.
- Other logs within a trace are kept iff that trace was sampled in — the decision mirrors the SDK's `TraceIdRatioBased` algorithm and is a pure function of the trace id, so every log in a trace agrees with the span sampler's choice. A kept trace keeps all its logs; dropped traces export zero of their INFO/below logs (no orphaned/empty traces).
- Trace-less logs (e.g. startup) fall back to a random 5% draw.

Both rates share one `SAMPLE_RATIO = 0.05` constant. Level filtering (`EnvFilter`) still applies on top.

## Known limitation / follow-up: error traces are not guaranteed
This is **head-based** sampling — the keep/drop decision is made at span creation from the trace id, before the request runs, so it cannot preferentially keep traces that go on to error (~95% of error *traces* are still dropped; their WARN/ERROR *logs* are kept, see above). Guaranteeing error traces requires **tail sampling** in an OTel Collector (`tailsamplingprocessor`: a `status_code: ERROR` policy that always keeps + a `probabilistic` 5% policy for the rest). We currently export directly to the Grafana Cloud gateway with no collector in the path, so that's an infra change, not reachable from app code, and out of scope for this PR.

## Test plan
- `cargo check`, `cargo clippy`, `cargo fmt --check` clean
- All `tracing::` tests pass, including a new statistical test confirming the trace-id sampler keeps ~5% (within 1%) and is deterministic per id

### DIFF
--- a/api/src/tracing.rs
+++ b/api/src/tracing.rs
@@ -259,6 +259,14 @@ where
     event: &tracing::Event<'_>,
     cx: &Context<'_, S>,
   ) -> bool {
+    // Always export problem logs regardless of sampling: these are the lines
+    // you want during an incident, and head sampling cannot keep error traces
+    // (the decision is made before the request runs). Keeping the log even when
+    // its trace was dropped means the message, fields, and trace_id survive
+    // even if the spans don't.
+    if event.metadata().level() <= &tracing::Level::WARN {
+      return true;
+    }
     match event_trace_id(event, cx) {
       // In a trace: keep the log iff the trace's `TraceIdRatioBased` decision
       // sampled it in. The decision is a pure function of the trace id, so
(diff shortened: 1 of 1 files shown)


KnorpelSenf (commented):
Does this ensure that traces of erroring requests are always kept? Is that worth considering?

igorbdl (approved):

PR #1465, opened 2026-06-19
fix(provenance): verify attested subject digest against published tarball

crowlKats:
Provenance verification checked the attestation's subject *name* but ignored its `subject.digest.sha256`, so an attestation was only bound to the package `name@version` and not to the bytes actually stored.

This records the SHA-256 of the uploaded tarball on the publishing task and compares the attestation's subject digest against it during `POST .../provenance`. Verification now fails closed when no recorded hash is available.

## Changes
- Migration: add nullable `tarball_hash` to `publishing_tasks`.
- Persist the tarball hash (already computed at upload) on the publishing task.
- Provenance handler looks up the recorded hash and passes it to `provenance::verify`, which rejects on digest mismatch.

## Tests
- New unit test for the digest comparison helper (incl. mismatch/empty cases).
- Updated `test_package_provenance` to record a hash so the existing DSSE-signature regression case still exercises signature verification.

cargo check / clippy / provenance unit + integration tests all pass locally.


crowlbot (commented):
Good security fix — binding the attestation's `subject.digest.sha256` to the actual uploaded bytes rather than trusting only the `name@version` string closes a real gap. The migration, DB helpers, `digest_matches` implementation (prefix stripping + case-insensitive compare), and test coverage all look correct.

**CI is failing: `cargo fmt`**

Two long hex string literals in the new test need line-wrapping to match rustfmt's 100-char limit:

```rust
// line 399 — needs wrapping to:
let hex =
  "1c3b44ea2ac86f7133791a4a004f633993784da783a3e0f5c226dd7a4141f9f5";

// line 410 — needs wrapping to:
let other =
  "0000000000000000000000000000000000000000000000000000000000000000";
```

Running `cargo fmt` in `api/` will fix this automatically.

**One design question: 500 for pre-migration versions**

When a version was published before this migration (no `tarball_hash` row), the handler returns `ApiError::InternalServerError`. This means provenance registration for those existing versions will start returning 500 rather than working as before. Is that intentional? If it is (deliberate fail-closed), a 422/412 with an explicit message might be less confusing to callers than a 500, and the server log already captures the `error!("no recorded tarball hash...")` context. If backwards compat matters, skipping the digest check when no hash is on record (with a warning log) is an alternative — though that does weaken the guarantee for those versions.

crowlbot (approved):
CI green, formatting fixed. The security fix is correct — the attestation is now properly bound to the actual published bytes. Approach is sound: fail closed for pre-migration versions is the right call given this is a security check. Approved.

PR #1470, opened 2026-07-02
Experiment: run the registry API as a wasm32-emscripten Cloudflare Worker

crowlKats:
**Draft / experimental — not for merge.** Sharing the approach for discussion.

Runs the *entire* `registry_api` as a Cloudflare Worker compiled to
`wasm32-unknown-emscripten` via the [workers-rs](https://github.com/cloudflare/workers-rs)
emscripten pipeline (which, unlike `wasm32-unknown-unknown`, supports tokio,
hyper, real sockets, DNS, and rustls+ring TLS).

**Status:** builds, links, and **runs in workerd** (`wrangler dev`) — the module
instantiates, the `fetch` export runs, config is read from the JS `env`, the
router builds, and tokio async networking is live (verified with a real
emscripten async DNS lookup). With the dummy smoke-test config it stops at
`Database::connect` (`localhost` unresolvable), i.e. everything up to the router
works.

### What's here (jsr-side source + config only)
- `api/src/main.rs`: native keeps `#[tokio::main]` + `Server::bind().serve()`;
  wasm adds a `#[wasm_bindgen(tokio)] fetch` export bridging
  `web_sys::Request ↔ hyper ↔ routerify`. Shared setup in `build_router()`.
- `async-tar` → **`astral-tokio-tar`** (tokio-based; async-std won't build on wasm).
- **jsonwebtoken 8→10 / jsonwebkey 0.3→0.4 / x509-parser 0.15→0.16** — moves to
  pure-Rust crypto and drops ring 0.16.20 (whose Montgomery multiply is asm-only,
  the one true link blocker). No jsr source changes needed for that migration.
- `cfg`-gate jemalloc / S3-rustls off wasm; `askalono` `gzip` (zstd C lib won't
  build on wasm); wasm-target deps; `.cargo/config.toml`; smoke-test `wrangler.toml`.

### Not committed (documented in `EMSCRIPTEN_EXPERIMENT.md`)
The `[patch.crates-io]` recipe points at machine-absolute checkouts (workers-rs
tokio fork + vendored wasm-bindgen) and a 24 MB `.emscripten-patches/` dir, so
**this branch is not CI/other-machine buildable by design**. Native build is
unaffected. See the doc for the full reproduction recipe and the list of ~12
dependency/toolchain blockers cleared.


crowlbot (commented):
Neat writeup — the wasm-vs-unix cfg gating theme across `ring`/`socket2`/`tar`/`deno_doc` and the pure-Rust crypto migration to clear the `ring` asm-only Montgomery-multiply wall are both good finds, and `EMSCRIPTEN_EXPERIMENT.md` makes the (very involved) repro path legible even though the patch tree itself isn't committed. Appreciate you sharing this for discussion rather than just letting it rot on a branch.

One design point worth flagging for if/when this moves past smoke-testing: both `to_hyper_request` and `to_web_response` fully buffer the body in memory (`req.array_buffer()` and `hyper::body::to_bytes`) rather than streaming. That's fine for typical API calls, but jsr's tarball upload/download paths (`api/src/tarball.rs`) can be tens of MB, and combined with the 8MB emscripten stack + `ALLOW_MEMORY_GROWTH` here, publish/download traffic through this path could get memory-pressure-y under concurrent requests in a single worker instance. Might be worth noting as a known gap alongside the 35MB binary size limit, or scoping this experiment to read-mostly/metadata routes if a follow-up gets serious.

Also: `env_into_process` sets process env from untrusted-ish `env` JS values via `unsafe { std::env::set_var }` — safe here since it's only ever called once through the `OnceCell::get_or_try_init` guard before any concurrent request handling starts, but worth a one-line comment noting *why* it's safe (single-caller-via-OnceCell, not just "single-threaded") so a future reader doesn't assume it's safe to call `env_into_process` again later, e.g. on worker env rotation.

No concerns with the native path — the `#[cfg(not(target_arch = "wasm32"))]` gating looks correct and `build_router` extraction is a clean shared entry point.

crowlbot (commented):
The runtime-reliability fixes make sense and match the failure modes you'd expect from this environment:

- Per-request router/pool rebuild to work around the "different request" I/O isolation is the right call given the constraint, and the doc is upfront that this makes concurrent requests contend (already flagged as a known gap) rather than hiding it.
- The lazy single-connection pool (`max_connections(1)`, no idle reaper) matches the goose reference pattern and the reasoning for why a parked/reused pooled connection hangs across the persistent reactor is convincing.
- DNS prewarm via `tokio::net::lookup_host` before the blocking `getaddrinfo` calls is a sensible fix, and `host_of` handles the userinfo/port stripping correctly for the cases this needs (DB/S3 URLs).
- Good catch tying `DATABASE_DISABLE_MIGRATIONS=1` to the eager-connection-at-router-build concern in the docs — that answers the question I would've asked about migrations now running on every request build.

One thing I traced through but didn't find an actual bug in: `env_into_process` now runs on every `fetch()` call instead of once behind the old `OnceCell` guard, right before `Config::try_parse()`. Since neither has an `.await` between them, there's no yield point for another request's task to interleave its own `set_var` calls in between — so it's sound under the single-threaded cooperative scheduler as-is. Given the `unsafe` block's safety comment ("single-threaded emscripten worker; no other threads race here") is now doing double duty — no OS threads *and* no interleaving within this synchronous span — it'd still be worth tightening that comment so a future change that adds an `.await` between the two calls doesn't silently reintroduce a real race.

No blockers from me for continued smoke-testing on this branch.

PR #1473, opened 2026-07-06
feat: disable diff pages and serve docs only for the latest version

crowlKats:
## Summary

Disables the package **diff view** and restricts **documentation pages to the latest version** of a package; requests for any other version redirect to the latest.

### Diff view (disabled, code kept)
- The three frontend diff routes return `404` behind a `DIFF_ENABLED` flag, and the **Diff** nav tab + **Jump to diff** button are removed.
- The API `get_diff_handler` returns a new `DiffDisabled` error behind the same flag.
- All diff rendering code (`DiffView`, `DiffVersionSelector`, `packageDataWithDiff`, the diff computation) is left intact — flip the flag(s) to re-enable.

### Docs latest-only
- `get_docs_handler`, `get_docs_search_handler`, and `get_docs_search_structured_handler` only serve a pinned `:version` if it is the current docs-latest version; anything else returns the new `DocsOnlyForLatestVersion` error. The `latest` alias still works.
- `packageDataWithDocs` turns `docsOnlyForLatestVersion` into a `302` to the canonical **versionless** docs URL (preserving entrypoint/symbol/all_symbols), so non-latest doc pages redirect to the latest.
- `source` (Files tab) is unaffected — all versions remain browsable there.

### Prerelease handling
Docs-latest resolves to the **latest unyanked stable version**, falling back to the **latest prerelease** for packages that only have prereleases (via a new `get_latest_unyanked_version_for_package_for_docs` DB method ordering stable ahead of prerelease). So prerelease-only packages keep working docs instead of 404ing.

## Behavior

| Package state | `…/doc` (versionless) | pinned latest | pinned older |
|---|---|---|---|
| Has stable release | serves stable latest | serves | → redirect to latest |
| Prerelease-only | serves latest prerelease | serves | → redirect to latest |
| Diff (any) | 404 | 404 | 404 |

## Tests
- `test_package_docs`: adds coverage for the `latest` alias, non-latest → `docsOnlyForLatestVersion` (docs + both search endpoints), and diff → `diffDisabled`.
- New `test_package_docs_prerelease_only`: prerelease-only package serves docs at `latest`/pinned, older version still rejected.

## Notes
- Includes a regenerated sqlx offline cache entry for the new query and a new `ok_prerelease` test tarball.
- `cargo clippy --all-targets -- -D warnings`, `cargo fmt --check`, `deno lint`, and both docs tests pass locally.

### DIFF
--- a/api/.sqlx/query-3019429dc85e4e9cf860c11760074cea191570f860c9b7da1caf4d1e7c001384.json
+++ b/api/.sqlx/query-3019429dc85e4e9cf860c11760074cea191570f860c9b7da1caf4d1e7c001384.json
@@ -1,0 +1,95 @@
+{
+  "db_name": "PostgreSQL",
+  "query": "SELECT scope as \"scope: ScopeName\", name as \"name: PackageName\", version as \"version: Version\", user_id, readme_path as \"readme_path: PackagePath\", exports as \"exports: ExportsMap\", is_yanked, uses_npm, meta as \"meta: PackageVersionMeta\", updated_at, created_at, rekor_log_id, license\n      FROM package_versions\n      WHERE scope = $1 AND name = $2 AND is_yanked = false\n      ORDER BY (version NOT LIKE '%-%') DESC, version DESC\n      LIMIT 1",
+  "describe": {
+    "columns": [
+      {
+        "ordinal": 0,
+        "name": "scope: ScopeName",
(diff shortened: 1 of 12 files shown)


crowlKats:
This is a temporary change until some solution is found regarding bot traffic.

nathanwhit (approved):

PR #1476, opened 2026-07-09
feat: add Cloudflare Turnstile captcha to login

crowlKats:
Gates the login flow behind a Cloudflare Turnstile captcha, verified server-side in the Rust API.

## `/login/:service` becomes a POST

This is the load-bearing change. As a GET link, anyone could navigate straight to `/login/github` and skip the widget entirely. As a form POST it carries the Turnstile token in the body — out of URLs, logs and `Referer` headers — and a bare link can no longer start a login flow. `login_handler` verifies the token before it touches the database, since the route is unauthenticated and inserts an `oauth_state` row on every call.

Two consequences worth reviewing closely:

- **307 → 303 on the redirect to the identity provider.** A 307 preserves the method, so the browser would have POSTed the login form to GitHub's authorize endpoint. There's a test pinning the 303.
- **A bad secret fails closed, not confusingly.** Probing the live endpoint: a rejected *token* returns HTTP 200 with `success:false`, but a wrong *secret* returns HTTP 400. They map to different errors — a bad token is the user's problem (400, "try again"), while a secret siteverify can't evaluate or a siteverify outage is ours (503, retryable, loudly logged) rather than telling users their captcha was invalid.

## Three bypass doors closed

Besides the login page, the header dropdown (`SignInMenu.tsx`) and the publish page (`new.tsx`) both linked directly to `/login/github`. The dropdown collapses to a single "Sign in" link to `/login`; `SignInMenu.tsx` is deleted, and `new.tsx` points at `/login` too.

## Config

Terraform provisions the widget itself (`cloudflare_turnstile_widget`), which exposes `sitekey` and `secret` as read-only attributes. So there is nothing to set by hand and the two keys cannot drift apart:

- `sitekey` → plain-text Worker binding `TURNSTILE_SITE_KEY` (public, like the Algolia search keys)
- `secret` → Secret Manager → `TURNSTILE_SECRET_KEY` on both Cloud Run services, plus the IAM accessor binding

The widget is scoped to `var.domain_name` (the frontend origin where it's embedded), and uses `mode = "managed"`: Cloudflare clears most visitors silently and asks the rest to tick a box.

The never-prompt modes (`non-interactive`, `invisible`) were considered and rejected. A bot faces the same passive bar in every mode, so they give up nothing bot-side — but they leave a visitor the passive signals cannot clear (Tor, a VPN, a hardened browser) with no way to prove themselves, and therefore no way to sign in at all. `invisible` additionally requires referencing Cloudflare's Turnstile Privacy Addendum from a privacy policy, and JSR does not publish one.

One deploy-time prerequisite: **the `cloudflare_api_token` the provider runs as needs Turnstile write permission**, or the apply fails on widget creation.

With no keys configured the captcha is disabled end to end: the frontend renders no widget and the API skips verification, so local dev and tests are unchanged.

## Two things to know

- **Signing in now requires JavaScript.** The buttons were plain links; they're now disabled until the widget resolves. Turnstile needs JS regardless, so this is inherent to the feature — but it is a real behavior change.
- **Terraform is unvalidated** — no `terraform` CLI was available in the environment, so the resource schema was taken from the provider docs and the style matched by hand. Worth a careful look, and it's the main reason for the `test-on-staging` label.

## Testing

- `cargo clippy` clean; 12 tests pass, including two new ones asserting a GET cannot start a flow and a POST returns 303 to the provider.
- The island was rendered server-side to confirm `formAction` serializes to a real `formaction` attribute and the buttons ship `disabled` with the hidden token input.
- Cloudflare's real `siteverify` was probed with their documented test keys to pin down the request/response contract — that's what surfaced the 400-on-bad-secret behavior above.

Not verified: the full browser flow against a real widget, and `terraform validate`.


piscisaureus (approved):
Rubberstamp

PR #1477, opened 2026-07-14
fix(frontend): restrict frame-src to block javascript: iframe CSP bypass

crowlKats:
## Summary

The interactive CSP set in `frontend/routes/_middleware.ts` relied on `script-src-attr 'none'` to neutralize package-controlled markup injected into docs pages (it blocks inline event-handler attributes like `onerror=`). But with no `script-src` / `default-src` / `frame-src` directive, an injected `<iframe src="javascript:...">` navigates rather than using an event-handler attribute, so it was not caught and executed same-origin.

This adds a `frame-src` directive restricting frames to the Cloudflare Turnstile host (the only legitimate iframe in the app — the login captcha), which blocks `javascript:`/`data:` frames while keeping login working.

## Notes

- Turnstile (`https://challenges.cloudflare.com`) is the only legitimate iframe; `Markdown.tsx` already sets `allowIframes: false`.
- Reported privately by Huli (@aszx87410); tracked in a repository security advisory.
- A full nonce-based `script-src` is a larger follow-up.

### DIFF
--- a/frontend/routes/_middleware.ts
+++ b/frontend/routes/_middleware.ts
@@ -109,11 +109,20 @@ const cache = define.middleware(async (ctx) => {
 // JS via addEventListener, never as inline HTML attributes, and legitimate
 // inline <script> elements (dark-mode bootstrap, island hydration) are governed
 // by script-src, not script-src-attr, so they keep working.
+//
+// `frame-src` closes the `script-src-attr` bypass: an injected
+// `<iframe src="javascript:...">` navigates rather than using an event-handler
+// attribute, so it is not caught by script-src-attr and — absent a script-src /
+// default-src / frame-src directive — executes same-origin. Restricting frame
+// sources to Cloudflare Turnstile (the only legitimate iframe, on the login
+// page) blocks `javascript:` and `data:` frames while keeping the captcha
+// working.
 const CSP = [
   "script-src-attr 'none'",
   "object-src 'none'",
   "base-uri 'self'",
   "frame-ancestors 'self'",
+  "frame-src https://challenges.cloudflare.com",
 ].join("; ");
 
 const securityHeaders = define.middleware(async (ctx) => {
--- a/frontend/routes/_middleware.ts
+++ b/frontend/routes/_middleware.ts
@@ -102,34 +102,65 @@ const cache = define.middleware(async (ctx) => {
 //
 // Package documentation embeds HTML rendered from package-controlled symbol
 // names. deno_doc does not escape every name, so a malicious package can inject
-// markup such as `<img src=x onerror=...>` that is rendered same-origin on
-// jsr.io. `script-src-attr 'none'` blocks all inline event-handler attributes
-// (onerror/onclick/...), which is the execution vector for that payload. This
-// is the right tool for a Preact/Fresh app: event handlers are attached from
-// JS via addEventListener, never as inline HTML attributes, and legitimate
-// inline <script> elements (dark-mode bootstrap, island hydration) are governed
-// by script-src, not script-src-attr, so they keep working.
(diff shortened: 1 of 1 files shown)


crowlbot (approved):
Confirmed no other iframe usage in the codebase besides Turnstile (`allowIframes: false` in `Markdown.tsx`, and grep for `iframe`/`turnstile` turns up only the login captcha), so restricting `frame-src` to `challenges.cloudflare.com` doesn't break anything else while closing the `javascript:`/`data:` iframe navigation bypass. CI green. LGTM.

aszx87410:
Hi, unfortunately `frame-src` cannot block `<iframe src="javascript:"></iframe>`. What you need is `script-src`.

crowlbot (changes requested):
@aszx87410 is right, and I should have caught this before approving — retracting my LGTM.

Checked this against the spec after your comment: this exact ambiguity is tracked upstream at [w3c/webappsec-csp#127](https://github.com/w3c/webappsec-csp/issues/127) ("Interaction of CSP and javascript: URLs in iframe src is not defined anywhere"). `javascript:` URL execution in a nested browsing context isn't a fetch, so it isn't gated by `frame-src` (which restricts *sources loaded into* the frame) — it's inline script execution in a context that, per spec, inherits the initiator's policy and is gated by `script-src`/`default-src`. Since this CSP has neither, `frame-src https://challenges.cloudflare.com` doesn't actually close the `<iframe src="javascript:...">` vector; it only stops the iframe from being pointed at a *different* origin.

This PR probably shouldn't merge as-is — it would close the security advisory as fixed while the underlying bypass most likely still works. Given inline `<script>` (dark-mode bootstrap, island hydration) is already relied on elsewhere, a full `script-src` lockdown needs nonces (already called out as a follow-up), but that follow-up may need to land together with this rather than after it. Requesting changes until this is verified against a real browser with a PoC, or the `script-src`/nonce work is included.

AugustinMauroy (approved):

crowlbot (approved):
Verified the nonce mechanism against `fresh` core (`freshframework/fresh`, `packages/fresh/src/{context,runtime/server/preact_hooks}.ts`) rather than taking the comment at face value:

- `RenderState.nonce = crypto.randomUUID().replace(/-/g, "")` — generated fresh per render from a CSPRNG, 122 bits of entropy. Not a counter or anything guessable.
- The vnode hook only stamps `nonce` onto `<script>`/`<style>` elements that don't already have one, and only during Preact rendering — so it can't land on markup injected via `dangerouslySetInnerHTML`, confirming the "package-controlled symbol names can't forge a nonce" claim.
- `(response as any)[NONCE_SYMBOL] = renderNonce` is set on **every** `ctx.render()` response unconditionally, independent of whether client JS is actually emitted — so the "every interactive HTML page carries a nonce, fail-closed otherwise" logic in this PR holds and shouldn't false-positive into `script-src 'none'` on a normal page.

That confirms this closes the gap from the earlier review: dropping `'self'` from `script-src` (given raw package files are served same-origin) and gating on a per-render nonce with `'strict-dynamic'` blocks `javascript:` URL execution regardless of entry point (inline injected `<script>`, event-handler-attribute fallback, or the `javascript:` iframe navigation aszx87410 flagged — since that's evaluated against `script-src` of the inheriting context, not `frame-src`).

Two non-blocking notes:
- The Cloudflare nonce-propagation caveat you left in the comment (challenge/bot-management scripts needing the header-propagated nonce) is real and worth a follow-up smoke test in production after deploy, as you noted.
- No test coverage for the CSP header itself (before or after this PR) — not a regression you're introducing, but a `_middleware.ts` test asserting `script-src` contains a nonce and no `'unsafe-inline'`/`'self'` would catch a future accidental widening.

Approving — this addresses the bypass correctly.

PR #1479, opened 2026-07-17
feat: move frontend scraping protection from the LB to the WAF

crowlKats:
Moves frontend scraping protection out of the LB Worker and onto the WAF, without ever touching the raw module serving path.

The in-Worker limits ran *after* both Workers had already been invoked, so a throttled scraper still cost an LB invocation plus a frontend invocation. WAF phases run before Workers, so the same verdict now costs nothing.

## Why this is shaped the way it is

Every WAF phase runs *before* Workers, so the WAF cannot see what `route()` in `lb/main.ts` decided — it has to re-derive "frontend vs. R2" from the request alone. That re-derivation (the `Accept`-header trick below) is the one awkward part of this PR, so it's worth recording the two alternatives that avoid it and why neither works here.

**Rejected: split modules onto their own hostname/zone** (the Cloudflare suggestion). If raw modules lived on `p.jsr.io`, the apex would be pure frontend and could take bot management wholesale with no path logic. That is genuinely the cleaner design — but it requires moving the module URL, and `jsr.io/@scope/pkg/1.0.0/mod.ts` is baked permanently into published source maps, stack traces, and import specifiers across the ecosystem. It can't move. The frontend is pinned to the apex too (package pages, docs, SEO), so both surfaces share one hostname *by design* — which is the very reason the LB has its classification logic. The degenerate form (LB classifies, then cross-zone `fetch()`es frontend traffic to a WAF-bearing zone) still contains the full classification logic *and* adds two failures: cross-zone subrequests overwrite `CF-Connecting-IP` with a fixed Worker address, so the WAF zone sees every user as one IP (per-IP limiting dies), and `cf_clearance` is per-zone, so a challenge issued to a proxied subrequest can never be solved by the browser. That zone could only ever `block`, never `challenge`.

**Rejected: attach the WAF to the frontend Worker instead of the LB.** Tempting, because the LB has already classified by the time traffic reaches the frontend Worker — it only ever sees genuine frontend traffic, so WAF there would need no path logic. But WAF/rate-limiting/bot-management aren't attachable to a Worker; they're a zone-ingress phase engine, and the frontend Worker isn't at an ingress — it has no route and is reached solely through the LB's `FRONTEND` service binding, which bypasses the network and the zone pipeline entirely. Giving it a hostname so a phase *could* run reintroduces the subrequest problems above: a same-zone forwarded `fetch()` bypasses Cloudflare security settings by default (the same gotcha already documented in `cloudflare_frontend.tf`, where the frontend routes API calls back through the LB binding for this reason), and either way the browser's real connection is to `jsr.io`, not the forwarded hop — so a challenge issued downstream is unsolvable and bot scoring would measure the LB Worker rather than the visitor. The only place with both a real WAF surface and the real client connection is the apex ingress, which is exactly where the LB route sits.

So re-deriving the classification at the apex is the price of getting managed-challenge, verified-bot data, and the real client IP at all. The decision is expressed once, in the one phase that can read request headers. (Enforcing *inside* the Worker instead — as the deleted `FRONTEND_RATELIMIT`/`DOCS_RATELIMIT` bindings did — reuses the classification perfectly but gives up all three, and costs a Worker invocation to reject.)

## What it does

**One custom rule** matches a raw module fetch — a module-shaped path whose `Accept` is not `text/html` — and skips three phases:

- `http_ratelimit`, so installs are never throttled.
- `http_request_sbfm`. This one matters most: SBFM's only knob on our plan is *"Definitely automated"*, and a CLI with a fixed JA3/JA4, no JS engine and no cookies scores 1 — exactly that bucket. Today, flipping that toggle in the dashboard would break every `deno add` instantly, with no path-level exemption. This makes the toggle survivable.
- `http_request_firewall_managed`, so a published file named `union_select.ts` can't trip OWASP and break its own package's installs. Managed Rules protect nothing on a static R2 GET served under a strict CSP.

**Two rate limiting rules** (exactly the plan cap) replace `FRONTEND_RATELIMIT` and `DOCS_RATELIMIT`, escalating to a managed challenge rather than a block — a false-positive scraper verdict against a real reader is then recoverable in one click instead of a dead 403. Challenges work because these rules live on the apex zone the browser is actually visiting.

That `skip` is non-terminating when it only names phases is what makes the rest work. It acts as a header-aware pre-filter, which matters because rate limiting rule *expressions* can't read request headers on our plan. Source pages (`/@scope/pkg/1.2.3/mod.ts`) are module-shaped and indistinguishable from a raw module fetch by path alone — but they don't need to be distinguished, because every real module fetch has already been skipped out of the phase. Anything module-shaped still arriving there can only be a source-view render, so a path-only rule is precise.

`isDocsDiffSourceRoute` and its tests go with the limits; the function existed only to choose between the two.

## The safety property

`waf_re_module_shaped` is deliberately **wider** than `isModuleFilePath()` rather than an exact mirror. Two copies of that regex will drift, and the direction of drift is what matters: a *narrower* copy rate-limits `deno add`, a *wider* one just leaves some frontend traffic unprotected. So it's a strict superset — verified across 1233 generated paths (all 546 R2-bound paths exempt, all doc/diff routes still protected, zero violations).

There's a warning on the local in the file. **Narrowing it breaks `deno add` globally.**

For the removal, I replayed the corpus from `lb/main_test.ts` against the new rules: all 15 doc, diff, and source paths the LB throttled are still throttled at the edge. The sole over-inclusion is `/@scope/pkg/meta.json` requested as HTML, which lands in the stricter bucket and renders a cheap 404.

## Worth a reviewer's eye

- **The `Accept` expression.** `any(starts_with(http.request.headers["accept"][*], "text/html"))` is composed from documented rules (`starts_with` is a function, not an operator — the `[*] starts_with "…"` form doesn't parse), but no CF doc shows this exact combination. Worth pasting into the dashboard expression builder once. The tempting `contains` variant is subtly wrong: it would un-exempt `Accept: application/json, text/html`, which the Worker still serves from R2.
- **The numbers are guesses** — 15/min on doc+diff+source, 300/min general — and should be tuned against real traffic. For reference, the limits they replace were 20/min and 240/min, though the old general limit also counted static assets, so it bound far tighter than 240 in practice.
- **Rollout.** `log` isn't available as a custom-rule action below Enterprise, so no log-only dry run. The skip rule has `logging.enabled = true` — watch Security Events to confirm it matches real `deno add` traffic. Since this PR also removes the in-Worker backstop, it's worth confirming the skip rule matches as intended before/soon after apply.
- **Entitlement.** Uses regex (`matches`), which is Business+ by plan and works via our granted WAF access. If it bounces, the error is literally `not entitled: … a Business plan or a WAF Advanced plan is required`.

No provider bump needed — this validates against the locked 5.19.1. (Latest is 5.22.0; there is no v6, v5 was the rewrite.)

## Follow-ups, deliberately not in this PR

- `lb/bots.ts` is a spoofable User-Agent allowlist that punches straight through to the frontend — `User-Agent: Slack` is a free bypass today. `cf.verified_bot_category` is ASN/reverse-DNS verified and could replace it.
- Bot-score rules (`cf.bot_management.score`) would catch scrapers staying under the rate limit, and `cf.verified_bot_category in {"AI Crawler"}` is a natural addition. Both are more aggressive than rate-limit-then-challenge, so better tuned separately.

---

### PR Checklist

- [x] The PR title follows [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/)
- [x] Is this closing an open issue? If so, link it, else include a proper description of the changes and reason behind them.
- [ ] Does the PR have changes to the frontend? — no.
- [ ] Does the PR have changes to the backend? — no. The LB Worker changes are deletions; the remaining `lb` tests pass.


crowlbot (approved):
Thorough writeup, and I traced the two regex-parity claims against current `main` to check they hold:

- `waf_re_module_shaped` (`^/@[^/]+/[^/]+/(meta\.json|[0-9])`) vs. `isModuleFilePath()` (`/^\/@[^/]+\/[^/]+\/(?:meta\.json|\d[^/]*_meta\.json|\d[^/]*\/.*)$/`): every alternative in the Worker's regex requires the third path segment to start with literal `meta.json` or a digit, which is exactly what the WAF regex checks (minus the trailing structure/anchor) — so it's a genuine strict superset, and the drift direction is the safe one (over-matching → merely unprotected, not blocked installs).
- `waf_re_api_route` vs. `isAPIRoute()`: the `startsWith` branches (`api/`, `login/`, `connect/`, `disconnect/`) are unanchored at the end in the WAF regex (matching prefix semantics), and the exact-match branches (`logout`, the three `sitemap*.xml`) correctly get a trailing `$`. This lines up field-for-field.

The `Accept` comparison also checks out: since Cloudflare's `http.request.headers["accept"][*]` yields the raw header value(s) rather than comma-splitting, `starts_with(..., "text/html")` on the whole string agrees with the Worker's `accept?.startsWith("text/html")` — so `"application/json, text/html"` is correctly treated as non-HTML on both sides, as called out in the PR body.

No code-level issues found. The one thing I can't verify from here is the live Cloudflare behavior called out in the PR itself (`mitigation_timeout = 0` requirement for `managed_challenge`, and whether the custom-rule expression actually parses in the dashboard) — CI doesn't run `terraform plan`/`validate` against the real zone on this PR, so that's on manual verification before/at apply time as you've already flagged. Given the corpus verification (1233 paths) and the explicit safety-direction argument, I'm comfortable approving on the logic; the rollout notes (watch Security Events, tune the guessed limits) are the right things to keep an eye on post-merge.

KnorpelSenf:
The promised Codex review:

## Findings

- [P1] The module skip rule can be bypassed into the frontend — `terraform/waf.tf:93`

  The WAF classifies module requests using only method, path, and `Accept`, while the Worker also checks `Sec-Fetch-Dest`/`Sec-Fetch-Site` and spoofable bot headers (`lb/main.ts:171`), `lb/main.ts:189`). For example:

  ```http
  GET /@scope/pkg/1.2.3/mod.ts
  Accept: */*
  Sec-Fetch-Dest: document
  ```

  This matches `skip_module_serving`, skipping rate limiting, SBFM, and managed WAF rules, but `canAccessModuleFile()` rejects it and sends it to the frontend. Spoofing one of `lb/bots.ts`’s bot headers has the same effect. Consequently, source-page scraping—and managed-rule inspection of those frontend requests—can be bypassed with one header. The skip expression must mirror the Worker’s complete routing decision, or routing must expose a signal the WAF can classify reliably.

No other actionable findings found.

crowlKats:
@KnorpelSenf Addressed in 11c8adae and 662dee46 — the P1 had two halves, and the skip rule now mirrors the Worker's complete routing decision for the apex:

- **`Sec-Fetch-Dest`** (11c8adae): the exemption now also requires the header to be absent, `empty`, or `image`/`video` with `Sec-Fetch-Site: same-origin`, matching the corresponding branch of `canAccessModuleFile()`. `Accept: */*` + `Sec-Fetch-Dest: document` no longer inherits the exemption. Sec-Fetch-* are forbidden request headers — browsers set them, scripts can't, and CLIs never send them — so module fetches satisfy the clause by absence and installs can't be throttled by it.
- **Spoofed bot headers** (662dee46): the skip rule now also requires that no `lb/bots.ts` pattern matches (UA prefixes + the Googlebot `From` header), since `isBot()` runs before the module-file branch and always routes to the frontend. Unlike `waf_re_module_shaped` this mirror must be *exact* in both directions — narrower re-opens the bypass, wider rate-limits installs from whatever it over-matches — so `lb/waf_test.ts` now enforces parity against `isBot()` on a corpus of real bot and CLI header sets.

With these two clauses the skip expression re-derives the full `handleRootRequest` decision: method, path shape, `Accept`, `Sec-Fetch-Dest`/`Site`, and the bot branch. A request the Worker sends to the frontend can no longer be exempted at the WAF by any single-header spoof, and the remaining intentional divergences are all in the safe direction (over-narrow edge cases get challenged, never mis-served).

One note for apply time: the new clauses add two more `matches` uses and an `any(... matches ...)` over a header array — same entitlement as the existing expressions, but worth including when pasting the expression into the dashboard builder for the parse check already flagged in the PR body.

crowlKats:
The first merge-queue run failed at the staging terraform apply: `exceeded the maximum number of rules in the phase http_ratelimit: 2 out of 1`. The two-rule design is sized to the production plan's cap; the staging zone's plan allows one rule in that phase. bc1097c0 gates the general 300/min rule behind `var.production`, so staging deploys only the doc/diff/source rule — which still exercises the interesting classification plus the `managed_challenge` + `mitigation_timeout = 0` combination against the real API.

Silver lining from the same run: `waf_custom` — the ruleset carrying every regex, the `Accept`/`Sec-Fetch-Dest` expressions, and the new bot mirror — was created successfully on the staging zone, so the "does this expression actually parse" risk flagged in the PR body is retired.

crowlKats:
Third staging bounce, different entitlement again: `not entitled: the use of operator Matches is not allowed, an higher Advanced Rate Limiting plan is required`. Regex in *rate limiting* expressions is a separate entitlement from the custom-rules regex (which staging accepted in `waf_custom`). Together with the 1-rule cap and 10s-only period, the staging plan can't express these rules in any form that shares an expression with production — so 00202fa9 reverts the scaled-down staging variant and gates the whole `waf_ratelimit` ruleset behind `var.production`, restoring the two-rule 60s design verbatim.

Consequence worth stating explicitly: **the prod apply is the first real validation of the rate limiting ruleset** (the custom ruleset is fully validated by staging). If prod's grant doesn't cover Advanced Rate Limiting, the prod apply will fail on this resource with the same error — everything else in the plan applies independently, and the fallback would be rewriting the two expressions with plan-universal operators (`starts_with`/`contains`/`ends_with`), at the cost of cruder matching.

PR #1486, opened 2026-08-11
fix(api): reject restricted tokens in check_admin_access

SashaMIT:
## Summary

`check_admin_access` was the only IAM check that did not enforce token permission restrictions. A staff user holding a least-privilege publish/device token (with a `permissions` allowlist) could still call `/api/admin/*` because the handler only checked `user.is_staff`.

Every sibling check already returns `MissingPermission` when `self.permissions.is_some()`. This PR applies the same deny-by-default via `require_unrestricted()`, with `check_publish_access` as the one allowlist exception.

`check_admin_access` also requires an interactive (web) credential, same as authorization approve. That means an unrestricted staff Personal/Device token is rejected on `/api/admin/*` and on `/debug/mem_stats` / `/debug/mem_dump` with `credentialNotInteractive`. That is intentional. Support tickets do not use that path: they go through `check_staff_access`, so an unrestricted staff PAT can still see and reply to tickets. A restricted staff token still gets `missingPermission`.

**Attacker:** anyone who obtains a restricted staff token (for example a scoped publish token belonging to a staff account) and calls `/api/admin/users` (or any other admin route).

## Test plan

- [x] `restricted_staff_token_cannot_access_admin` asserts `GET /api/admin/users` is 403 `missingPermission` for a scoped staff PAT, and 403 `credentialNotInteractive` for an unrestricted staff PAT
- [x] `test_ticket` asserts an unrestricted staff PAT can still `GET` another user's ticket, and a restricted staff token gets `missingPermission`
- [ ] CI `api` suite green

### DIFF
--- a/api/src/api/admin.rs
+++ b/api/src/api/admin.rs
@@ -520,10 +520,11 @@ mod tests {
     // A least-privilege publish token held by a staff user must not carry
     // /api/admin authority. check_admin_access previously skipped the
     // permissions guard that every other IAM check applies.
-    let permission =
-      crate::db::Permission::PackagePublish(crate::db::PackagePublishPermission::Scope {
+    let permission = crate::db::Permission::PackagePublish(
+      crate::db::PackagePublishPermission::Scope {
         scope: t.scope.scope.clone(),
-      });
+      },
+    );
     let restricted = crate::token::create_token(
       &t.db(),
       t.staff_user.user.id,
@@ -545,9 +546,10 @@ mod tests {
       .await;
 
     // Unrestricted staff web token still works.
+    let staff_token = t.staff_user.token.clone();
     t.http()
       .get("/api/admin/users")
-      .token(Some(&t.staff_user.token))
+      .token(Some(&staff_token))
       .call()
       .await
       .unwrap()
--- a/api/src/api/admin.rs
+++ b/api/src/api/admin.rs
@@ -1,35 +1,35 @@
 // Copyright 2024 the JSR authors. All rights reserved. MIT license.
-use crate::external::algolia::AlgoliaClient;
-use crate::s3::Buckets;
 use crate::NpmUrl;
 use crate::RegistryUrl;
+use crate::external::algolia::AlgoliaClient;
+use crate::s3::Buckets;
 use hyper::Body;
 use hyper::Request;
-use routerify::prelude::RequestExt;
 use routerify::Router;
+use routerify::prelude::RequestExt;
 use routerify_query::RequestQueryExt;
-use tracing::field;
-use tracing::instrument;
 use tracing::Instrument;
 use tracing::Span;
+use tracing::field;
+use tracing::instrument;
 
 use crate::db::*;
 use crate::iam::ReqIamExt;
 use crate::ids::ScopeDescription;
 use crate::publish::publish_task;
 use crate::util;
+use crate::util::ApiResult;
+use crate::util::LicenseStore;
+use crate::util::RequestIdExt;
 use crate::util::decode_json;
 use crate::util::pagination;
 use crate::util::search;
 use crate::util::sort;
-use crate::util::ApiResult;
-use crate::util::LicenseStore;
-use crate::util::RequestIdExt;
 
-use super::map_unique_violation;
-use super::types::*;
 use super::ApiError;
 use super::PublishQueue;
+use super::map_unique_violation;
+use super::types::*;
 
 pub fn admin_router() -> Router<Body, ApiError> {
   Router::builder()
--- a/api/src/api/admin.rs
+++ b/api/src/api/admin.rs
@@ -1,35 +1,35 @@
 // Copyright 2024 the JSR authors. All rights reserved. MIT license.
-use crate::NpmUrl;
-use crate::RegistryUrl;
 use crate::external::algolia::AlgoliaClient;
 use crate::s3::Buckets;
+use crate::NpmUrl;
+use crate::RegistryUrl;
 use hyper::Body;
 use hyper::Request;
-use routerify::Router;
 use routerify::prelude::RequestExt;
+use routerify::Router;
 use routerify_query::RequestQueryExt;
-use tracing::Instrument;
-use tracing::Span;
 use tracing::field;
 use tracing::instrument;
+use tracing::Instrument;
+use tracing::Span;
 
 use crate::db::*;
 use crate::iam::ReqIamExt;
 use crate::ids::ScopeDescription;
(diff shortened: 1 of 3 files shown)


SashaMIT:
recheck

SashaMIT:
Pushed a follow-up for CI: cloned `staff_token` before `t.http()` (E0502) and rustfmt on the new test. CLA signed on my side — please re-run the CLA check if it still shows pending.

crowlKats:
Reviewed — nothing blocks merging. The guard is correct and consistent with its five siblings, web/staff-UI tokens never carry `permissions` (so no admin-UI lockout), the `tickets.rs` `is_ok()` fallbacks are already gated upstream by `check_current_user_access`, and the regression test genuinely covers the vulnerable path. A few observations, most-severe first:

1. **Sixth hand-copied guard** (`api/src/iam.rs:247`) — this is the sixth verbatim copy of `if self.permissions.is_some() { return Err(ApiError::MissingPermission); }` (lines 37, 64, 104, 219, 232, and now 247), and the opt-in copy-paste design is the root cause of the hole this PR fixes: the next `check_*` method written without the guard silently regains full authority for restricted tokens. The copies are already drifting (the comment at `iam.rs:105` says "scope admin access", copied from the wrong method). A private `require_unrestricted()` helper — or a checked `principal()` accessor that denies when `permissions` is set — with `check_publish_access` as the one explicit exception would flip the default from allow to deny, confined to `iam.rs` with no call-site churn. Could land here or as a follow-up.

2. **Unrestricted staff PATs still carry full admin authority** (pre-existing, not a defect of this PR) — a staff Personal/Device token created with `permissions: null` still passes `check_admin_access`, so a leaked staff PAT grants all `/api/admin` routes, version delete, and the heap-dump endpoint. `check_authorization_approve_access` already requires `self.interactive` for the same reason; admin access could require the same. (`sudo` is no mitigation — it's derived from the client-supplied `x-jsr-sudo` header.)

3. **Nit:** the `Principal::GitHubActions` arm of `check_admin_access` is now runtime-unreachable (OIDC `IamInfo` always sets `permissions: Some(...)`), and the error code for OIDC callers changes from `actorNotAuthorized` to `missingPermission` — misleading, since no permission can ever grant admin. Same pattern as the sibling checks; worth a comment at most.

4. **Nit:** the new test inlines fully-qualified `crate::db::...`/`crate::token::create_token` paths instead of extending the tests module's `use` block, diverging from the module's import style and the analogous restricted-token test in `package.rs`.

SashaMIT:
Thanks for the review. Glad the guard and the test landed clean.

On (1), I agree the sixth copy is the same class that made the hole. Happy to fold a require_unrestricted helper into this PR if you want it here, or I can open a follow-up after this merges. (2) and the nits I will leave as follow-up unless you want them in this one.

crowlKats:
@SashaMIT i would prefer them in this PR

SashaMIT:
Got it. Folded them into this PR.

require_unrestricted() is now the default for every check except publish. Admin also requires interactive, same as approve, so an unrestricted staff PAT gets credentialNotInteractive. Comment on the OIDC arm, and the test imports match package.rs.

crowlKats:
Code review (Claude Code). Overall: the PR correctly fixes the real vulnerability, and the `require_unrestricted()` refactor is behavior-identical for the five pre-existing guards. A few findings, most severe first:

**1. Collateral regression in `api/src/api/tickets.rs` (confirmed)**
`get_handler` (~L55) and `post_message_handler` (~L148) gate non-creator access via `iam.check_admin_access().is_ok()`, falling through to `TicketNotFound`. The new interactive+unrestricted requirement means staff on Personal/Device tokens (or any token with a permissions allowlist) silently lose ticket visibility/replies, masked as a 404. `tickets.rs` isn't touched by this diff and no test covers the change.

**2. Broader breaking change than the PR text discloses (confirmed)**
The `self.interactive` guard also rejects fully *unrestricted* staff PATs from all of `/api/admin/*`. That's deliberate and tested here, but the title/description only mention restricted tokens — worth stating explicitly that non-web staff credentials break too.

**3. `/debug/mem_stats` and `/debug/mem_dump` locked out for PATs (plausible)**
Both handlers in `api/src/jemalloc_profiling.rs` use `check_admin_access`, so PAT-based `curl` during an incident now gets 403 `credentialNotInteractive`. Neither handler has test coverage, so nothing signals whether this lockout is intended.

**4. The guard is still per-method opt-in (confirmed)**
`require_unrestricted()` must be hand-called at the top of six `check_*` methods — the exact shape that let `check_admin_access` miss it originally. A typestate split (`fn unrestricted(&self) -> Result<UnrestrictedIam<'_>, ApiError>` hosting the six guarded methods, `check_publish_access` staying on the base type) would make omission a compile error. Alternatively, an `iam.rs` test asserting every `check_*` rejects `Some(permissions)`.

**5. "Unreachable" comment on the `GitHubActions` arm (confirmed, minor)**
`check_admin_access`'s comment (~L242) states a cross-file convention (OIDC `IamInfo` always sets `permissions: Some`) as fact, and only one of six identically-dead `GitHubActions` arms carries it — inviting the inference that the other five are live, or that the deny arm is safe to remove. Better stated once on `require_unrestricted()`.

**6. Redundant third test block (confirmed, minor)**
The "Unrestricted staff web token still works" section of `restricted_staff_token_cannot_access_admin` (~L574) duplicates what `list_users` already asserts more thoroughly in the same module (same endpoint, same token, plus item counts and a filtered query). Can be dropped.

**7. Interactive/staff guard hand-rolled twice (plausible, minor)**
`check_admin_access` duplicates the `user.is_staff` guard across two match arms, and the interactive-only invariant is separately hand-coded in `check_authorization_approve_access` — a shared `require_interactive()`-style helper would keep them in lockstep.

crowlKats:
sorry for more annoying reviews, however anything touching permissions & tokens needs to be handled very carefully

SashaMIT:
No problem at all and good due dilligence :) Happy to work together

SashaMIT:
Appreciate you walking the tickets path. That PAT 404 was real.

Tickets now use `check_staff_access` (unrestricted staff, no interactive), so a staff PAT can still see and reply. `/api/admin` and the debug heap endpoints stay interactive-only. Called that out in the PR text.

Also moved the OIDC note onto `require_unrestricted`, shared `require_interactive` with approve, dropped the duplicate web-token admin assert (`list_users` already covers it), and added the staff-PAT ticket cases.

Left a full typestate split for a follow-up. Happy to do that next if you still want omission to be a compile error.

PR #1492, opened 2026-08-21
docs: add terms of service and privacy policy

littledivy:
Adds a Terms of Service and a Privacy Policy for JSR, and links both from the **Reference** section of the docs TOC (right after "Usage policy").

⚠️ **These documents are adapted from Deno Deploy's [Terms and Conditions](https://docs.deno.com/deploy/terms_and_conditions/) and [Privacy Policy](https://docs.deno.com/deploy/privacy_policy/), and require review by legal counsel before merging.** They were written by adapting Deno's existing texts, not drafted by a lawyer for JSR specifically. The account deletion and data retention section in particular makes concrete factual claims about what JSR retains and asserts specific legal bases for that retention — it needs to be checked against both the law and the shipped behavior.

### What was adapted

- Deploy/Subhosting product descriptions replaced with JSR: a public registry for JavaScript and TypeScript packages at https://jsr.io, operated by Deno Land Inc.
- Dropped the Paid Products paragraph and the entire FEES/Subscription section; the ToS instead states JSR is currently provided free of charge, with Deno reserving the right to change its offerings. Sections renumbered accordingly.
- Dropped Deploy-specific language (deployments, regions, V8 isolates) and the reserved right to "change the regions in which the Services run", replaced with registry-appropriate rights (blocking packages/versions/scopes, changing quotas and limits).
- "User Data" reworked into **User Content**: publishing makes content public, published versions are immutable, and you grant Deno a license to host and distribute your content — surviving account deletion to the extent needed to keep published versions available. References the existing [usage policy](https://jsr.io/docs/usage-policy).
- Account credentials now describe authentication via an identity provider such as GitHub.
- Contact email is help@jsr.io throughout. Deno Land Inc.'s San Diego mailing address is kept in the privacy policy.
- Arbitration, class action waiver, NY governing law, warranty disclaimers, limitation of liability, and indemnification are kept essentially verbatim.

### Account deletion and data retention (new — needs the closest legal review)

Section V of the privacy policy discloses the behavior introduced in #1331:

- **Hard-deleted:** name, email address, avatar.
- **Retained as a pseudonymous tombstone:** the internal account UUID (`meta.original_actor_id` on the user's former audit entries) and the identity-provider numeric IDs (`github_id`, `gitlab_id`) in the `user_delete` audit entry, so the account can be re-identified through the provider if legally required. Stated retention bases: GDPR Art. 17(3)(b) and 17(3)(e), and CCPA §1798.105(d).
- **Reassigned to the system service account:** audit log entries, and support tickets/messages.
- **Published packages:** immutable and remain public after deletion, including any personal information the publisher included in package contents or metadata. Sole-owned scopes transfer to the service account; scopes with other members transfer to another member.
- **Deletion holds:** the admin-settable `deletion_hold` flag blocks account deletion while a legal or moderation matter is pending (evidence preservation / litigation hold).

### Judgment calls worth a look

- I disclosed the **support ticket / ticket message reassignment** to the service account as well. It wasn't in the original list of behaviors to document, but `delete_user` does reassign `tickets` and `ticket_messages`, and leaving a real retention behavior out of the policy seemed like the wrong default. Drop it if you'd rather keep the section narrower.
- I **dropped the DPA reference** from the ToS preamble. Deno's terms incorporate a Data Processing Addendum, which as far as I can tell doesn't exist for JSR; referencing a non-existent document seemed worse than omitting it. Restore it if a DPA does apply.
- Dates are set to today and the copyright lines to 2026 — adjust to whatever the actual effective date should be.
- The limitation of liability clause is kept verbatim, including the "monies you have paid us in the preceding twelve months" prong, which for a free service means the $100 floor always governs. Left as-is deliberately, but it reads oddly for a free registry.

`deno fmt` and `deno task lint:frontend` both pass.

Opened as a draft — this shouldn't merge until counsel has signed off.

### DIFF
--- a/frontend/docs/privacy-policy.md
+++ b/frontend/docs/privacy-policy.md
@@ -1,0 +1,396 @@
+---
+description: This privacy policy describes the personal information JSR collects, how it is used, and your choices with respect to it.
+---
+
+_Last updated: August 21st, 2026_
+
+Deno Land Inc. (“Deno,” “we,” “us,” or “our”) collects and uses personal
+information in order to provide its products and services to you. This Privacy
+Policy (the “Policy”) describes the personal information we collect, the
+purposes for which we use it, the parties with whom we may share it, and your
+choices with respect to such information. For purposes of this Privacy Policy,
+“personal information” means any information that relates to you as an
+individual and could reasonably be used to identify you. This Privacy Policy
+applies to our collection and use of personal information through (i) our
+website at [https://jsr.io](https://jsr.io) (the “Site”); (ii) any websites,
+applications or other digital properties that link to this Privacy Policy; and
+(iii) the products and services (the “JSR Offerings”) we offer to you on our
+platform (the “Platform”), including the JSR registry, the JSR website, and the
+JSR API.
+
+By accessing or using the Site or any other digital property that links to this
+Privacy Policy, you may learn about JSR and our technology platform, and
+registered users may also access the JSR Offerings (collectively, the
+“Services”). To the extent permitted by applicable law, your use of JSR
+constitutes your acknowledgment and/or consent to the practices described in
+this Policy.
+
+This Privacy Policy incorporates our [Terms of Service](/docs/terms-of-service)
+(the “Terms”). Capitalized terms that are not defined in the Privacy Policy have
+the meaning given to them in the Terms.
+
+**I. The Information We Collect, And How We Collect It**
+
+We collect the following categories of information, which may include personal
+information (collectively, the “**Information**”).
+
+**1\. Information You Provide To Us**
+
+We collect information from and about you directly when you provide it to us.
+This information may be collected when you contact us, fill out a form, create
+an account, publish a package, open a support ticket, access or participate on
+our Sites, or otherwise interact with us. This information may include:
+
+_Contact Information._ We collect your contact information when you voluntarily
+provide it to us. For example, you may disclose contact information to us when
+you contact us for support, submit information by mail, telephone, in person or
+electronically, or when you sign up for our newsletters and other marketing
+communications. Contact Information typically includes first name, last name,
+e-mail address, postal address, organization, telephone number and other
+information that identifies you or can be used to identify or contact you.
+
+_Account Credentials_. When you register to create an account with us, you
+authenticate via an identity provider such as GitHub, and we will collect
+certain additional personal information, including your name, email address,
+avatar, and the user name, numeric user identifier, and public profile
+information associated with your identity provider account.
+
+_User Content_. We collect the packages, package versions, source code,
+documentation, package metadata, and scope names that you publish to JSR, along
+with records of your publishing activity. Please note that JSR is a public
+registry: everything you publish, and the scopes and packages associated with
+your account, are publicly visible. Do not publish confidential information,
+credentials, or personal information that you do not wish to make public.
+
+In addition to Contact Information, Account Credentials, and User Content, we
+may collect other kinds of information, such as:
+
+- Comments, questions, and requests you may make, including support tickets and
+  the messages within them;
+
+- Information about your preferences, such as your preferred methods of
+  communication and the types of information in which you are interested;
+
+- Details of downloads from our Sites;
+
+- Records and copies of your correspondence (including email addresses and phone
+  numbers), if you contact us; and
+
+- Any other information you voluntarily provide.
+
+**2\. Information Obtained From Third Parties**
+
+We may receive certain information about you from other sources, including
+publicly available sources (such as public records and social media platforms),
+as well as our service providers and marketing partners. In particular, when you
+authenticate to JSR through an identity provider such as GitHub, that provider
+shares certain account information with us as described above.
+
+When we collect personal information from users and visitors of other sites on
+which you have interacted with us, we will do so in accordance with the terms of
+use and privacy policies of those sites and applicable law. We may also receive
+personal information when you comment on our social media advertisements, post
+comments about us, or tag us in a public-facing social media post. Personal
+information may also be collected by the third-party social media sites that
+host our social media pages. These sites may provide aggregate information and
+analysis to us about visitors’ use of our social media pages. This allows us to
+better understand and analyze our user growth, general demographic information
+about the users of these pages, and interaction with the content that we post.
+Overall, this information may be used to help us understand the types of
+visitors and users of our social media pages and use of the content. This
+Privacy Policy does not cover personal information collected by such third-party
+social media sites. For more information on their privacy and security practices
+please review the privacy policies and terms of use on their respective
+websites.
+
+**3\. Information Collected Automatically**
+
+We and our service providers may automatically obtain certain information about
+you, your electronic device, and your interactions with us, including the
+following:
+
+- _Device data_. We may collect data such as the type of device and its
+  operating system and settings, browser type, mobile device carrier, country,
+  IP address, and unique identifiers.
+
+- _Internet and other or electronic activity data_. This includes information
+  about your interaction with our Sites, emails, and other online content.
+
+- _Tracking Data_. We may collect tracking data using first and third-party
+  cookies, pixels, web server logs, web beacons, and similar data collection and
+  tracking technologies on the Sites, third party websites, apps and online
+  services, and across your devices (such as IP address, browser, type, ISP,
+  platform type, device type). Third parties such as advertising networks and
+  analytics providers may also collect information about your online activities
+  over time and across different websites and devices when you access or use the
+  Sites.
+
+- _Audit logs_. We keep an audit log of security- and moderation-relevant
+  actions taken on the Platform, such as publishing a package, changing scope
+  membership, and administrative actions. Audit log entries record the account
+  that performed the action and the time at which it was performed.
+
+**II. How We Use And Share Your Information**
+
+Deno uses the Information for the purpose for which it was collected and in a
+manner that is consistent with this Privacy Policy. These functions include
+operation, maintenance and improvements to the Sites, providing our products and
+services, operating the registry and distributing published packages,
+solicitation of your feedback, gaining a better understanding of our customers
+and visitors of our Sites, responding to your requests and questions, and
+informing you about our organization, products, services, events, and other
+areas of interest.
+
+_Public Information_. Certain information is published as part of operating a
+public registry. Your user name, display name, and avatar, the scopes you belong
+to, and the packages and package versions you publish are publicly visible
+through the Site and the JSR API, and may be downloaded, cached, and mirrored by
+third parties.
+
+_Analytics Services_. We may use third-party web analytics services, such as
+Google Analytics, to help us understand and analyze how Site visitors use our
+services. For more information on how Google Analytics uses data collected
+through our Sites, visit
+[www.google.com/policies/privacy/partners](http://www.google.com/policies/privacy/partners).
+
+_Aggregated Data_. We may analyze your personal information in aggregate form
+which does not identify you personally (“**Aggregated Data**”). The Aggregated
+Data may be used to operate, maintain, manage, and improve the Sites, shared
+with our affiliates, agents, and business partners, and otherwise used and
+disclosed for lawful business purposes. We do not re-identify de-identified or
+aggregated information.
+
+_Service Providers/Vendors_. Like many businesses, we hire other companies to
+perform certain business-related services. We may disclose personal information
+to certain types of third party companies but only to the extent needed to
+enable them to provide such services, for example web hosting, content delivery,
+disaster recovery, client survey and marketing, and data storage.
+
+_Reorganization_. If, in the future, Deno undergoes a corporate, partnership, or
+business reorganization, we may transfer the Information, including personal
+information, to the new or surviving entity.
+
+_Protection of Rights and Compliance_. We may use your Information to protect
+the rights, privacy or safety of you, us or others; to ensure our compliance
+with legal and contractual requirements; and to prevent and investigate illegal,
+unethical, or unauthorized activities (including cyberattacks and identity
+theft).
+
+If Deno intends on using or disclosing your personal information in any manner
+that is not consistent with this Privacy Policy, you will be informed of such
+anticipated use prior to or at the time at which the personal information is
+collected.
+
+**III. How We Protect Your Information**
+
+We take commercially reasonable steps to protect your personal information from
+loss, misuse, and unauthorized access, disclosure, alteration, or destruction.
+Please understand, however, that no security system is impenetrable. We cannot
+guarantee the security of our databases, nor can we guarantee that the personal
+information that you supply will not be intercepted while being transmitted to
+and from us over the Internet.
+
+**IV. Data Retention**
+
+Deno determines the retention period for all Information based on the purposes
+for which we collect and/or receive the Information and/or tax, legal and
+regulatory requirements. In addition to this, we may consider other factors,
+such as the nature and sensitivity of the data, and whether we can achieve the
+purpose for which we collected the data through other means.
+
+**V. Account Deletion And Data Retention**
+
+You may delete your JSR account at any time from your account settings. This
+section describes what happens to your Information when you do.
+
+**1\. Information That Is Deleted**
+
+When you delete your account, your account record is permanently deleted from
+our systems. This includes your name, your email address, and your avatar, which
+are not retained.
+
+**2\. Information That Is Retained**
+
+Deleting your account does not erase the record that the account existed or the
+actions it took on the Platform. Specifically:
+
+- _Audit log entries._ Audit log entries created by your account are reassigned
+  to a system service account, so that the audit trail of actions taken on the
+  Platform remains complete and attributable.
+
+- _Pseudonymous identifiers._ We retain a minimal, pseudonymous record of the
+  deleted account. Each of your former audit log entries retains your internal
+  account identifier (a UUID, stored as `meta.original_actor_id`), and the audit
+  log entry recording the deletion itself (`user_delete`) retains that internal
+  account identifier together with the numeric user identifiers assigned to you
+  by your identity provider (for example, your GitHub or GitLab numeric user
+  ID). This record allows the account to be re-identified through the identity
+  provider if we are legally required to do so. It does not include your name,
+  email address, or avatar.
+
+- _Support tickets._ Support tickets you created and messages you sent within
+  them are reassigned to the system service account and retained.
+
+We retain this information because it is necessary for compliance with a legal
+obligation to which we are subject, for reasons of security and the prevention
+and investigation of fraud and abuse, and for the establishment, exercise, or
+defense of legal claims. We rely on Article 17(3)(b) and Article 17(3)(e) of the
+EU General Data Protection Regulation and on Section 1798.105(d) of the
+California Consumer Privacy Act as the bases for retaining it notwithstanding a
+deletion request.
+
+**3\. Published Packages Remain Public**
+
+Published package versions are immutable and are not removed when you delete
+your account. They remain publicly available through the Site, the JSR API, and
+any third-party caches and mirrors, so that packages which depend on them
+continue to resolve. Any personal information you chose to include in the
+contents or metadata of a package you published — for example, an author name or
+email address in a source file or manifest — remains part of that published
+version and remains public. See our [immutability](/docs/immutability)
+documentation for more detail.
+
+If you are the only member of a scope, that scope and its packages are
+transferred to the system service account when your account is deleted. If a
+scope you created has other members, ownership of that scope is transferred to
+another member.
+
+**4\. Deletion Holds**
+
+Our administrators can place a deletion hold on an account. While a deletion
+hold is in place, the account cannot be deleted, and a deletion request made
+from account settings will not proceed. We use deletion holds to preserve
+evidence while a legal or moderation matter concerning an account is pending —
+for example, an active abuse investigation, a dispute over a package or scope,
+or a legal claim or litigation hold. The hold is removed once the matter is
+resolved, after which the account can be deleted as described above.
+
+If you have questions about a deletion hold on your account, contact us at
+[help@jsr.io](mailto:help@jsr.io).
+
+**VI. Your Privacy Choices**
+
+**1\. Your Information**
+
+You may request access to, correction of, or deletion of the personal
+information we maintain about you, and we will endeavor to respond promptly to
+your request. In order to make such a request, please contact us as indicated
+below. Please see the section titled “Account Deletion And Data Retention” above
+for information about what is retained when an account is deleted.
+
+**2\. Marketing Communications**
+
+You may opt-out of marketing-related emails by clicking on the “unsubscribe”
+link located on the bottom of any marketing email or emailing us at
+[help@jsr.io](mailto:help@jsr.io). We will use commercially reasonable efforts
+to process such requests in a timely manner. Please note that even if you
+opt-out of marketing-related emails, you will continue to receive
+service-related and other non-marketing emails.
+
+**3\. Tracking Technology**
+
+You can choose not to permit tracking technologies, such as cookies and web
+beacons, when you use our services, but blocking some types of these tracking
+technologies may interfere with your experience.
+
+_Browser-Based Opt-Outs_. You may be able to disable tracking technologies using
+your web browser settings. Please review your browser’s instructions or visit
+[All About Cookies](https://allaboutcookies.org/) for general information. Note
+that your web browser may have settings that allow you to transmit a “Do Not
+Track” signal when you use online services. Like many websites, our Sites are
+not currently designed to respond to “Do Not Track” signals received from
+browsers.
+
+_Self-Regulatory Program Opt-Outs_. Two self-regulatory programs are available
+to help you control the use of tracking technologies on your browsers — the
+[Digital Advertising Alliance](https://digitaladvertisingalliance.org/) and the
+[Network Advertising Initiative](https://thenai.org/). Both programs help to
+regulate vendors in the digital advertising space. One function of their
+self-regulatory programs is to give you the ability to opt out of targeted (or
+interest-based) advertising, including the use of tracking technologies, from
+their member companies. You can visit the Digital Advertising Alliance’s Your Ad
+Choices website to opt out of targeted advertising for participating vendors.
+The Network Advertising Initiative similarly assists with opt outs through their
+Opt Out of Interest-Based Advertising webpage.
+
+_Google Analytics Opt-Out._ To opt out of Google Analytics cookies, visit
+Google’s [My Ad Center](https://myadcenter.google.com/personalizationoff) and/or
+download the Google Analytics Opt-Out Browser Add-On.
+
+**VII. Children**
+
+We do not knowingly collect personal information from children under the age of
+18 through the Sites. If you are under 18, please do not give us any personal
+information. We encourage parents and legal guardians to monitor their
+children’s Internet usage and to help enforce our Privacy Policy by instructing
+their children never to provide personal information through the Sites without
+their permission. If you have reason to believe that a child under the age of 18
+has provided personal information to us, please contact us, at
+[help@jsr.io](mailto:help@jsr.io) and we will endeavor to delete that
+information from our databases.
+
+**VIII. External Websites**
+
+The Sites may contain links to third-party websites. These third-party sites may
+collect information about you if you click on a link. We have no control over
+the privacy practices or the content of these websites. As such, we are not
+responsible for the content or the privacy policies of those third-party
+websites. You should check the applicable third-party privacy policy and terms
+of use when visiting any other websites.
+
+**IX. Important Notice To Non-U.S. Residents**
+
+The Sites are hosted in and provided from the United States and other countries.
+If you are located outside of the United States, please be aware that any
+information you provide to us may be transferred to the United States or other
+countries where the privacy laws may not be as protective as those in your
+country of origin. If you are located outside the United States and choose to
+use the Sites, you consent to any transfer and processing of your personal
+information in accordance with this Privacy Policy, and you do so at your own
+risk.
+
+**X. Notice To California Residents**
+
+Pursuant to Section 1798.83 of the California Civil Code, residents of
+California have the right to obtain certain information about the types of
+personal information that companies with whom they have an established business
+relationship (and that are not otherwise exempt) have shared with third parties
+for direct marketing purposes during the preceding calendar year, including the
+names and addresses of those third parties, and examples of the types of
+services or products marketed by those third parties. In order to submit such a
+request, please contact us using the contact information provided at the end of
+this document. Please note, however, that we do not share, nor have we shared in
+the past, personal information with third parties for direct marketing purposes.
+
+**XI. Notice To Nevada Residents**
+
+If you are a resident of Nevada, you have the right to opt-out of the sale of
+personal information to third parties. You can exercise this right by contacting
+us at [help@jsr.io](mailto:help@jsr.io) with the subject line “Nevada Do Not
+Sell Request” and providing us with your name and the email address. Please
+note, however, that we do not sell any personal information to third parties.
+
+**XII. Changes To This Privacy Policy**
+
+This Privacy Policy is effective as of the date stated at the top of this
+Privacy Policy. We may change this Privacy Policy from time to time. Any such
+changes will be posted on the Sites. By accessing the Sites after we make any
+such changes to this Privacy Policy, you are deemed to have accepted such
+changes. Please be aware that, to the extent permitted by applicable law, our
+use of the Information is governed by the Privacy Policy in effect at the time
+we collect the Information. Please refer back to this Privacy Policy on a
+regular basis.
+
+**XIII. How To Contact Us**
+
+Please reach out to [help@jsr.io](mailto:help@jsr.io) for any questions,
+complaints, or requests regarding this Privacy Policy, and include in the
+subject line “Privacy Policy", or contact us by mail at:
+
+Deno Land Inc.\
+1111 6th Ave Ste 550\
+PMB 702973\
+San Diego CA, 92101\
+USA
+
+**© 2026 Deno Land Inc. All rights reserved.**
--- a/frontend/docs/privacy-policy.md
+++ b/frontend/docs/privacy-policy.md
@@ -42,12 +42,9 @@ an account, publish a package, open a support ticket, access or participate on
 our Sites, or otherwise interact with us. This information may include:
 
 _Contact Information._ We collect your contact information when you voluntarily
-provide it to us. For example, you may disclose contact information to us when
-you contact us for support, submit information by mail, telephone, in person or
-electronically, or when you sign up for our newsletters and other marketing
-communications. Contact Information typically includes first name, last name,
-e-mail address, postal address, organization, telephone number and other
-information that identifies you or can be used to identify or contact you.
+provide it to us — for example, when you open a support ticket or email us for
+help. In practice this is your email address and whatever you choose to put in
+your message. We do not run a newsletter and we do not send marketing email.
 
 _Account Credentials_. When you register to create an account with us, you
 authenticate via an identity provider such as GitHub, and we will collect
@@ -80,28 +77,12 @@ may collect other kinds of information, such as:
 
 **2\. Information Obtained From Third Parties**
 
-We may receive certain information about you from other sources, including
-publicly available sources (such as public records and social media platforms),
-as well as our service providers and marketing partners. In particular, when you
-authenticate to JSR through an identity provider such as GitHub, that provider
-shares certain account information with us as described above.
-
-When we collect personal information from users and visitors of other sites on
-which you have interacted with us, we will do so in accordance with the terms of
-use and privacy policies of those sites and applicable law. We may also receive
-personal information when you comment on our social media advertisements, post
-comments about us, or tag us in a public-facing social media post. Personal
-information may also be collected by the third-party social media sites that
-host our social media pages. These sites may provide aggregate information and
-analysis to us about visitors’ use of our social media pages. This allows us to
-better understand and analyze our user growth, general demographic information
-about the users of these pages, and interaction with the content that we post.
-Overall, this information may be used to help us understand the types of
-visitors and users of our social media pages and use of the content. This
-Privacy Policy does not cover personal information collected by such third-party
-social media sites. For more information on their privacy and security practices
-please review the privacy policies and terms of use on their respective
-websites.
+When you authenticate to JSR through an identity provider — GitHub or GitLab —
+that provider shares certain account information with us: your name, email
+address, avatar, username, and numeric user identifier. That is the only routine
+source of information about you that we receive from a third party. We do not
+buy personal information, and we do not receive it from data brokers or
+marketing partners.
 
 **3\. Information Collected Automatically**
 
@@ -109,26 +90,31 @@ We and our service providers may automatically obtain certain information about
 you, your electronic device, and your interactions with us, including the
 following:
 
-- _Device data_. We may collect data such as the type of device and its
-  operating system and settings, browser type, mobile device carrier, country,
-  IP address, and unique identifiers.
+- _Server logs_. Our servers and our CDN record requests made to JSR, including
+  your IP address, user agent, the path requested, and the time of the request.
 
-- _Internet and other or electronic activity data_. This includes information
-  about your interaction with our Sites, emails, and other online content.
+- _Download counts_. When a package is downloaded we record the scope, package
+  name, version, and the user agent that requested it, so that we can show
+  download statistics. We do not record your IP address alongside this.
 
-- _Tracking Data_. We may collect tracking data using first and third-party
-  cookies, pixels, web server logs, web beacons, and similar data collection and
-  tracking technologies on the Sites, third party websites, apps and online
-  services, and across your devices (such as IP address, browser, type, ISP,
-  platform type, device type). Third parties such as advertising networks and
-  analytics providers may also collect information about your online activities
-  over time and across different websites and devices when you access or use the
-  Sites.
+- _Search_. Search on JSR is powered by Algolia. The text you type into the
+  search box is sent to Algolia in order to return results. We do not tell
+  Algolia which result you clicked, and we do not keep any identifier that would
+  let searches be linked to each other or to you.
+
+- _Bot protection_. Our login page uses Cloudflare Turnstile to distinguish
+  people from automated abuse. Turnstile receives signals from your browser in
+  order to make that determination.
 
 - _Audit logs_. We keep an audit log of security- and moderation-relevant
   actions taken on the Platform, such as publishing a package, changing scope
   membership, and administrative actions. Audit log entries record the account
   that performed the action and the time at which it was performed.
+
+We do not track you. JSR contains no advertising networks, no third-party
+analytics or advertising trackers such as Google Analytics, no tracking pixels,
+and no cross-site tracking of any kind. We do not sell advertising, and we do
+not profile you or build a record of your behaviour across visits.
 
 **II. How We Use And Share Your Information**
 
@@ -147,11 +133,23 @@ to, and the packages and package versions you publish are publicly visible
 through the Site and the JSR API, and may be downloaded, cached, and mirrored by
 third parties.
 
-_Analytics Services_. We may use third-party web analytics services, such as
-Google Analytics, to help us understand and analyze how Site visitors use our
-services. For more information on how Google Analytics uses data collected
-through our Sites, visit
-[www.google.com/policies/privacy/partners](http://www.google.com/policies/privacy/partners).
+_Service Providers._ JSR is built on a small number of third-party services, and
+those providers process data on our behalf in order to run the registry. They
+are:
+
+| Provider          | What it does for JSR                                                                                                          |
+| ----------------- | ----------------------------------------------------------------------------------------------------------------------------- |
+| Google Cloud      | Runs the API, the database, background job queues, scheduled jobs, secrets, and log storage                                   |
+| Cloudflare        | CDN and edge routing, package and documentation storage (R2), download counts, and Turnstile bot protection on the login page |
+| Algolia           | Powers package and documentation search                                                                                       |
+| Postmark          | Sends transactional email — scope invitations, access token notices, and support ticket replies                               |
+| GitHub and GitLab | Provide authentication when you sign in                                                                                       |
+
+We use these providers only to operate the Services. This list is the complete
+set of third parties that receive personal information from JSR; if it changes
+we will update this page. Our infrastructure is defined in public Terraform in
+the [JSR repository](https://github.com/jsr-io/jsr), so this list can be checked
+against what we actually run.
 
 _Aggregated Data_. We may analyze your personal information in aggregate form
 which does not identify you personally (“**Aggregated Data**”). The Aggregated
@@ -181,7 +179,36 @@ that is not consistent with this Privacy Policy, you will be informed of such
 anticipated use prior to or at the time at which the personal information is
 collected.
 
-**III. How We Protect Your Information**
+**III. Legal Bases For Processing (EEA and UK)**
+
+If you are in the European Economic Area or the United Kingdom, we process your
+personal information on the following bases:
+
+- _Performance of a contract._ We process your account information and the
+  content you publish because it is necessary to provide the registry to you
+  under our [Terms of Service](/docs/terms-of-service).
+
+- _Legitimate interests._ We process server logs, download counts, search
+  analytics, audit logs, and bot-protection signals because we have a legitimate
+  interest in keeping JSR secure, available, and useful, and in understanding
+  how it is used. We balance that interest against your rights, which is why
+  these records are minimal and short-lived.
+
+- _Consent._ Where we rely on your consent — for example, for storage on your
+  device that is not strictly necessary to deliver the Services — you may
+  withdraw it at any time, and withdrawing it will not affect processing that
+  took place beforehand.
+
+- _Legal obligation._ We process and retain information where the law requires
+  it, and to establish, exercise, or defend legal claims.
+
+You have the right to access, correct, delete, restrict, and object to our
+processing of your personal information, and to receive a copy of it in a
+portable format. To exercise any of these rights, contact
+[help@jsr.io](mailto:help@jsr.io). You also have the right to lodge a complaint
+with your local data protection supervisory authority.
+
+**IV. How We Protect Your Information**
 
 We take commercially reasonable steps to protect your personal information from
 loss, misuse, and unauthorized access, disclosure, alteration, or destruction.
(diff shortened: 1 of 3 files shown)


piscisaureus (changes requested):
I compared these documents against what crates.io actually publishes (the crates.io usage policy, the Rust Foundation privacy notice, and their data access page), and checked every factual claim in the privacy policy against the code on `main` and the open PRs. The direction is right and parts of this are better than anything crates.io has, but several claims are not true as of today, a couple won't be true even after the open PRs land, and there is leftover Deploy boilerplate that contradicts the new text.

## Blocking: claims that are false on `main`

**1. The entire account deletion section (privacy §VI) describes #1331, which is unmerged.** On `main` there is no deletion capability at all — `Database::delete_user` is `#[cfg(test)]`-gated, and the settings page says to email help@jsr.io (`frontend/routes/account/settings.tsx:100`). Checked claim-by-claim against the #1331 branch, the section is otherwise a faithful description: the tombstone with the provider numeric IDs, the one-year name/email retention with a daily scheduled job that strips them (365-day cutoff, tested), audit-log reattribution to the service account, ticket reassignment, sole-scope transfer, and the `deletion_hold` flag all match. This PR must not take effect before #1331 is merged and deployed.

**2. The Algolia claims are false until #1493 lands.** "We do not tell Algolia which result you clicked, and we do not keep any identifier that would let searches be linked" — on `main` we run a full Insights click-and-conversion pipeline keyed by a persistent per-browser UUID in `localStorage` (`frontend/utils/algolia_insights.ts`, `frontend/islands/SearchConversion.tsx`, `clickAnalytics: true` in both `GlobalSearch.tsx` and `packages.tsx`). #1493 removes all of it. The "no third-party analytics" claim is also only defensible after that PR — the Insights beacon is an analytics beacon. Same merge-order requirement.

**3. "Expired sign-in state is deleted hourly, and is never kept longer than an hour" — wrong on both counts, and no open PR fixes it.** The cleanup job runs daily, not hourly (`terraform/scheduler.tf` — `clean-oauth-states`, `schedule = "0 0 * * *"`; its own description says "older than 1 hour", so the cron and the intent already disagree). And it only cleans `oauth_states`: expired `tokens` and `authorizations` rows are never deleted, only filtered out at read time. Either make the job hourly and add cleanup for the other two tables, or soften the claim.

**4. The provider table claims to be "the complete set" but omits Grafana Cloud.** The API and both Cloudflare workers export OTLP logs and traces to grafana.net (`terraform/prod.tfvars`, `api/src/tracing.rs`, `terraform/lb.tf`, `terraform/cloudflare_frontend.tf`). Worse, the exported spans contain personal data: the `#[instrument]` fields on user create/update include `user.email`, `user.name`, and `user.github_id` (`api/src/db/database.rs:233,257,285`), and every request span carries the full URL and user agent (`api/src/traced_router.rs`). We should (a) drop email/name from those instrument fields, (b) add Grafana to the table, and (c) pin its retention — the "30 days, verifiable in Terraform" claim only holds for the GCP `_Default` bucket (`terraform/logging.tf`); Grafana retention is configured nowhere in this repo. The Rust Foundation notice's Sentry disclosure ("we've configured Sentry to delete [the IP] as soon as it's received") is the model to copy here.

**5. The browser storage list is not "the complete list".** Also present: `darkMode` and `preferredUsage` in `localStorage`; the `gh-login-*` / `gl-username-*` username caches (`frontend/utils/client_cache.ts`); the `oauth_login_csrf` cookie set during the login flow (`api/src/auth/mod.rs`); the staff-only `sudo` cookie (`frontend/islands/UserMenu.tsx`); Cloudflare's own edge cookies; and, until #1493, `algolia:userToken` / `algolia:pendingConversion`. Either enumerate honestly or drop the completeness claim.

## Inaccurate vs #1331 as written

- **"An anonymous placeholder account" is not what the code does.** Deleted users simply vanish: version pages render no publisher (`package_versions.user_id` goes NULL), and scope memberships cascade away. The only stand-in is the sole-member-scope case, which gets the visible, named "JSR" bot account — not anonymous. Describe the actual behavior, or implement a placeholder.

## Wording fixes (code is fine, text overclaims)

- **"cannot be modified or deleted"**: modification, true. Deletion — a staff-only version delete endpoint exists (`api/src/api/package.rs`, blocked only when the package has JSR dependents), and `docs/immutability.md` already documents exceptional removal. Suggest: "contents cannot be modified. Versions cannot be deleted by users; JSR staff may remove a version in exceptional cases." Right now the ToS contradicts both the shipped code and our own docs.
- **Yank**: yanked versions stay resolvable on JSR and direct tarball URLs keep working, but they are omitted from the npm-compat manifest (`api/src/npm/mod.rs` — "We don't publish yanked versions in the NPM manifest"), so npm clients resolving through the manifest won't find them. Worth a clause.
- **Email**: there are four transactional types, not three — the ticket-created acknowledgement is distinct from ticket replies (`api/src/api/tickets.rs`). "Acknowledgements and replies for support tickets you opened" covers it.
- **Download counts**: the 4h-granularity table is purged after 7 days, but per-version *daily* counts are kept forever. Calling those "aggregate totals" is a stretch; say "daily per-version counts are kept indefinitely."
- **Undisclosed retention**: we store identity-provider OAuth access and refresh tokens at rest (`github_identities` / `gitlab_identities`). That belongs in the collection or retention section.
- Verified and accurate as written: download counts store no IP (`lb/analytics.ts`), audit log schema is actor + time + action + meta with no IP or user agent, Turnstile is login-only, DB PITR window is 7 days, the Report Package button exists and goes to our own ticket system.

## Leftover Deploy boilerplate that contradicts the new text

- Privacy §II's purpose list includes "solicitation of your feedback … informing you about our organization, products, services, events, and other areas of interest" — contradicted by §VII.2 "We do not send marketing email."
- Privacy §II has *two* service-provider passages: the specific table ("this list is the complete set") and, a few paragraphs later, the generic "Like many businesses, we hire other companies … client survey and marketing, and data storage." Delete the generic one.
- ToS §6 bans decompiling/reverse-engineering "any software … accessible through the Services" — nonsensical for a registry that distributes source code and is itself MIT-licensed.
- ToS §2 says an account is required to use the JSR Offerings; downloading packages requires no account.

## Comparison with crates.io

crates.io has no terms of service at all — its legal surface is a community-drafted usage policy (adapted from PyPI's AUP), the Rust Foundation privacy notice, a data access page, and a security page. So the liability cap / indemnification / governing-law apparatus here has no crates.io counterpart; keeping it is a deliberate choice, not something to reconcile.

Where this draft is genuinely better than what crates.io publishes:

* Removal due process — notice, a stated reason, ~3 weeks to respond, and appeal. crates.io says the opposite ("content and account deletion is permanent and there is no basis to reverse these moderation actions").
* The narrowly-scoped hosting license that survives account deletion only to keep published versions available. crates.io has no license grant at all.
* Concrete retention numbers. The Rust Foundation notice retains server logs for 1 year and everything else "as long as necessary"; this draft is stricter and more transparent — once it's true.
* The plain-English passages (publication permanence, the leaked-credential paragraph, the trademark paragraph).

Worth adopting from crates.io in follow-ups:

* A **data access page** with concrete endpoints and limits — theirs documents the sparse index, the no-rate-limit CDN, RSS feeds, daily DB dumps, and API limits ("1 request per second, and an identifying user-agent"). Our only load language is "no unreasonable load, in our sole discretion"; hard numbers would serve mirror operators better and give us something objective to enforce.
* Content categories our usage policy lacks: obfuscated code, license-key circumvention, harassment/violent content, and sexually obscene content (with their artistic/educational carve-out). Those belong in `usage-policy.md`, not the ToS.

## Product change worth making alongside

The login page has no acceptance step and no links to these documents (`frontend/islands/LoginForm.tsx` is just the two provider buttons plus Turnstile), so acceptance is pure browsewrap. Add "By signing in you agree to the Terms of Service and Privacy Policy" with links on the login page, and put both links in the site footer. Cheap, and it strengthens everything else in the document.

## For counsel

* New York governing law and forum for a San Diego-based company is inherited from the Deploy terms — confirm it's deliberate.
* "Continued use constitutes acceptance" of privacy policy changes is weak under GDPR.
* The absence of a cookie/consent banner is defensible only while browser storage is strictly necessary — one more reason #1493 must be deployed before this takes effect, since the Algolia per-browser UUID is exactly the kind of storage that would otherwise require consent.
* The GDPR art. 17(3)(b)/(e) and CCPA §1798.105(d) citations for the one-year tombstone retention should be checked against #1331's final behavior.

Merge order: #1493 and #1331 first (merged *and deployed*), then this.

piscisaureus (changes requested):
Thanks for reworking these around a package registry rather than a serverless platform. I compared this head (`638e25bf`) with the current crates.io Usage/Data Access policies, the Rust Foundation privacy/copyright policies, current `main` (`9cbcf044`), and all open PRs.

Requesting changes because several concrete privacy promises are not true in production today, and the Terms still mix useful registry policy with generic Deploy/SaaS boilerplate.

One framing point: crates.io does not publish a conventional comprehensive ToS. It has a much shorter [Usage Policy](https://crates.io/policies), [Data Access Policy](https://crates.io/data-access), and the Foundation-wide [Privacy Policy](https://rustfoundation.org/policy/privacy-policy/). The Rust Foundation is an [independent nonprofit](https://rustfoundation.org/about/), not part of the Linux Foundation. Consequently, crates is a good model for registry usage/moderation, but it does not supply JSR with contract formation, warranty, liability, indemnity, governing-law, or user-content-license terms.

## Substantial policy differences

- **Acceptable use:** crates is shorter but substantially more complete and precise. It explicitly addresses phishing, fraud, obfuscation, privacy abuse/doxxing, harassment and threats, sexual exploitation, fake accounts, spam, cryptomining, package-name trading, Cargo/index abuse, and the boundary for legitimate security-research tools. These should be added to JSR's existing Usage Policy rather than embedded in the ToS.
- **Package removal:** JSR is stricter and more dependency-preserving. Publishers cannot delete a version after publication; they can yank it, while JSR reserves exceptional takedown authority. crates permits deletion in the first 72 hours and later under low-download/no-dependent conditions. I prefer JSR's model for ecosystem stability, but “cannot be deleted” needs to be qualified by the later operator-removal exceptions.
- **Moderation:** JSR's notice, reasons, response opportunity, human appeal, and possible restoration are clearer and fairer. crates preserves much more operator discretion and makes some removals permanent. The tradeoff is that JSR promises processes the software does not currently enforce.
- **Registry licensing and permanence:** JSR's explanation of public packages, immutable versions, yanking, caching, mirrors, npm tarballs, and account-deletion consequences is considerably better than Rust's broad notice. The operator license nevertheless needs review: “non-sublicensable” fits awkwardly with CDNs/mirrors, and the claim about ownership of all modifications/adaptations/translations looks like inherited boilerplate.
- **Generic service terms:** crates has no equivalents for JSR's $100 liability cap, broad publisher indemnity, NY forum, one-sided equitable relief, restrictive Deno-content rules, reverse-engineering prohibition, or unilateral no-notice changes. These are not inherently wrong merely because crates lacks them, but they need registry-specific legal review. The restrictive content/reverse-engineering provisions fit poorly with an MIT-licensed project, public API, generated documentation, and the later permission to link and mirror.
- **Copyright:** JSR provides an informal complaint and appeal path, while Rust has a separate [copyright policy](https://rustfoundation.org/policy/copyright-policy/). JSR's section is not a complete DMCA notice/counter-notice procedure. If Deno wants Section 512 safe-harbor treatment, the [Copyright Office guidance](https://www.copyright.gov/512/) describes designated-agent registration, notice, counter-notice, and restoration requirements.
- **Privacy:** JSR's JSR-specific inventory, provider table, retention schedule, and account-deletion explanation are much easier to understand than the broad Foundation policy. The problem is that the draft turns desirable intentions into absolute factual promises which current production contradicts.

There are also internal Terms conflicts:

- Credentials may be disabled “at any time for any reason,” while termination is later limited to breach/legal necessity and ordinarily requires notice.
- Versions “cannot be deleted,” followed by circumstances in which JSR may delete them.
- Anonymous browsing/downloading conflicts with “you must be an Authorized User.”
- The restrictive Deno “Content” clause conflicts with the open-source site, public API, mirrors, and broad logo permission.
- No-notice unilateral changes undermine the reliability of the detailed moderation promises.

## Where the language is better

- crates wins on acceptable-use taxonomy, ownership disputes, the security-research boundary, concision, and operational discretion.
- JSR wins on package visibility, immutability, yanking, caches/mirrors, account-deletion consequences, the provider-table format, and mapping data to retention/legal bases—provided those statements are verified.
- For moderation, keep JSR's notice/reason/appeal structure but make three weeks a service target rather than an unconditional commitment, with explicit legal, security, privacy, abuse, and infrastructure exceptions.
- For privacy rights and international transfers, combine JSR's specificity with Rust's identity-verification, alternate-legal-basis, US-processing, and government-access caveats.
- Neither policy handles under-16 users well: both allow supervised use while saying they do not knowingly collect child data, without explaining how consent is obtained.

If crates wording was adapted, its policy source is [CC BY 4.0](https://github.com/rust-lang/crates.io/blob/main/svelte/src/routes/policies/%2Bpage.svelte); please add attribution and identify modifications as required.

## Statements that do not match the implementation

### Algolia tracking

The draft says JSR does not tell Algolia which result was clicked and keeps no linkable identifier. Current code creates a persistent `algolia:userToken` and reports click/conversion events containing that token, query ID, object ID, and result position ([implementation](https://github.com/jsr-io/jsr/blob/9cbcf044bedb28fa5df35fc0e2bc4a7c01e93433/frontend/utils/algolia_insights.ts#L3-L129)).

[PR #1493](https://github.com/jsr-io/jsr/pull/1493) removes future Algolia Insights tracking and is approved/green. It should land before this policy, but it also needs a one-release cleanup of existing `algolia:userToken`/`algolia:pendingConversion` browser keys and a decision about historical Insights data held by Algolia.

### “Complete” browser-storage list

The sign-in cookie and Turnstile are not the complete list. JSR also uses local storage for `darkMode`, `preferredUsage`, cached GitHub/GitLab data, and currently Algolia identifiers; it also has OAuth-CSRF and staff-sudo cookies. #1493 removes only the Algolia entries. Please enumerate storage by purpose, or remove “In full” and “That is the complete list.” A more durable promise is “no advertising or cross-site behavioral tracking,” not “we do not track you.”

### “Complete” provider list

Production exports telemetry to Grafana Cloud ([production configuration](https://github.com/jsr-io/jsr/blob/9cbcf044bedb28fa5df35fc0e2bc4a7c01e93433/terraform/prod.tfvars#L8-L10)), which is absent from the provider table. HTTP spans include the full URI and user agent, and some instrumented spans include names, emails, avatars, and provider IDs. Cloudflare Workers observability destinations are dashboard-managed, so the list is also not entirely verifiable from public Terraform.

Please inventory the actual production data flows, disclose Grafana/observability retention, redact query strings and unnecessary PII from telemetry, and call the table something like “subprocessors used to operate JSR” rather than “the complete set of third parties that receive personal information.” The latter is also contradicted by the later generic vendors, complaint disclosures, public registry users, legal recipients, and corporate-successor provisions.

### Download retention

JSR sends package, version, and user-agent datapoints to Cloudflare Analytics Engine ([implementation](https://github.com/jsr-io/jsr/blob/9cbcf044bedb28fa5df35fc0e2bc4a7c01e93433/lb/analytics.ts#L5-L46)). Cloudflare documents [three-month retention](https://developers.cloudflare.com/analytics/analytics-engine/limits/). JSR's seven-day cleanup applies only to the imported four-hour database records, not those finer Cloudflare records. Either change the data flow or disclose the three-month raw retention. No open PR addresses this.

### OAuth-state retention

The cleanup deletes records older than an hour, but Terraform schedules it once daily, not hourly ([scheduler](https://github.com/jsr-io/jsr/blob/9cbcf044bedb28fa5df35fc0e2bc4a7c01e93433/terraform/scheduler.tf#L17-L29)). An expired record may remain for almost 25 hours. Change the cron to hourly or change the policy. No open PR addresses this.

### Account deletion

Current `main` does not provide self-service deletion; the settings page directs users to support ([current UI](https://github.com/jsr-io/jsr/blob/9cbcf044bedb28fa5df35fc0e2bc4a7c01e93433/frontend/routes/account/settings.tsx#L93-L103)).

[PR #1331](https://github.com/jsr-io/jsr/pull/1331) implements much of the described flow, but it should not land unchanged:

- It does not revoke/delete the GitHub/GitLab identity records containing OAuth access and refresh tokens. Those identity tables are parents of `users`, so deleting a user does not cascade to them ([GitHub schema](https://github.com/jsr-io/jsr/blob/9cbcf044bedb28fa5df35fc0e2bc4a7c01e93433/api/migrations/20230728143939_github_identities.sql#L1-L12), [GitLab schema](https://github.com/jsr-io/jsr/blob/9cbcf044bedb28fa5df35fc0e2bc4a7c01e93433/api/migrations/20250630231604_gitlab_identities.sql#L1-L12)). These credentials are not disclosed in the policy at all.
- It supports deletion holds, contradicting “at any time,” and checks the hold outside the deletion transaction, leaving a race.
- It does not create an anonymous placeholder account; the UI simply omits publisher attribution.
- The policy says retained identifiers are in a “single entry,” but the PR adds `original_actor_id` to every historical audit entry.
- Tombstone name/email/provider IDs are visible to authorized staff through the admin audit UI, so “not shown to anyone” is false.
- A daily anonymization job means the identifying fields may remain approximately 365–366 days.
- It needs the timestamp mapping fix from [#1345](https://github.com/jsr-io/jsr/pull/1345).

Please fix those behaviors or describe them accurately, and add tests covering provider-token deletion/revocation, audit attribution, holds, scope transfer, and anonymization.

### Moderation and restoration

Reporting exists, and publisher immutability/yanking are substantially implemented. However, staff deletion does not require a notice, reason, deadline, second reviewer, or appeal state. Removal hard-deletes database and stored objects; the normal publication path prevents republishing a previously completed version. Thus the promised three-week process and restoration are not technically enforced or supported by a tested restore path. No open PR addresses this.

If these promises are important, model moderation cases explicitly: report, notice, response deadline, evidence, reviewer, decision, emergency override, appeal, and restoration. Otherwise state them as operational goals rather than guarantees.

### Verified or mostly verified items

- The Google logging bucket has the promised [30-day retention](https://github.com/jsr-io/jsr/blob/9cbcf044bedb28fa5df35fc0e2bc4a7c01e93433/terraform/logging.tf#L2-L10), but Grafana/Cloudflare telemetry retention remains outside that promise.
- Database point-in-time transaction logs are configured for [seven days](https://github.com/jsr-io/jsr/blob/9cbcf044bedb28fa5df35fc0e2bc4a7c01e93433/terraform/db.tf#L40-L44); confirm that this accurately describes the complete backup/recovery behavior.
- Publisher version immutability, yanking, required package licenses, Turnstile, and the absence of marketing/newsletter code are substantially implemented.
- Access, correction, restriction, objection, portable export, SCC/vendor contracts, and child-data requests may be manual operational processes, but cannot be verified from the repository. Document internal procedures before making unqualified promises.

Relevant additional PRs: [#1486](https://github.com/jsr-io/jsr/pull/1486) improves admin-data access controls but currently conflicts; [#1270](https://github.com/jsr-io/jsr/pull/1270) would add private packages and make the all-public wording stale; [#1243](https://github.com/jsr-io/jsr/pull/1243) would require webhook-data and 30-day retention disclosures; [#1470](https://github.com/jsr-io/jsr/pull/1470) could change the API provider description.

## Recommended shape before merge

1. Land #1493 plus browser-key/historical-data cleanup.
2. Correct #1331, including OAuth credential revocation/deletion and accurate hold/tombstone/attribution language, then land it before promising self-service deletion.
3. Correct the OAuth schedule and Cloudflare/Grafana disclosures or change the implementation.
4. Split this into a short counsel-reviewed service contract, the registry-specific Usage/Moderation Policy, and a factual privacy/data-retention notice.
5. Narrow aggregate-data use to security, fraud/abuse and supply-chain investigation, reliability, capacity planning, ecosystem research, and genuinely deidentified statistics—not unrestricted “lawful business purposes.”
6. Remove generic customer/events/marketing/survey language and the remaining Deploy-oriented reverse-engineering/content/equitable-relief boilerplate unless counsel identifies a concrete need.
7. Add an effective date/version history and material-change notice. If contractual assent matters, record the version accepted during signup/publishing rather than relying only on continued browsing.
8. Have privacy/product counsel review the categorical GDPR Article 17 and CCPA deletion-exception claims, SCCs/vendor contracts, under-16 consent, international-representative obligations, DMCA procedure, indemnity, forum, and liability language. These legal exceptions are contextual and should not be asserted categorically based only on implementation comments.

The registry-specific parts are a strong improvement. The remaining work is mostly to separate the useful package-registry commitments from inherited service boilerplate and ensure every factual privacy promise matches production before publication.

piscisaureus (commented):
## Supplemental comparison: npm Open Source Terms

The substantive npm comparator is the [npm Open Source Terms](https://docs.npmjs.com/policies/open-source-terms/), together with the policies it incorporates. This comparison strengthens the recommendation above, with one refinement: **npm is a better structural model than crates.io for JSR's short legal contract, while crates.io remains the better model for the usage/moderation policy.**

npm's free-registry terms are a roughly 3,450-word full contract. They cover contract formation, use of the registry, account responsibility, the package-content license, indemnity, warranty disclaimers, a liability cap, termination, governing law, and disputes. The proposed JSR Terms cover the same category of issues in roughly 4,400 words. Unlike crates, this is an apples-to-apples legal instrument.

npm then keeps operational subjects in separate documents: [Unpublish Policy](https://docs.npmjs.com/policies/unpublish/), [package/name disputes](https://docs.npmjs.com/policies/disputes/), [Code of Conduct](https://docs.npmjs.com/policies/conduct/), [DMCA](https://docs.npmjs.com/policies/dmca/), [crawler/replication policy](https://docs.npmjs.com/policies/crawlers/), and the new [Dual-Use Content Policy](https://docs.npmjs.com/policies/dual-use/). That separation is a particularly good model for JSR.

### Where npm is better

1. **Plain, registry-specific access grant.** npm says users may search, download, publish, and manage packages through its CLI, documented public APIs, and website. This is clearer than JSR's generic “limited, non-transferable, non-sublicensable, revocable” service license and avoids the draft's conflict between anonymous use and the requirement to be an Authorized User. JSR should distinguish anonymous browsing/downloading, authenticated publishing/management, documented APIs, and mirrors.

2. **Policy changes include notice.** npm posts a new effective date and says account holders will be notified by email, a website message, or both before continued use constitutes acceptance ([Changes](https://docs.npmjs.com/policies/open-source-terms/#changes)). JSR reserves changes without prior notice. npm's wording is materially better, although recording the accepted version on signup/publish would be stronger than either document.

3. **Package-analysis permission is explicit and bounded.** npm's [Your Content](https://docs.npmjs.com/policies/open-source-terms/#your-content) license permits it to copy, publish, and analyze packages, share analyses, and run package code for analysis—but not to use that code as product functionality. This directly addresses malware detection, abuse, supply-chain research, compatibility analysis, and vulnerability work. JSR currently has a narrow operating/transformations license plus an unrelated, overly broad Aggregated Data clause. A JSR-specific clause should explicitly permit static/dynamic package analysis and execution solely for security, abuse prevention, registry integrity, compatibility, documentation, and supply-chain research, with carefully limited sharing of advisories and deidentified results. That would be much clearer than “otherwise used and disclosed for lawful business purposes.”

4. **License duration follows retained copies.** npm's special operator license ends when the last copy disappears from its backups, caches, and systems, while independent open-source licenses may continue. JSR says its license survives deletion and termination without a similarly clear endpoint. Because JSR deliberately preserves versions, its license may last indefinitely in practice, but the drafting should say it lasts only while content remains stored or served under the registry's permanence/removal policy.

5. **Indemnity is narrower.** npm limits indemnity to third-party proceedings alleging that the user's content or use violates third-party IP, the agreement, or applicable law. JSR additionally reaches broad negligence, gross-negligence, fraud, and other warranty theories. npm is the better ceiling for a free community registry; counsel may be able to narrow JSR further.

6. **Liability drafting has an important legal carveout.** npm caps direct damages at $50 rather than JSR's $100, but the number is not the interesting difference. npm expressly says some warranty exclusions and damages limitations may not apply in jurisdictions that prohibit them ([Limits on Liability](https://docs.npmjs.com/policies/open-source-terms/#limits-on-liability)). JSR should include mandatory-law and consumer-rights carveouts and a severability clause.

7. **Package-risk disclaimer is more concrete.** npm plainly says publisher-provided ownership and licensing metadata may be wrong and that npm is not providing legal advice. JSR should add that specific warning rather than relying only on generic third-party-content disclaimers.

8. **Name disputes are separated from general moderation.** npm's [name policy](https://docs.npmjs.com/policies/disputes/) states first-come/active-use rules, prohibits buying and selling names, says it will not transfer a name merely because somebody else wants it, and routes trademark claims through a formal process. JSR's existing Usage Policy is already close, but “always” contacting an owner conflicts with the Terms' emergency exceptions. Use a separate policy with a normal notice practice and explicit exceptions.

9. **Actual DMCA policy.** npm delegates to GitHub's full notice/counter-notice process and identifies a DMCA agent and submission channels. This reinforces the earlier recommendation that JSR's informal copyright paragraph should become a separate, counsel-reviewed DMCA policy if Deno intends to seek Section 512 safe-harbor treatment.

### Where JSR is better—and should not copy npm

1. **Due process.** npm may remove content without notice merely on suspicion and may permanently expel users without warning, although its Code of Conduct now links to an appeal form. JSR's reason, notice, response, human review, and possible restoration approach is better for a community registry. It just needs realistic qualifiers and implementation.

2. **Dependency stability.** npm allows unpublishing within 72 hours if there are no dependents, or later when there are no registry dependents, fewer than 300 weekly downloads, and a single owner ([Unpublish Policy](https://docs.npmjs.com/policies/unpublish/)). JSR's publisher-immutable/yank-only model better protects downstream builds. I would retain it.

3. **No arbitration/class waiver.** npm's 2022 terms require individual arbitration and waive class proceedings. The current JSR head correctly removed those clauses; npm is not a reason to restore them.

4. **Open security data.** npm restricts vulnerability/security data to personal or internal-business use and prohibits providing it to others or incorporating it into products. That is a poor fit for JSR's open, noncommercial ecosystem and should not be copied.

5. **Open registry access.** npm prohibits automated website access and imposes strong crawler restrictions, then offers registry replication as the approved path. JSR is more permissive about mirrors and proxies. Keep that openness, but add a concrete Data Access/Mirroring Policy documenting supported endpoints, dumps or feeds, rate limits, attribution, deletions, and security advisories.

6. **Privacy minimalism.** npm collects substantially more registry telemetry: a per-command session identifier, dependency graphs, Node/npm/OS versions, CI and scope headers, referrer, IP, URL, and user agent; it also uses data for recommendations, product announcements, sales, and marketing ([npm Privacy](https://docs.npmjs.com/policies/privacy/)). JSR's intended no-advertising/no-marketing posture and short retention periods are materially better. Do not use npm's privacy notice as a data-collection template.

### Two npm mechanisms worth adopting

**Removal tombstones and replication notices.** npm explains that some erased packages receive placeholder versions describing the erasure, and its registry APIs send update messages to replicators. That is a better technical foundation than JSR's present hard deletion plus an unimplemented promise to restore. For JSR, removal should normally create a durable tombstone identifying the affected scope/package/version, status, non-sensitive reason category, time, and appeal outcome, while caches/mirrors receive a machine-readable removal update. This also preserves version non-reuse and gives downstream users an intelligible failure.

**Declared dual-use packages.** npm's policy, updated July 28, 2026, allows legitimate penetration-testing, security-research, and obfuscation tools but requires persistent `contentPolicy: { class: "dual-use" }` metadata plus a `DISCLOSURE` file. It also applies stronger publishing authentication and additional scanning ([Dual-Use Content Policy](https://docs.npmjs.com/policies/dual-use/)). This is a thoughtful supply-chain control that goes beyond both the JSR and crates text. JSR should consider a lighter version—declaration plus disclosure and risk-appropriate review/scanning—but only after the metadata, UI, authentication, and enforcement path exists. It should not be promised purely in policy.

### Privacy wording comparison

npm's privacy notice is not a good overall model: it took effect in 2020, has no useful retention schedule, names provider categories rather than actual subprocessors, and reflects paid/commercial processing that JSR does not perform. JSR's data inventory and proposed retention table are much better once corrected.

npm does have three pieces of wording worth borrowing:

- its opening immediately warns that public packages may contain personal information and that publishers must consider what others can do with it;
- it explains that account closure does not erase packages, that npm cannot recall copies from downloaders, and that open-source licenses may require permanent attribution;
- its GDPR/CCPA rights sections say npm may verify identity and may deny or narrow a request where another legal basis applies, with an explanation to the requester.

npm is also clearer on children: its service is intended for users 16 and older, and inadvertently collected under-16 data is deleted. JSR currently permits under-16 use with parental consent but implements no age or guardian-consent workflow. Unless that workflow will be built, a straightforward 16+ rule is clearer and more truthful.

### Revised recommendation

Use **npm Open Source Terms as the structural starting point for a shorter JSR contract**, then remove npm-specific commercial restrictions, paid-service assumptions, arbitration, class waiver, security-data restrictions, and crawler restrictions. Use **crates.io for the registry Usage/Moderation Policy**, and use JSR's own verified production inventory for privacy.

Concretely, the ideal document set is:

1. short JSR Open Registry Terms, modeled structurally on npm;
2. JSR Usage/Moderation Policy, modeled substantively on crates with JSR's stronger notice/appeal goals;
3. Package Removal and Name Dispute policies, incorporating npm's tombstone/replication approach;
4. a separate DMCA policy;
5. a supported Data Access/Mirroring Policy; and
6. a factual JSR Privacy Notice derived from production configuration rather than any other service's boilerplate.

This comparison does not change the prior request for changes; it provides a better replacement architecture for the Terms rather than merely trimming the current Deploy-derived draft.

piscisaureus:
Follow-up to my review: I also compared this draft against npm's policy suite — the [Open-Source Terms](https://docs.npmjs.com/policies/open-source-terms) (the actual ToS), the [Privacy Q&A](https://docs.npmjs.com/policies/privacy), and the [Unpublish Policy](https://docs.npmjs.com/policies/unpublish). A few things there are worth folding into this PR.

### Worth adopting from npm

1. **Change notification.** npm commits to notifying account holders of term changes "by email to the address provided for your Account, by a message on the Website, or both," and only falls back to check-the-page for anonymous users. Our draft has the pure "check the page yourself; continued use is acceptance; no prior notice" clause, which I already flagged as the weakest part of the document. We have transactional email; adopting npm's commitment costs almost nothing and materially strengthens enforceability of future changes.

2. **A content-analysis license clause.** npm's license lets it "copy, publish, and *analyze* Your Content, and to share its analyses with others," including running code in a package to analyze it — while expressly not gaining the right to run that code for product functionality. Our draft's license covers store/reproduce/distribute/generate docs, search indexes, and npm tarballs "solely in order to operate the Services." Malware scanning, static/dynamic analysis, and publishing security findings about a package are arguably outside that grant, and the Aggregate Data clause doesn't cover them either (an analysis of one malicious package is neither anonymous nor aggregated). Since combating abuse and supply-chain attacks is exactly the kind of analysis we want to be able to do, we should add an npm-style analyze clause.

3. **A concrete abuse threshold.** npm bans website scraping but explicitly permits registry replication via its public APIs, then draws a bright line: "under no circumstances are five million requests … in a single month-long period … remotely reasonable." crates.io's data access page similarly gives hard numbers (1 request/second plus an identifying user-agent). Our only load language is "unreasonable load, in our sole discretion." This reinforces the data-access-page suggestion from my review — mirror operators and tooling authors get a number, and we get something objective to enforce.

4. **Say what happens after an exceptional removal.** npm publishes placeholder versions with explanatory READMEs for erased packages and sends deletion messages through its replication APIs. Our draft says nothing about what mirrors and dependents see after a staff removal — and today the staff version-delete endpoint doesn't even clean up the npm-compat tarball. If we're keeping removal as a staff-only exceptional action, a sentence about tombstone/notification behavior (and matching implementation) would complete the story.

5. **Consider an ad-content rule.** npm's "Commercial Content" section allows funding links, sponsor logos, and paid-product info, but bans ads in READMEs and packages that display ads at install/run time. Neither our usage policy nor crates.io's has this. Given the ecosystem's history with install-time ads, one sentence in `usage-policy.md` is cheap insurance.

### For counsel

- **Arbitration.** npm kept binding AAA arbitration plus a class-action waiver (California law, $50 cap). This draft dropped arbitration in favor of NY courts, so it now has neither arbitration nor a class waiver. That should be an explicit decision, not an artifact of editing.
- **"True, accurate, current, and complete information."** npm requires only a working email address and expressly allows pseudonyms. Our inherited clause sits oddly with GitHub/GitLab-based auth, where pseudonymous accounts are normal and fine. Suggest softening or dropping it.
- Trivia: npm's own docs are internally inconsistent on minimum age (13 in the terms, 16 in the privacy doc). We're consistently 16 — keep it that way.

### Style, take or leave

npm's Open-Source Terms are the best-written registry ToS out there: short declarative sentences and a bolded "Important Terms" summary up front pointing at the three clauses that actually bite (disclaimers, liability limit, arbitration). Our best passages are already at that level; a similar callout at the top of the ToS would help. npm's privacy doc uses a Q&A structure ("What data does npm collect about me, and why?") that is notably easier to navigate than traditional numbered sections — an option if we ever restructure.

### Where our draft already beats npm — no action, just context

The privacy comparison is lopsided in our favor. npm collects a per-CLI-run session identifier, your project's full dependency tree, CI markers and referrer headers, uses first-party analytics cookies, uses your email for product announcements and "messages about paid services you might want," and states no retention period for anything. Account closure is "contact support … we may retain some data about you internally." Against that, this draft's no-analytics/no-marketing stance, concrete retention windows, and fully specified deletion procedure put JSR well ahead of both npm and crates.io — once the implementation gaps from my review are closed, the document can honestly claim that.

piscisaureus:
Consolidating the four review comments above into one concrete, prioritized list. Where they overlap this supersedes them; nothing here adds new findings except the forum and acceptance recommendations at the end.

## A. Merge sequencing (gates for this PR)

1. **Land #1493 first**, extended with: a one-release cleanup that removes the existing `algolia:userToken` (localStorage) and `algolia:pendingConversion` (sessionStorage) keys from returning browsers, and a decision on the historical Insights data Algolia already holds (purge request, or disclose its retention).
2. **Fix and land #1331 second.** Required fixes before it merges:
   - Delete (and ideally revoke upstream) the `github_identities` / `gitlab_identities` rows on account deletion — they hold OAuth access and refresh tokens, are parents of `users`, and do not cascade today.
   - Move the `deletion_hold` check inside the deletion transaction (current check races).
   - Incorporate the timestamp-mapping fix from #1345.
3. **This PR takes effect only after both are deployed.** Several §V/§VI claims cite GDPR/CCPA provisions for behavior that must exist in production first.

## B. Code changes with no open PR

4. **Sign-in state:** change `clean-oauth-states` from daily to hourly, and add cleanup jobs for expired `tokens` and `authorizations` rows (currently never deleted, only filtered at read time). Otherwise soften the policy claim.
5. **Telemetry:** stop exporting `user.email` / `user.name` / avatar as span attributes (the `#[instrument]` fields on user create/update); consider redacting query strings from `http.url`; pin Grafana Cloud retention; then add Grafana to the provider table.
6. **Download counts:** the raw datapoints (including user-agent) live in Cloudflare Analytics Engine for ~3 months per Cloudflare's documented limits; the 7-day cleanup only covers the imported Postgres rows. Either change the pipeline or disclose the 3-month raw retention.
7. **Acceptance capture:** add a conspicuous "By signing in you agree to the [Terms of Service] and [Privacy Policy]" line on the login page, record the accepted ToS version + timestamp at account creation, and put both links in the site footer. (See section E.)
8. **If we keep the removal-process promises:** model moderation cases explicitly (notice, reason, response deadline, appeal state, tested restore path), emit machine-readable removal tombstones for mirrors/replicators (npm-style), and make staff version-deletes clean up the npm-compat tarball. Otherwise reword the promises as normal practice with express exceptions rather than guarantees.

## C. Text corrections in this PR

9. **Account deletion (§VI):** describe actual behavior — attribution is removed, not replaced by an "anonymous placeholder"; sole-member scopes transfer to the visible, named JSR service account; the retained identifiers are in the tombstone *and* as `original_actor_id` on historical audit entries (not "a single entry"); "not shown to anyone" → "not publicly visible" (staff can see the tombstone via the admin audit UI); retention is ~365–366 days given the daily job; disclose OAuth token storage and their deletion (per A.2).
10. **Browser storage (§VII.3):** enumerate by purpose — `token`, `oauth_login_csrf`, staff `sudo`, `darkMode`, `preferredUsage`, `gh-login-*`/`gl-username-*` caches, Turnstile, Cloudflare edge cookies — or drop "that is the complete list." Replace "we do not track you" with the durable version: "no advertising and no cross-site behavioral tracking."
11. **Provider table (§II):** retitle to "subprocessors used to operate JSR," add Grafana Cloud, drop the "complete set of third parties" claim (contradicted by legal disclosures, complaint forwarding, corporate-successor clauses), and delete the duplicate generic "client survey and marketing" vendors paragraph plus the "informing you about our organization, products, services, events" purpose language.
12. **Immutability/yank:** "contents cannot be modified; versions cannot be deleted by users; JSR staff may remove a version in exceptional cases" (matches code and `immutability.md`); note yanked versions are excluded from the npm-compat manifest though pinned tarball URLs keep working.
13. **Email:** four transactional types — "acknowledgements and replies for support tickets you opened" plus scope invites and token notices.
14. **ToS internal conflicts:** delete the reverse-engineering / proprietary-markings / safety-features bullets from §6; fix §2 so anonymous browsing/downloading doesn't require an "Authorized User" (npm's access-grant structure — anonymous read, authenticated publish/manage, documented APIs, mirrors — is the model); reconcile "disable any Account Credentials at any time for any reason" with §16's notice-before-termination promise; make "around three weeks" a normal-practice target with the stated exceptions.
15. **License (§5):** reconcile "non-sublicensable" with CDN/mirror distribution; drop the inherited "you own all modifications, improvements, adaptations" boilerplate; tie the license's endpoint to content remaining stored or served under the permanence/removal policy; **add an npm-style analysis clause** (copy, analyze, and run package code solely for security, abuse prevention, registry integrity, compatibility, and supply-chain research, with sharing of advisories and deidentified results) — this is the clause that actually licenses malware scanning and abuse analysis, which the current "solely in order to operate the Services" grant and the over-broad "lawful business purposes" Aggregate Data clause both miss. Narrow the Aggregate Data purposes to match.
16. **Age:** adopt a plain 16+ rule (npm-style: under-16 data deleted on discovery) instead of "a parent or legal guardian must consent and supervise," since no consent workflow exists or is planned.
17. **Liability/indemnity:** add the mandatory-law carve-out ("some jurisdictions do not allow…") and a severability clause; consider narrowing indemnity to npm's scope (third-party claims that your content or use infringes/violates), dropping the negligence/fraud prongs.
18. **Housekeeping:** attribute crates.io (CC BY 4.0) if any wording was adapted; keep a version history and effective-date practice; adopt npm's change-notification commitment (email or site message to account holders before continued use counts as acceptance). Watch-list for future drift: #1270 (private packages would make "everything you publish is public" stale), #1243 (webhooks would add a subprocessor/retention disclosure), #1470 (API hosting description).
19. **Structure (recommended, not blocking):** split into the document set both comparisons point at — a short registry ToS (npm's Open-Source Terms as the structural model, minus arbitration, crawler bans, and security-data restrictions), the Usage/Moderation Policy (crates.io as the substantive model, keeping our stronger notice/appeal goals), a removal/name-dispute policy with the tombstone mechanism, a separate counsel-reviewed DMCA policy with a registered designated agent if we want §512 safe-harbor treatment, a Data Access/Mirroring page with concrete endpoints and rate limits, and the privacy notice derived from the verified production inventory.

## D. Governing law and forum

20. **Change the forum from New York to California, if we keep a forum clause at all.** NY is inherited from the Deploy terms and has no connection to Deno Land (San Diego), the infrastructure, or the users; a forum with no nexus to either party is also the flavor most likely to draw enforceability challenges. California (San Diego, or npm's San Francisco pattern) is defensible and cheaper for us to litigate in. Note §17 (equitable relief) contains its own consent to New York jurisdiction — it must change consistently with §18, or be folded into it.

## E. Click-through acceptance: recommended, and cheap

npm captures assent at account creation; crates.io doesn't bother. Both are rational: **crates.io has almost no contract to enforce** — no liability cap, no indemnity, no forum clause — so nothing turns on whether a user "agreed." This draft, like npm's, is a real contract, and its most valuable clauses (the $100 cap, indemnity, forum, the content license) only help if a court finds the user assented. US courts routinely decline to enforce pure browsewrap and routinely enforce clickwrap and sign-in-wrap where the notice is conspicuous at signup. Having chosen to ship a ToS with teeth, we should capture the assent that makes the teeth real — otherwise the cap and indemnity are decorative against exactly the users they exist for.

Concretely, in increasing order of strength:

- **Sign-in-wrap (recommended minimum):** conspicuous "By signing in you agree…" adjacent to the provider buttons, plus recording the accepted ToS version and timestamp at account creation. Near-zero friction; generally held sufficient when the notice is clear. The CLI/device flow already routes through the website login, so this covers publishers too.
- **Checkbox clickwrap:** strongest evidence, slightly more friction; probably unnecessary for a registry, counsel's call.
- **On material changes:** notify account holders (email or site banner, per the npm-style change clause in C.18) and record re-acceptance on next login rather than relying on continued browsing.

Anonymous downloaders can't assent under any of these — true for npm as well. For them the disclaimers and the packages' own licenses do the work; the contract matters chiefly for account holders, which is exactly where assent is capturable.

piscisaureus:
Two refinements to the consolidated recommendations above, after checking the California/Ninth Circuit authorities.

## Governing law and forum

I agree that New York should be removed. Deno Land's stated address is in San Diego, so California has the obvious relationship to the operator and is the more natural governing law. California courts generally favor forum clauses where the selection is voluntary, has a logical nexus to a party or dispute, and does not impair mandatory consumer rights ([*America Online v. Superior Court*](https://law.justia.com/cases/california/court-of-appeal/4th/90/1.html)).

Concrete recommendation:

- Use California governing law, subject to non-waivable protections under the law applicable to the user.
- Decide with counsel whether an **exclusive** forum is useful at all. For a free registry, governing-law certainty may be more valuable than forcing every dispute into one courthouse.
- If an exclusive forum is retained, specify the state courts in San Diego County and the United States District Court for the Southern District of California, with a carveout wherever applicable law gives a user the right to bring a claim elsewhere.
- Remove or update every separate New York reference, especially the consent-to-jurisdiction language in the equitable-relief section. Prefer folding equitable relief into the single disputes section rather than maintaining two forum provisions.
- Do not restore npm's arbitration or class-action waiver merely because npm has them; that should remain a separate, explicit counsel/business decision.

## Affirmative acceptance

An acceptance step is not necessary merely to publish a privacy notice or operate a public registry. It **is necessary in practical terms if Deno expects the contractual provisions in this draft—the content license, indemnity, liability cap, termination rules, governing law, and forum—to be reliably enforceable against account holders and publishers.** That explains the npm/crates difference: npm has a full contract worth capturing assent to; crates.io largely publishes operational policies and has no comparable indemnity, cap, or forum clause.

The current proposal is browsewrap: it says use constitutes acceptance, but the login page neither presents the terms nor asks for assent. The Ninth Circuit rejected terms where a site merely exposed a hyperlink without affirmative acknowledgment in [*Nguyen v. Barnes & Noble*](https://cdn.ca9.uscourts.gov/datastore/opinions/2014/08/18/12-56628.pdf). It later explained that online formation requires reasonably conspicuous notice plus an action that unambiguously manifests assent, and noted that courts routinely enforce checkbox clickwrap ([*Berman v. Freedom Financial Network*](https://cdn.ca9.uscourts.gov/datastore/opinions/2022/04/05/20-16900.pdf)). California's *Sellers* decision also shows that inconspicuous fine-print sign-in-wrap can fail even when text nominally says clicking agrees to terms.

I therefore recommend **actual checkbox clickwrap**, not merely a footer link and not fine-print sign-in-wrap:

1. At first account creation/sign-in, before enabling account writes or issuing publishing credentials, require an unchecked box adjacent to conspicuous links: **“I agree to the JSR Terms of Service and acknowledge the JSR Privacy Notice.”** “Acknowledge” is deliberate: the privacy notice describes processing; it should not imply that blanket contractual consent is the legal basis for all processing.
2. Store the account ID, Terms version/effective date (or immutable document hash), and acceptance timestamp. The repository should retain each historical version so the accepted text can be reconstructed. An IP address is unnecessary for this evidence.
3. Existing account holders should accept before their next authenticated write—at minimum before publishing, creating a token, or changing scope membership. This covers existing API-token users who might otherwise never revisit the OAuth login screen.
4. For material future changes, notify account holders by email or a conspicuous site notice and require re-acceptance before the next authenticated write. Merely changing the page and expecting users to check it is weak; the Ninth Circuit held that users have no obligation to hunt periodically for modified terms in [*Stover v. Experian*](https://cdn.ca9.uscourts.gov/datastore/opinions/2020/10/21/19-55204.pdf).
5. Do not force anonymous readers/downloaders through a click. They can use the public registry without an account, but Deno should not assume the contractual indemnity, forum, or liability cap is enforceable against them. Package licenses, statutory rights, and non-contractual notices still apply.

A conspicuous “By signing in…” notice may sometimes be enforceable, but because an unchecked checkbox is cheap here and produces substantially cleaner evidence, I would use the checkbox. If JSR does not want any acceptance flow, the alternative is to simplify the document into policy/notices and stop relying on assent-dependent protections; keeping a full npm-style contract while using crates-style no-assent presentation is the weakest combination.

piscisaureus:
From a real human:
* The recommendations make the policies overly specific. E.g. I do not want to commit to an exact blocking / appeals process. I do not want to guarantee any specific timeline or that people will get a response at all if they e.g. appeal to our decisions. We might have ~0.01 humans working on moderation at any point in time. It should leave us some leeway to organize our business processes.
* It is also a free service - we strive to make it a useful package registry for javascript users but we **do not guarantee** it will be available in general or to anyone specifically. The purpose of our policies:
  - Document our intent - a useful javascript package registry that users can rely on
  - Document that we can't accept liability if we don't succeed at that.
  - Explain how we balance the interests of package users vs the interests of package authors
* It is also overly specific on a technical front - It should allow for some evolution of the service. We may not use a dark mode cookie next week. We could use datadog and not grafana next month. The OAuth refresh tokens are a technical detail that does not belong in a TOS or privacy policy IMO - just document in more generic terms what we store and why.
* Some guarantees I think we can safely make - we collect only the data we need to operate the service. We do not sell/share your data with third parties (with a change of control exception). We may analyze your data (which is mostly packages you make publicly available plus very limited PII we get from oauth providers plus whatever you provide us). When you delete your account, PII gets anonymized/deleted after some retention period to allow us to combat abuse (such as supply chain attacks).

piscisaureus:
Reconciling the guidance above with the two refinement comments, and revising the consolidated list accordingly. Short version: adopt the California/forum refinement wholesale; adopt checkbox clickwrap with one change to the rollout mechanics; and the over-specificity point is right — it invalidates a meaningful chunk of the earlier consolidated list, including several of its own items, and shrinks what is actually blocking.

## Adopted: governing law and forum

The refinement is right on every point: California governing law (subject to non-waivable protections of the user's home law); question with counsel whether an *exclusive* forum earns its keep at all for a free registry; San Diego County / S.D. Cal. if one is kept, with a carveout where law gives the user the right to sue elsewhere; remove every New York reference including the equitable-relief consent and fold equitable relief into a single disputes section; and do not restore arbitration or a class waiver merely because npm has them. Nothing to push back on.

## Adopted with one change: acceptance

The checkbox recommendation is right, and its closing framing is the correct way to put the decision: a full npm-style contract presented crates-style with no assent is the weakest combination — either capture assent or simplify until nothing depends on it. Since we are keeping the liability cap, indemnity, and forum clauses, assent it is; and an unchecked checkbox on what is already a dedicated login page with two buttons is negligible friction for materially cleaner evidence than fine-print sign-in-wrap. Keep the mechanics: record account ID, terms version, and timestamp (no IP needed); re-acceptance on material changes with actual notice; never gate anonymous reads.

One pushback, on point 3: requiring existing account holders to accept "before their next authenticated write" means intercepting the entire API surface, including CLI token flows — significant machinery for marginal evidentiary gain, and exactly the kind of process commitment the note above says to avoid building. Simpler: require acceptance at next *web* sign-in and before creating new tokens; existing CLI tokens keep working. Whether publishing with a pre-existing token also needs gating is a counsel question — don't build it preemptively.

## The over-specificity point is right — and it cuts deeper than trimming

It doesn't just soften wording; it removes some blockers entirely, because several "implementation gaps" were only gaps relative to over-specific text.

**Moderation and appeals.** The draft currently promises pre-removal notice with "ordinarily around three weeks" to respond (stated three times), "we review everything that is reported," "a person will look at it again," and restoration when a response resolves the matter. With ~0.01 humans on moderation those are promises the service cannot keep, and both earlier reviews reacted by recommending workflow machinery — case states, deadlines, appeal tracking, tested restore paths. The better fix is the crates.io posture: reserve full discretion, state intent, promise nothing. Something like: "We may remove content and suspend accounts at our discretion. Where practical we will notify the publisher and give them an opportunity to respond. You may contest a decision at help@jsr.io. JSR is a free service and we cannot guarantee a response or any particular timeline." That deletes the three-week clause, the everything-gets-reviewed clause, and the entire build-moderation-tooling recommendation. (The removal-tombstone idea survives as good engineering for mirrors, decoupled from any policy promise.)

**Technical specifics in the privacy policy.** Agreed: per-key cookie lists, a "complete" named vendor table, and OAuth refresh-token rows freeze implementation details into a legal document and turn every frontend tweak into a policy amendment. Replace with categories plus durable guarantees:

- *Browser storage:* "cookies and local storage only as strictly necessary to operate the site and to remember your preferences; never for advertising or cross-site tracking." No enumeration. This also keeps the no-consent-banner position stable across UI changes.
- *Recipients:* categories — hosting and infrastructure, content delivery, search, email delivery, monitoring and observability, identity providers. If we still want the verifiability gesture, a plain non-legal docs page can list current vendors and change freely. This absorbs the Grafana-vs-Datadog point: the category "monitoring and observability" belongs in the policy; the vendor name doesn't.
- *Credentials:* "identifiers and credentials needed to authenticate you with your identity provider" covers token storage generically.
- *Retention:* keep round numbers only where they're cheap and stable ("logs for about 30 days," "about one year after account deletion"), and drop the ones that manufacture falsity — "hourly," "that is the complete list," exact fine-grained-count windows. "Expired sign-in state is deleted automatically" is enough, which also downgrades the hourly-cron change from blocking to hygiene, and dissolves the Cloudflare Analytics Engine disclosure question.

**The four guarantees in the note are the right core, and they're already nearly true:** we collect only the data we need to operate; we do not sell or share it with third parties, except providers acting on our behalf and a change-of-control; we may analyze the data (which pairs exactly with the npm-style analysis license recommended earlier — that clause is what makes "we may analyze your data" contractually clean); and on account deletion PII is deleted or anonymized after a retention period that lets us combat abuse and supply-chain attacks. Add "no marketing email" and "no advertising or cross-site tracking" and that is the complete promise set. Everything else in the privacy policy is description, not commitment, and should read as such.

## What is still genuinely blocking

De-specification shrinks the gate list to items where even the *generic* promises fail:

1. **#1493 (Algolia)** — "no cross-site tracking" and "only the data we need" are false while a persistent per-browser identifier reports clicks to a third party. Land it, plus the cleanup of already-stored browser keys.
2. **#1331 (account deletion)** — "you can delete your account" and the retention-period promise depend on it. It should also delete the identity rows holding provider OAuth tokens (a "keep only what we need" failure even if the policy never names them) and fix the deletion-hold race.
3. **Telemetry span attributes** — exporting user emails and names to the observability vendor fails "only the data we need to operate" regardless of which vendor it is. Small code fix.
4. **The text pass on this PR** — the de-specification rewrite above, plus the previously flagged internal contradictions, which all survive: marketing-purposes language, the duplicate generic vendors paragraph, the reverse-engineering clause, the Authorized-User overbreadth, "cannot be deleted" vs. staff removal, the plain 16+ rule, liability carve-outs and severability, CC-BY attribution if crates.io wording was used — plus the California forum change and the checkbox flow.

Withdrawn or downgraded from the earlier lists: moderation workflow machinery, the three-week process, hourly crons as a gate, per-cookie disclosure, naming Grafana in the policy, the Analytics Engine retention disclosure, and the six-document restructure. On that last one: with the ToS slimmed by the discretion rewrite, three documents (terms, privacy notice, usage policy) are enough. A separate DMCA page remains a real counsel question — §512 safe harbor has statutory elements that cannot be vague — but it can be a follow-up rather than a gate on this PR.

piscisaureus:
## Revised recommendation after the operator constraints above

I agree with this correction. It should supersede the parts of the earlier reviews that proposed encoding today's implementation or a fully specified moderation workflow into binding policy.

The durable goal should be:

1. explain what JSR is trying to provide and how it balances package users against publishers;
2. reserve enough operational discretion to run a free, lightly staffed service;
3. disclaim availability and package risk appropriately;
4. describe personal-data practices truthfully by category and purpose without freezing the current architecture; and
5. retain the contractual rights actually needed to operate and secure a package registry.

### Terms: state principles, not service levels

- Say that JSR is a free public package registry provided on an **as-is/as-available** basis, without an SLA or a promise that the service or any particular feature/account/package will remain available.
- Preserve the core ecosystem bargain: publication is normally permanent because downstream users may depend on it; publishers can yank versions; JSR may restrict, suspend, or remove packages, versions, scopes, or accounts when it believes that is appropriate for security, legal compliance, abuse prevention, service stability, or enforcement of the policies.
- Do **not** guarantee a response, a specific response period, an appeal, a second reviewer, or restoration. A durable formulation is that JSR *may*, where reasonably practicable, notify an affected publisher and consider information they provide, but is not obligated to do so and retains final discretion.
- Keep exceptional removal possible even where it breaks dependents. The policy should explain that JSR weighs publisher interests, downstream reliance, security, privacy, law, and registry integrity—not promise a fixed outcome or process.
- Keep the package-content license necessary to store, reproduce, transform, document, index, distribute, cache, and mirror packages. Add a bounded right to inspect, analyze, and execute package content for malware detection, abuse prevention, supply-chain research, compatibility, reliability, and registry improvement.
- Remove Deploy-specific restrictions that remain a poor fit regardless of specificity: reverse engineering of software available through JSR, broad proprietary-content restrictions that conflict with the open-source site/API, and the generic customer/events/marketing language.

The previous recommendation for a full moderation state machine is therefore withdrawn as a policy prerequisite. Tombstones, structured cases, and replication notices may still be good product features, but the Terms should not promise them before the team decides they are worth operating.

### Privacy: categories, purposes, and criteria

The notice does not need to enumerate every cookie key, database column, scheduler, or current vendor. It should describe stable categories such as:

- account and authentication information received from identity providers;
- public profile, scope, package, documentation, and package metadata;
- service usage, request, download, security, and audit information;
- support, abuse-report, legal, and other correspondence; and
- short-lived authentication and security information stored in the browser or service.

The purposes can likewise remain durable: operating and improving JSR, authenticating users, publishing and distributing packages, reliability/capacity work, preventing and investigating abuse and supply-chain attacks, communicating about accounts or reports, enforcing policies, and complying with law.

Avoid “we collect only what we need,” which is a difficult absolute to prove. Prefer **“we seek to minimize personal information and collect, use, and retain information reasonably necessary and proportionate to operate, secure, maintain, and improve JSR and comply with law.”**

Similarly, “we do not share data with third parties” is literally incompatible with cloud/CDN/email/authentication providers, public registry publication, complaint handling, and legal process. A durable promise is:

> We do not sell personal information or share it for advertising or cross-site behavioral profiling. We disclose information only as reasonably necessary to operate and secure JSR, provide public registry functionality, work with service providers acting on our behalf, investigate abuse, comply with law, protect users or the service, or complete a merger, acquisition, or other change of control.

Describe service-provider **categories** in the policy—hosting/storage, CDN/security, search, authentication, transactional email, and observability—rather than promising that a named table is the complete and permanent list. If transparency about current vendors is desired, maintain a separate non-contractual subprocessor/status page that can evolve without rewriting the core notice.

Use retention criteria rather than scheduler-level guarantees:

- public packages and associated metadata are generally retained indefinitely because others may rely on them, subject to exceptional removal;
- account information is retained while the account exists;
- after deletion, identifying account information is deleted or anonymized after a limited retention period reasonably necessary for security, fraud/abuse and supply-chain investigations, legal obligations, and claims;
- operational/security records are kept only for as long as reasonably necessary for their purpose; and
- backups and caches expire through ordinary rotation.

Counsel should ensure the final notice still supplies the categories, purposes, recipient categories, retention periods **or criteria**, rights, and other disclosures applicable law requires. Generality should make the notice durable, not vague about what JSR actually does.

### What remains necessary before this PR lands

Most of the technical findings from the reviews are now reasons to remove brittle language, not reasons to freeze production around the draft:

- Remove exact “hourly,” “never longer than one hour,” “complete list,” and exhaustive browser-storage claims.
- Do not promise self-service deletion “from account settings” until that exists. Describe what happens *when an account is deleted*, with appropriate security/legal retention, without specifying UI or internal row-reassignment mechanics.
- Replace “anonymous placeholder account” and “single entry” with the stable outcome: the public profile and attribution may be removed or anonymized, while public packages and security/abuse records may remain.
- Replace “we do not track you” with the narrower durable promise that JSR does not use personal information for advertising or cross-site behavioral profiling.
- Remove the conflicting generic vendor/marketing paragraphs.
- Describe immutability accurately: publishers cannot modify published contents or ordinarily delete versions; JSR may exceptionally remove them.

#1493 and #1331 therefore do not have to be artificial merge-order gates **if the policy stops promising their exact future behavior**. They remain useful product/privacy improvements, and OAuth credentials should still be revoked/deleted during account deletion as an engineering matter, but token table names and cascade behavior do not belong in the public policy.

### Contract formation and forum

The earlier California and acceptance recommendations still stand:

- Use California governing law. If an exclusive forum is retained, use San Diego County/Southern District of California with mandatory-law protections; omitting an exclusive forum is also reasonable for a free service.
- If the document keeps an operator content license, indemnity, liability cap, termination rights, governing law, or forum clause, require affirmative acceptance from account holders/publishers. An unchecked checkbox saying **“I agree to the JSR Terms of Service and acknowledge the JSR Privacy Notice”**, with conspicuous links, is the cleanest implementation. Record the Terms version and acceptance timestamp.
- Anonymous browsing and downloads should remain frictionless. Do not assume assent-dependent provisions are enforceable against anonymous users.

In short: use npm as a structural example for a compact registry contract, crates.io for the package-policy principles, and JSR's actual operations for the privacy categories—but deliberately avoid promising today's vendor list, database layout, or a moderation organization that JSR does not have.

piscisaureus:
Superseded by #1495

PR #1494, opened 2026-08-21
fix(terraform): bump cloudflare provider to 5.23.0

crowlKats:
The staging terraform state's `cloudflare_worker.jsr_frontend` entry was written by cloudflare provider 5.23.0 during a state repair (re-importing the Cloud Tasks queues after the Cloud Tasks API transiently returned NOT_FOUND on GetQueue for the CI service account, which caused the failed staging deploys and dropped the queues from state). The pinned 5.19.1 can no longer decode that state entry, so every staging apply now fails with:

```
Error: Resource instance managed by newer provider version
Error decoding "cloudflare_worker.jsr_frontend" from previous state: unsupported attribute "propagation_policy"
```

This bumps the locked cloudflare provider to 5.23.0 (within the existing `>= 5.19.1, < 6.0.0` constraint) and adds darwin hashes for local use. Once this merges, the staging apply also recreates the two `cloudtasks.enqueuer` IAM bindings that the failed applies destroyed, and #1493 (currently wedged in the merge queue on this) can merge.

### DIFF
--- a/terraform/.terraform.lock.hcl
+++ b/terraform/.terraform.lock.hcl
@@ -2,18 +2,20 @@
 # Manual edits may be lost in future updates.
 
 provider "registry.terraform.io/cloudflare/cloudflare" {
-  version     = "5.19.1"
+  version     = "5.23.0"
   constraints = ">= 5.19.1, < 6.0.0"
   hashes = [
-    "h1:HkKPMZ/n+QiExkRUSLjGMTGnuIaph+k932LiTp7CKZM=",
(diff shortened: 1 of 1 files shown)


nathanwhit (approved):

crowlbot (approved):
Lockfile-only change, version bump stays within the existing `>= 5.19.1, < 6.0.0` constraint, and the rationale (staging state was written by 5.23.0 during a repair, 5.19.1 can't decode the resulting `propagation_policy` attribute) is clear and verifiable from the error in the PR description. Nothing to flag — approving to unblock the staging apply / #1493.

PR #1497, opened 2026-08-21
fix: fall back to newest prerelease as latest version for packages with no stable release

### LINKED ISSUE #382: Packages without one or more stable release (according to semver) isn't usable
Packages without a stable release does not show the readme on the jsr.io default page

As an example, https://jsr.io/@pup/pup only shows "This package has not published a stable release yet."

Additionally, there is no meta.json: https://jsr.io/@pup/pup/meta.json gives 404

This makes pre-releasing binary packages really hard, as we have to instruct the users to install specific versions like `@pup/pup@1.0.0-rc.14`, instead of just `@pup/pup`

Expected behavior is

- that the latest stable version is shown on jsr.io and meta.json
- **but** if there is no stable version available 
   - the latest pre-release is shown with a badge "PRE-RELEASE" or similar on jsr.io
   - It is also expected that meta,json is generated for pre-release packages, with an additional key "prerelease" containing the latest pre-release version
   - it is also expected that deno install jsr:@pup/pup installs the latest pre-release version if there are no stable.

### LINKED ISSUE #586: Bug: `deno add` only looks for published versions
## Steps to reproduce

1. Run ` deno run -A jsr:@fresh/init`

```sh

### LINKED ISSUE #637: `jsr add` does not install the `latest` tag even when specified
```
jsr add @std/encoding
```

This adds the the prerelease `1.0.0-rc.2` version to `package.json`:

```jsonc
// package.json
"@std/encoding": "npm:@jsr/std__encoding@^1.0.0-rc.2"
```

However, running `jsr show @std/encoding` displays `0.224.3` as the `latest` version.

```
@std/encoding@0.224.3 | latest: 0.224.3 | versions: 39
```

Specifying `latest` does not fix this:

```
jsr add @std/encoding@latest
```

However, this may be an issue with the JSR registry or even `npm` itself because running `npm install` with an alias results in the same issue.

```
npm install @std/encoding@npm:@jsr/std__encoding
```

```jsonc
// package.json
"@std/encoding": "npm:@jsr/std__encoding@^1.0.0-rc.2"
```

### LINKED ISSUE #663: JSR Score for pre-release versions
It's useful to know the score ahead of release instead of after the release to make changes before a public release.

I assumed the score displayed when you visit a pre-release version represents the same but it does not: https://github.com/denoland/std/pull/5264#issuecomment-2242157430

### LINKED ISSUE #745: cannot install packages with release candidates
https://jsr.io/@cliffy/command/versions

image

```
$ deno cache jsr:@cliffy/command
error: Could not find version of '@cliffy/command' that matches specified version constraint '*'
```

cliffy only has release candiates and thus cannot be downloaded without explicit version marker.

crowlKats:
For packages that only have prerelease versions, the "latest" version is now the newest unyanked prerelease instead of nothing, matching npm's behavior for such packages. This applies consistently to:

- the `package_latest_version` / `package_version_meta` SQL fragments (package pages, package lists, score),
- `meta.json`'s `latest` field,
- the `latest` version alias (`/versions/latest`, `/versions/latest/source`) — the previously docs-only fallback is now the general behavior, so the separate `_for_docs` query is folded in,
- the npm compatibility layer's `latest` dist-tag, which is now also computed by semver order instead of first-row-by-collation, so a prerelease can no longer shadow a newer stable release (fixes the `@std/encoding@1.0.0-rc.2` situation from #637),
- the download chart's recent-versions list.

A new migration replaces the stable-only partial index with one matching the new `ORDER BY (version NOT LIKE '%-%') DESC, version DESC` pattern, so the latest-version lateral join stays an index scan. Unlike earlier revisions of this PR it no longer `INCLUDE`s `meta`, per the index-row-size cap fix from #1506.

The score API and score page now also accept a specific version (`/@scope/pkg@1.0.0-rc.1/score`, `?version=` on the API), so the score of a prerelease can be checked before promoting it to a stable release (#663).

Frontend: prerelease-only packages now get docs, nav tabs, and a score; the version header shows a "pre-release" chip; the "is 0 versions behind" banner degenerate case shows a sensible message; `api.md` documents the `latest` field.

Also fixes a latent bug: five `UPDATE packages ... RETURNING` queries and the sitemap query used uncorrelated subqueries (`WHERE scope = scope AND name = name`), returning globally-wrong `version_count` / `latest_version`.

The Deno CLI side is handled in denoland/deno#36655 (`deno add`/`install` + LSP, fixes #586) and denoland/deno_graph#661 (graph resolution for `deno run`/`deno cache`, fixes #745 once released and the pin is bumped in Deno).

Part of #897.

Fixes #382
Fixes #637
Fixes #663


crowlKats:
@crowlbot rereview

PR #1499, opened 2026-08-22
feat: add repository information to metadata endpoints

### LINKED ISSUE #1031: Add repository information to metadata endpoints
[Our](https://depfu.com) clients have asked us numerous times to support JSR for the npm-ish package managers we support - The main holdup was that we weren't supporting aliased packages (`@std/http@npm:@jsr/std__http@^1.0.12` etc.) which we've now implemented. While testing, I realised that currently the [npm compat metadata endpoint](https://npm.jsr.io/@jsr/luca__cases) does not return a `"repository"` field as [npm normally does](https://registry.npmjs.com/opencolor).

We're using this URL to then gather more information about a package and specifically to find a CHANGELOG file or find out if GitHub releases are used to document changes.

This information is clearly available in the JSR database, as it is displayed on the website, so can we please add this to the metadata in the API as well? This is also missing from the [native JSR metadata endpoint](https://jsr.io/@luca/cases/meta.json).

crowlKats:
Fixes #1031

Adds the linked GitHub repository of a package to both registry metadata endpoints so tools consuming them (e.g. depfu) can discover the source repository:

- **npm compatibility endpoint** (`npm.jsr.io/@jsr/scope__pkg`): adds the npm-standard `repository` field at the package level and per version:
  ```json
  "repository": { "type": "git", "url": "git+https://github.com/owner/repo.git" }
  ```
- **native JSR endpoint** (`jsr.io/@scope/pkg/meta.json`): adds a `githubRepository` field:
  ```json
  "githubRepository": { "owner": "owner", "name": "repo" }
  ```

Both fields are omitted entirely when the package has no linked repository, so existing consumers are unaffected.

Since these documents are cached in the buckets/CDN, linking or unlinking a GitHub repository via `PATCH /api/scopes/:scope/packages/:package` now regenerates both documents and purges their cache entries. The upload-and-purge logic that was previously duplicated verbatim in the yank and version-delete handlers is extracted into a shared `upload_package_manifests` helper and reused in all four places.

Also documents the new fields in `frontend/docs/api.md` and adds a test covering the full lifecycle (no repo → linked → unlinked).

Note: this intentionally does not add `repository` to the `package.json` inside published npm tarballs, as that would require an `NPM_TARBALL_REVISION` bump and regeneration of all tarballs; the issue only concerns the metadata endpoints.

### DIFF
--- a/api/src/api/package.rs
+++ b/api/src/api/package.rs
@@ -488,8 +488,7 @@ pub async fn update_handler(mut req: Request<Body>) -> ApiResult<ApiPackage> {
   }
 
   // Every update variant mutates the aggressively-cached `:package` response,
-  // so cache-bust it (and the scope aggregates) afterwards. Built before the
-  // match because the GitHub-repository arm consumes `scope`/`package_name`.
+  // so cache-bust it (and the scope aggregates) afterwards.
   let registry_url = req.data::<RegistryUrl>().unwrap().0.clone();
   let purge_urls = crate::s3_paths::package_api_cache_urls(
     &registry_url,
@@ -497,7 +496,7 @@ pub async fn update_handler(mut req: Request<Body>) -> ApiResult<ApiPackage> {
     &package_name,
   );
 
-  let result = match body {
+  let result: Result<ApiPackage, ApiError> = match body {
     ApiUpdatePackageRequest::Description(description) => {
       let npm_url = &req.data::<NpmUrl>().unwrap().0;
       let buckets = req.data::<Buckets>().unwrap().clone();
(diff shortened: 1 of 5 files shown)


crowlbot (approved):
Clean addition — verified the backwards-compatibility claim (`skip_serializing_if = "Option::is_none"` on both `PackageMetadata::github_repository` and the npm `repository` fields, so existing consumers see no new required field), and the `upload_package_manifests` extraction is wired into all 4 places that need to regenerate the manifests (both `GithubRepository` update arms, `version_update_handler`, `version_delete_handler`) — traced each call site, no gaps. The new lifecycle test (no repo → linked via yank-triggered regeneration → unlinked) exercises exactly the path that matters: that a repository link/unlink actually busts the cached `meta.json`/npm manifest rather than leaving stale cached copies. CI green. Approving.

PR #1500, opened 2026-08-22
fix: include dependents that use the npm compatibility layer

### LINKED ISSUE #842: Dependents not counted, if installed through "npm:@jsr/"
My package "[maplibre-gl-draw](https://jsr.io/@birkskyum/maplibre-gl-draw/dependencies)" depend on another jsr package "[geojson-normalize](https://jsr.io/@birkskyum/geojson-normalize/dependents)":

Screenshot 2024-11-22 at 00 18 05

But presumably because it's installed though the npm, then the dependent doesn't show up:
https://jsr.io/@birkskyum/geojson-normalize/dependents

Screenshot 2024-11-22 at 00 19 55

crowlKats:
Closes #842

Packages that depend on a JSR package through the npm compatibility layer are stored as `npm` dependencies named `@jsr/scope__name`, but the dependents queries only matched `jsr` dependencies named `@scope/name` — so such dependents never appeared on the dependents page or in the dependent count.

`Database::list_package_dependents` / `Database::count_package_dependents` now take the scope and package name and match either form, using `NpmMappedJsrPackageName` to build the npm-mapped name. Since the fix is purely query-side, existing rows need no migration and npm-layer dependents show up on deploy.

### PR Checklist

- [x] The PR title follows
- [x] Is this closing an open issue? If so, link it, else include a proper
      description of the changes and reason behind them.
- [ ] Does the PR have changes to the frontend? If so, include screenshots or a
      recording of the changes.
- [x] Does the PR have changes to the backend? If so, make sure tests are added.


crowlbot (approved):
Small, correctly-scoped fix. Verified `NpmMappedJsrPackageName`'s `Display` impl produces `@jsr/scope__name` (matching how npm-compat deps are actually stored per `dependency_name`), confirmed both `list_package_dependents` and `count_package_dependents` (plus the `DependentCountCache` wrapper) were updated consistently, and all 3 call sites in `package.rs` pass `&scope`/`&package` correctly. The new end-to-end test publishes a package with an `npm:@jsr/scope__foo` import and asserts it shows up in `foo`'s dependents list/count (2 → 3) — exercises the actual bug, not just the query in isolation. CI green. Approving.

PR #1501, opened 2026-08-22
feat: handle inbound support email as tickets

crowlKats:
Support mail sent to JSR is handled outside the ticketing system today, and the ticketing system could not have taken it anyway: `tickets.creator` and `ticket_messages.author` were both `NOT NULL REFERENCES users(id)`, so only a signed-in user could open a ticket or write on one. Outbound Postmark sends also set no `Message-ID`, so a reply to a JSR notification email could never be threaded back onto the ticket it answered — hence the "Please DO NOT reply to this email" line in the templates.

This adds inbound-email handling so mail to support lands in the same place as tickets opened on the web, and replies thread in both directions.

## What it does

Postmark delivers inbound mail to `POST /api/hooks/postmark` (`api/src/api/hooks.rs`), authenticated with basic auth against a new `POSTMARK_WEBHOOK_PASSWORD`. Threading resolution goes `In-Reply-To`, then `References`, then the `TICKET-YYYYMMDD-XXXXX` number in the subject for clients that drop those headers.

A message that matches nothing opens a ticket owned by an email address rather than an account. The auto-reply carries a claim token that grants the reporter read and reply access on the web, and binds the ticket to their account if they make one. Attachments are stored in a new R2 bucket and served back as opaque bytes, never under the sender's claimed content type.

`closed bool` becomes a five-state enum (`open` / `waiting_on_user` / `waiting_on_support` / `closed` / `spam`), so a reply can reopen a closed ticket, the queue shows who is owed a response, and junk mail has somewhere to sit. The admin list gains a status filter.

## Design notes

- **The auto-reply is stored as a real outbound `ticket_messages` row** (`author IS NULL` = system), not as a separate event kind. That way a single lookup on `ticket_messages.email_message_id` covers auto-replies, staff replies and inbound mail alike, so no extra column or events table is needed to thread a reply back.
- **No background job queue.** The API has no persistent queue, so the webhook processes inline. The unique index on `email_message_id` is what makes a redelivered webhook a no-op, and Postmark retries anything non-2xx.
- **No Slack notifications** — there is no Slack integration here; the staff `newerTicketMessagesCount` badge is the channel.
- The acknowledgement for a *web*-created ticket is deliberately not stored and carries no `Message-ID`, so a reply to it falls back to the subject match. That avoids adding a system message to every existing user's thread.

## Judgement calls worth a look

- Mail failing SPF/DKIM still opens a normal `open` ticket — a misconfigured sender with a real problem should not vanish — but the address is recorded as unproven and shown that way in the UI.
- A reporter email is never auto-linked to a matching JSR account, even when verified. That would make takeover only as hard as spoofing a header chain, and the claim flow already covers the case.

## Deploying

- **The migration drops `tickets.closed`, so it has to land with the API deploy, not ahead of it** — a still-running old revision would break.
- `POSTMARK_WEBHOOK_PASSWORD` needs adding as a GitHub secret (wired through `TF_VAR_postmark_webhook_password` for staging and prod).
- The Postmark inbound stream's webhook URL is set in the Postmark dashboard, not terraform: `https://webhook:<password>@api.jsr.io/hooks/postmark`. This and the rest of the flow are documented in `architecture.md`.

## Testing

219 API tests pass, 10 of them new: ticket creation from mail, redelivery idempotency, both threading paths, reopen-on-reply, auth rejection, missing `Message-ID`, the full claim lifecycle, and an attachment round-trip including the content-type hardening. Clippy `-D warnings`, `cargo fmt`, `sqlx prepare --check`, `deno task lint:frontend`, `deno fmt`, license headers and `terraform fmt` are all clean.


crowlbot (approved):
Reviewed the highest-risk surface of this PR in depth — inbound email is about as hostile an input source as exists, and the security posture here is strong:

- **Webhook auth** (`postmark_inbound_handler`): basic-auth against `POSTMARK_WEBHOOK_PASSWORD`, compared via a real constant-time `secret_eq` (XOR-fold over zipped bytes, length-checked first) — not a naive `==`.
- **Attachment content-type hardening**: verified independently at *both* ends — stored to R2 with `content_type: "application/octet-stream"` regardless of what the sender claimed (`store_attachments`), and served back with the same hardcoded type in `get_attachment_handler`, with an explicit comment naming the threat (HTML/SVG attachment executing script on the API origin). The sender's claimed type is kept only as display metadata, never trusted for serving.
- **`Content-Disposition` filename**: the attacker-controlled filename is reduced to ASCII-graphic, quote/backslash-free characters for the `filename=` fallback and fully percent-encoded (`NON_ALPHANUMERIC`) for the RFC 5987 `filename*` form — no header-injection path via CR/LF or quote-breaking.
- **Email body extraction**: `InboundEmail::body()` only ever reads `stripped_text_reply`/`text_body` (Postmark's plain-text extraction) — the HTML body is never touched, so there's no stored-HTML/XSS vector from an inbound email at all, independent of whatever the frontend does with the string.
- **Claim tokens**: `Uuid::new_v4()` (122 bits, CSPRNG) as a single-use capability — `claim_ticket`'s `UPDATE ... SET claim_token = NULL WHERE id = $1 AND claim_token = $2` validates and invalidates atomically, no TOCTOU window.
- **Ticket access control**: returns `TicketNotFound` rather than `Forbidden` on a failed access check specifically to avoid confirming ticket-ID existence to someone guessing — nice detail.
- Both judgement calls called out in the PR description (SPF/DKIM-failed mail still opens a ticket but is marked unverified; no auto-linking a reporter email to an existing account) are the right calls for the stated threat model.

Two minor, non-blocking observations:
1. `check_ticket_access`'s claim-token comparison (`Uuid::parse_str(provided) == Ok(claim_token)`) isn't constant-time, unlike `secret_eq` for the webhook password. Given a UUIDv4's entropy and that this is over HTTP (not a tight local timing oracle), practical exploitability is negligible — just flagging the inconsistency in case it wasn't a deliberate choice.
2. The subject-line `TICKET-YYYYMMDD-XXXXX` threading fallback has ~100k possible values per day, which is guessable with enough attempts. Worst case is spam/noise injected into an existing ticket's thread as an unverified sender — it doesn't grant read access or claim capability, so the blast radius is small, but worth being aware it's a softer identifier than the claim token.

The deployment-order note (migration drops `tickets.closed`, must land with the API deploy) is correctly called out already — nothing to add there.

Given the size of this PR I didn't trace the terraform/infra files or the frontend ticket UI in the same depth, but the backend attack surface that actually matters for untrusted input is solid. Approving.

crowlbot (commented):
Reviewed the new commit (outbox-based email delivery) — the design holds up well:

- `enqueue`'s row-then-nudge ordering, `get_pending_email_delivery`'s terminal-state check, and the sweep's 5-minute staleness threshold (comfortably longer than both a normal delivery and the queue's own `max_backoff`) together avoid the obvious double-send races — a task still genuinely in flight won't be re-driven by the sweeper.
- All five send sites were consistently migrated onto `enqueue`, and each now logs-and-continues on a queue failure instead of failing the user's request, matching the stated goal.
- Good test coverage of the state machine: sent rows become invisible to both the delivery path and the sweeper, and `record_email_delivery_failure` is tested on both sides of the abandon threshold.

**CI is red on `check` and `test`, and it's not a code issue — it's a migration filename collision.** The new `api/migrations/20260824000000_email_deliveries.sql` shares its `20260824000000` version prefix with `api/migrations/20260824000000_package_versions_latest_index_row_size.sql`, which merged into `main` via #1506 earlier today. `sqlx migrate run` fails with `duplicate key value violates unique constraint "_sqlx_migrations_pkey"` because both migrations register under the same version number. Rebasing/renaming this migration to a later timestamp (e.g. `20260824000001_email_deliveries.sql`) should clear both failures — happy to re-review once that's pushed.

Holding off on approval until CI is green given the fix is code-adjacent, not purely cosmetic like a `deno fmt` issue.

PR #1502, opened 2026-08-22
chore: update deno_doc to 0.204.0

### LINKED ISSUE #865: correctly link openai type references
For example `API.Audio`

https://jsr.io/@openai/openai@4.75.0/doc/~/default

Screenshot 2024-12-11 at 18 14 43

### LINKED ISSUE #917: 404 autogenerated link

### LINKED ISSUE #919: Types imported through `import *` don't get linked

### LINKED ISSUE #922: Better type linking on docs gen

### LINKED ISSUE #957: Re-exported interfaces are missing in the auto-generated docs

### LINKED ISSUE #1013: Symbols missing documentation from "implements"

### LINKED ISSUE #1301: Some doc symbols lead to 404
Repro:

1. Navigate to https://jsr.io/@zod/zod/doc
2. Click the topmost symbol

Perhaps it is related to `.` characters inside the symbol name?

crowlKats:
Brings in the upstream docs-gen fixes for the P1 board issues and the zod 500s, plus the JSR-side changes needed to surface them.

## Upstream (deno_doc 0.204.0)

- denoland/deno_doc#837 — type references through namespace imports (`import * as ns`), aliased imports, and default imports get linked; links into other entrypoints are verified to exist; internal (non-exported / `@internal`) symbols are no longer linked, since JSR doesn't serve pages for them.

  Fixes #919, fixes #865, fixes #917, fixes #922
- denoland/deno_doc#838 — declarations are no longer lost when an export name collides: `export { foo as Bar }` next to `export interface Bar`, or the same name coming through multiple `export *` modules (the `@babel/types` case, where builder functions shadowed the same-named interfaces from an `export type *`).

  Fixes #957
- denoland/deno_doc#841 — `symbol_listing_limit` (set here to 2048): namespace-heavy packages otherwise produce module symbol listings with tens of thousands of rows. zod's entrypoint docs context serializes to 41 MB, blowing the docs API response past Cloud Run's 32 MiB response cap and turning every `/doc`, `all_symbols`, and `search_structured` request for the package into an empty 500 (measured against zod 4.4.3's production doc nodes; with the cap the page is 1.2 MB). Also fixes the "Symbols" panel linking namespace members by their bare name.

  Fixes #1301, and fixes the `https://jsr.io/@zod/zod/doc` 500s
- denoland/deno_doc#842 — `implements ns.Interface` renders as a proper qualified, linked name instead of `[ns.Interface]`; class members inherit docs from the interfaces they implement; `@example` on interface construct/call signatures renders inline.

  Fixes #1013, and the remainder of #1250

## JSR-side changes

- `generate_symbol_page` no longer drops a node's non-reference declarations when resolving a reference, and when one symbol name yields multiple doc nodes their declarations are merged before rendering — so e.g. `@babel/types`' `ArrayExpression` page shows both the re-exported builder function and the interface.
- The frontend's JSX doc components render the same context JSON that deno_doc's handlebars templates do, so every template change in the 0.201.0 → 0.204.0 range is mirrored:
  - **DocEntry** renders the new `params` / `return_doc` / `examples` fields (parameter, return, and `@example` docs for entries with no symbol page of their own — class constructors, construct signatures, call signatures), with the same bordered-left layout as deno_doc's `.docEntryParams`.
  - **Example** containers carry their anchor id, so `#example_N` links scroll correctly.
  - **ModuleDoc** renders the `symbols_trimmed` note ("Some deeply nested symbols were omitted from this overview; they are listed on the page of their containing namespace."). Typed locally for now — `symbols_trimmed` is missing from `@deno/doc@0.204.0`'s `html-types` (upstream fix filed).
  - **Tag** no longer title-cases `"other"` tags (the `@since` version chips carry preformatted text); the dead valued-tag branch is removed since only `permissions` and `other` carry values. The `other` tag color tokens already existed in our CSS.
  - A base `.anchor` CSS rule so the clickable heading anchor links now embedded in rendered markdown (plain `class="anchor"` markup, unlike the Anchor component's inline utilities) are hidden until hover like every other anchor.
  - The `@event`/`@fires`/`@listens` sections and interface-member rendering changes arrive as existing section kinds (`see`, `doc_entry`), which the components already handle.
- deno_graph `0.109` → `0.110.2`: two new `BuildOptions` fields, kept at their defaults; `@deno/doc` frontend dependency `^0.201.0` → `^0.204.0`.

## Tests

`test_package_docs_merged_export_name` regression test with a tarball fixture mirroring the babel shape. Full `registry_api` suite passes (205 tests) on 0.204.0; frontend typechecks, lints, and builds.

Note: existing packages need a docs regeneration (or a new version publish) to pick up the parser-level fixes, since stored doc nodes are cached per version; the render-level fixes (linking, listing cap, template mirrors) apply on deploy.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1056,9 +1056,9 @@ dependencies = [
 
 [[package]]
 name = "deno_doc"
-version = "0.203.0"
+version = "0.204.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "b77987b58779235c841893c3434dd7ce9eadd64fd53695fe2ddd99210e951c97"
+checksum = "05c121a45c4001f3b89cc7af2e3cab71514db6807ded0c9a0317b118d3d3fe87"
 dependencies = [
  "anyhow",
  "cfg-if",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1056,9 +1056,9 @@ dependencies = [
 
 [[package]]
 name = "deno_doc"
-version = "0.201.0"
+version = "0.203.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "4bbfdb2c55ab195429d00c54d44da0f57d72ee36dff54e6c8de1a594b5ec26e0"
+checksum = "b77987b58779235c841893c3434dd7ce9eadd64fd53695fe2ddd99210e951c97"
 dependencies = [
  "anyhow",
  "cfg-if",
@@ -1110,9 +1110,9 @@ dependencies = [
 
 [[package]]
 name = "deno_graph"
-version = "0.109.0"
+version = "0.110.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "18b1634ebdfde4f040b2c40f2510969e755e5b782319a01dc9cab88ef25b7e3f"
+checksum = "f95ac3cf9d132772c24beabf61d76f6706b7d5a23cca6d3d3f4cf2d30a7ad796"
 dependencies = [
  "async-trait",
  "boxed_error",
--- a/api/Cargo.toml
+++ b/api/Cargo.toml
@@ -95,10 +95,10 @@ flate2 = "1"
 thiserror = "2"
 async-tar = "0.4.2"
 async-compression = { version = "0.4", features = ["futures-io", "gzip"] }
-deno_graph = "=0.109.0"
+deno_graph = "=0.110.2"
 deno_ast = { version = "0.53.0", features = ["view"] }
 # sync with frontend/deno.json
-deno_doc = { version = "=0.201.0", features = ["comrak"] }
+deno_doc = { version = "=0.203.0", features = ["comrak"] }
 deno_error = "0.7.0"
 comrak = { version = "0.29.0", default-features = false }
 ammonia = "4.0.0"
--- a/api/Cargo.toml
+++ b/api/Cargo.toml
@@ -98,7 +98,7 @@ async-compression = { version = "0.4", features = ["futures-io", "gzip"] }
 deno_graph = "=0.110.2"
 deno_ast = { version = "0.53.0", features = ["view"] }
 # sync with frontend/deno.json
-deno_doc = { version = "=0.203.0", features = ["comrak"] }
+deno_doc = { version = "=0.204.0", features = ["comrak"] }
 deno_error = "0.7.0"
 comrak = { version = "0.29.0", default-features = false }
 ammonia = "4.0.0"
--- a/api/src/analysis.rs
+++ b/api/src/analysis.rs
@@ -167,6 +167,8 @@ async fn analyze_package_inner(
         unstable_text_imports: false,
         jsr_metadata_store: None,
         unstable_css_imports: false,
+        unstable_config_imports: false,
+        prefer_cached_jsr_versions: false,
       },
     )
     .await;
@@ -658,6 +660,8 @@ async fn rebuild_npm_tarball_inner(
         unstable_text_imports: false,
         jsr_metadata_store: None,
         unstable_css_imports: false,
+        unstable_config_imports: false,
+        prefer_cached_jsr_versions: false,
       },
     )
     .await;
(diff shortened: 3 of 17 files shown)


KnorpelSenf (approved):
LGTM. Good progress these days!

crowlbot (approved):
Reviewed the JSR-side logic (the dependency bump itself is low-risk). `resolve_doc_node_references` correctly fixes the drop bug: previously any node with a reference declaration flat-mapped to *only* the resolved reference targets, discarding sibling non-reference declarations on the same node — now both the resolved references and a stripped copy of the non-reference declarations are kept. The multi-node merge in `generate_symbol_page` (dedup by `(location, kind)`, preferring the node matching the requested name as the base) correctly handles the case where the same symbol name resolves to multiple doc nodes.

The new `merged_export_name` fixture mirrors the described babel scenario precisely (function re-exported under an alias colliding with a same-named interface from `export type *`), and the test asserts both docblocks ("builder function" / "node interface") appear on the rendered symbol page — a real assertion on the actual bug, not just "it doesn't crash." CI green. Approving.

crowlbot (approved):
Reviewed the delta since my previous approval (commits `e31f33a`, `46d9760` — deno_doc 0.203.0 → 0.204.0 bump + the `SYMBOL_LISTING_LIMIT`/frontend mirror work; the merge/dedup logic I already reviewed is untouched):

- `SYMBOL_LISTING_LIMIT = 2048` in `docs.rs`: well-justified (zod's docs context was 41 MB, over Cloud Run's 32 MiB response cap; capped it's 1.2 MB) and passed straight through to `symbol_listing_limit` on the generate context.
- `DocEntry.tsx`: adds `params`/`return_doc`/`examples` rendering for entries with no dedicated symbol page (constructors, construct/call signatures) — recursing `DocEntry` for params is a clean way to reuse the existing rendering rather than duplicating it.
- `Example.tsx`: `id={anchor.id}` on the container fixes `#example_N` anchor links actually scrolling to the example — was previously only on `Anchor`'s own markup.
- `styles.css` `.anchor` base rule: needed because markdown-embedded heading anchors arrive as plain `class="anchor"` markup rather than through the `Anchor` component's inline utility classes — matches the existing hover-to-reveal pattern used elsewhere.
- `Tag.tsx`: dropping `titleCase` for "other" tag values (e.g. `@since` chips) is correct — those already carry preformatted text.
- `ModuleDoc.tsx`: the `symbols_trimmed` cast is a reasonable stopgap for a field missing from `@deno/doc@0.204.0`'s published `html-types`, clearly flagged to remove once upstream catches up.

CI is green (including `check` and `frontend`, so the `Tag.value` cast removal type-checks). No concerns — still approved.

PR #1503, opened 2026-08-22
fix: transpile TSX/JSX in npm-compat tarballs

lucacasonato:
## Summary

npm-compat tarballs previously left `.tsx` sources untranspiled. `MediaType::Tsx` fell through the packing match (it was not treated like TypeScript/MTS), so packages such as [`@luca/highlightable-textarea`](https://jsr.io/@luca/highlightable-textarea) shipped a raw `index.tsx` as the npm entry. Downstream consumers then failed to load the module.

This change:

- Treats `MediaType::Tsx` like TypeScript: emit JS (+ source map) and optional fast-check `.d.ts`.
- Reads package JSX `compilerOptions` (`jsx`, `jsxImportSource`, `jsxFactory`, `jsxFragmentFactory`) from `deno.json(c)` / `jsr.json(c)`, and honors `@jsxImportSource` file pragmas.
- Rewrites specifiers *after* `fold_program`, so injected `jsx-runtime` imports (e.g. `npm:preact/jsx-runtime` → `preact/jsx-runtime`) are remapped for npm.
- Bumps `NPM_TARBALL_REVISION` 11 → 12 so existing packages get new tarballs.

## Tests

New npm-tarball specs:

- `transpile_tsx.txt` — TSX + `@jsxImportSource` pragma
- `transpile_jsx_with_import_source.txt` — JSX + pragma
- `transpile_tsx_compiler_options.txt` — TSX + `compilerOptions.jsx` / `jsxImportSource`

Fixes #996

PR #1504, opened 2026-08-22
fix: transpile TSX/JSX in npm-compat tarballs

### LINKED ISSUE #24: Support publishing with JSX
To publish with JSX, we need to take the JSX relevant config options from `tsconfig.json` / `deno.json` and put them into the source files via pragmas.

For example, when publishing this index.jsx file with this deno.json file:

```jsx
import { renderToString } from "npm:preact-render-to-string";
console.log(renderToString(<div>jsx</div>));
```

```json
{
  "compilerOptions": {
    "jsx": "react-jsx",
    "jsxImportSource": "npm:preact"
  }
}
```

emit this before publish:

```jsx
/** @jsxRuntime automatic *//** @jsxImportSource npm:preact */
import { renderToString } from "npm:preact-render-to-string";
console.log(renderToString(<div>jsx</div>));
```

### LINKED ISSUE #996: JSX/React - possible "NPM tarball transpile not working correctly for TSX"
Simple reproduction repo at: https://github.com/philcockfield/deno-jsr-jsx-issue
Detailed notes in [README](https://github.com/philcockfield/deno-jsr-jsx-issue) ↑

---

## Context
In or around the release of `deno 2.2` the JSX warnings for `unsupported-jsx-tsx` were removed, suggesting the ability to cleanly publish JSX components from within `.tsx` of a package.

The sample repo shows an error when importing into a `Deno + Vite` application, a JSR published module that exports a React/JSX component from a `.tsx` file, with no special 3rd part bundling or transpliation.  Just what Deno/JSR prepares as it publishes the module.

@lucacasonato suggested this may be an issue with the "NPM tarball transpile [maybe] not working correctly for TSX."

---

# README
From the [reproduction repo](https://github.com/philcockfield/deno-jsr-jsx-issue):

### Structure

- `./sample-react/deno.json`: The sample JSR registry module exposing a `<Foo>` React component `.tsx` file.
- `./sample-vite/deno.json`: Vite application that imports from `sample-react`

### Run

To recreate the error start the Vite server:

```bash
cd ./sample-vite
deno task install  # runs `deno install` with --allow-scripts to ensure the Vite plugins fully install.
deno task dev
```

The server will start and render the `<App>` component that in turn imports from JSR:

```ts
import { Foo } from "jsr:@phil/sample-react";
```

This will cause errors to be output into the browser's console:

```
3b0776a……3b3bc6015c5772816:7 Uncaught TypeError: jsx is not a function
    at Foo (3b0776a……bc6015c5772816:7:26)
```

lucacasonato:
## Summary

npm-compat tarballs previously left `.tsx` sources untranspiled. `MediaType::Tsx` fell through the packing match, so packages such as [`@luca/highlightable-textarea`](https://jsr.io/@luca/highlightable-textarea) shipped a raw `index.tsx` as the npm entry.

This change:

- Treats `MediaType::Tsx` like TypeScript: emit JS (+ source map) and optional fast-check `.d.ts`.
- Honors JSX settings from the pragmas `deno publish` / `jsr publish` already inject into jsx/tsx files (`@jsxRuntime`, `@jsxImportSource`, `@jsxFactory`, `@jsxFragmentFactory`). Does not re-read `compilerOptions` from `deno.json` / `jsr.json`.
- Rewrites specifiers *after* `fold_program`, so injected `jsx-runtime` imports (e.g. `npm:preact/jsx-runtime` → `preact/jsx-runtime`) are remapped for npm.
- Bumps `NPM_TARBALL_REVISION` 11 → 12 so existing packages get new tarballs.

## Tests

- `transpile_tsx.txt` — TSX + `@jsxImportSource` pragma
- `transpile_jsx_with_import_source.txt` — JSX + pragma
- `transpile_tsx_classic_factory.txt` — classic `@jsxFactory` / `@jsxFragmentFactory`

Fixes #24 
Fixes #996
Supersedes #1503


api/src/npm/tarball.rs
```diff
@@ -146,19 +146,25 @@ pub async fn create_npm_tarball<'a>(
         {
           declaration_rewrites
             .insert(module.specifier(), resolved.specifier.clone());
+        } else if let Some(source_specifier) = source_specifier {
+          declaration_rewrites.insert(module.specifier(), source_specifier);
```
lucacasonato, line 150: What's this for?

lucacasonato: So we don't emit `"types": "./index.jsx"`.

Without this, `create_npm_exports` follows the original `.jsx` (the leftover source we still pack) and treats it as the types entry. Same fallback as the TS arm: map declaration to the emitted `.js` so types and default match and types is omitted.

api/src/npm/emit.rs
```diff
@@ -126,3 +129,139 @@ pub fn transpile_to_dts(
...
+  let factory = pragma_value(&header, "@jsxFactory");
+  let fragment_factory = pragma_value(&header, "@jsxFragmentFactory");
+
+  let automatic = JsxAutomaticOptions {
+    development: matches!(runtime.as_deref(), Some("react-jsxdev")),
+    import_source: import_source.clone(),
+  };
+
+  match runtime.as_deref() {
+    Some("precompile") => {
```
crowlKats, line 158: **Panic: `@jsxRuntime react-jsx|react-jsxdev|precompile` crashes tarball generation.** These arms advertise runtime values that swc's directive validator hard-rejects: `JsxDirectives::from_comments` emits a `Level::Error` ("Runtime must be either `classic` or `automatic`.") for any block-comment `@jsxRuntime` value other than classic/automatic, `ensure_no_fatal_swc_diagnostics` turns that into `Err`, and `create_npm_tarball` calls `transpile_to_js(...).unwrap()` (tarball.rs:280). A published `index.tsx` starting with `/** @jsxRuntime react-jsx */` panics the publish/rebuild task instead of returning an error — and the `NPM_TARBALL_REVISION` 11->12 bump replays this over every already-published `.tsx` package. (Pre-existing for `.jsx` on main; newly reachable for `.tsx`.)

api/src/npm/emit.rs
```diff
@@ -126,3 +129,139 @@ pub fn transpile_to_dts(
...
+    }
+  }
+}
+
+fn classic_jsx_runtime(
+  factory: Option<String>,
+  fragment_factory: Option<String>,
+) -> JsxRuntime {
+  JsxRuntime::Classic(JsxClassicOptions {
+    factory: factory.unwrap_or_else(|| "React.createElement".into()),
```
crowlKats, line 184: **Panic: unvalidated `@jsxFactory` token reaches swc's eager `parse_expr_for_jsx`.** The factory string is handed to swc's `react::Options.pragma`, which is parsed at transform *construction* — before deno_ast installs its diagnostics HANDLER — so a non-expression value is a straight `panic!`. Prose is enough to trigger it: a doc comment like `/** Supports @jsxFactory, @jsxFragmentFactory */` in the first 50 lines makes `pragma_value` extract `","` as the factory, and swc panics with `failed to parse jsx option pragma: ',' is not an expression`, killing tarball generation. No JSX element is required in the file.

api/src/npm/emit.rs
```diff
@@ -126,3 +129,139 @@ pub fn transpile_to_dts(
...
+        skip_elements: None,
+        dynamic_props: None,
+      }))
+    }
+    Some("automatic") | Some("react-jsx") | Some("react-jsxdev") => {
+      Some(JsxRuntime::Automatic(automatic))
+    }
+    Some("classic") => Some(classic_jsx_runtime(factory, fragment_factory)),
+    _ => {
+      if import_source.is_some() {
```
crowlKats, line 170: **Panic: false-positive `@jsxImportSource` + a real `/** @jsx h */` pragma.** When the raw-text scan matches an `@jsxImportSource` occurrence that swc ignores (line comment, prose), this arm forces `JsxRuntime::Automatic`. swc then sees the file's genuine `@jsx h` block pragma with runtime Automatic and emits the fatal "pragma cannot be set when runtime is automatic" error -> `fold_program` returns `Err` -> `.unwrap()` panic at tarball.rs:280. Example trigger: `/** @jsx h */` plus `// TODO: switch to @jsxImportSource preact`. This file transpiled fine before the PR.

api/src/npm/emit.rs
```diff
@@ -126,3 +129,139 @@ pub fn transpile_to_dts(
...
+    factory: factory.unwrap_or_else(|| "React.createElement".into()),
+    fragment_factory: fragment_factory
+      .unwrap_or_else(|| "React.Fragment".into()),
+  })
+}
+
+/// Value after `@tag`. `@jsxImportSource` must not match `@jsxImportSourceTypes`.
+fn pragma_value(text: &str, tag: &str) -> Option<String> {
+  let mut search_from = 0;
+  while let Some(rel) = text[search_from..].find(tag) {
```
crowlKats, line 193: **`pragma_value` scans raw text, not leading block comments — false positives ship broken imports with no dependency.** deno_graph only honors the pragma in leading *block* comments via the anchored `(?i)^[\s*]*@jsxImportSource\s+(\S+)` regex, and package.json `dependencies` come solely from graph modules. A `// @jsxImportSource preact` line comment, prose (`/** Use the @jsxImportSource pragma */` -> import source `"pragma"`, empirically verified), or a string literal in the first 50 lines makes emit force Automatic and publish `import { jsx as _jsx } from "preact/jsx-runtime"` while `dependencies` stays empty -> ERR_MODULE_NOT_FOUND for every npm consumer, with no publish-time error. Consider consuming `module_info.jsx_import_source` (already computed at tarball.rs:251) or the ParsedSource leading comments instead.

api/src/npm/emit.rs
```diff
@@ -126,3 +129,139 @@ pub fn transpile_to_dts(
...
+    let idx = search_from + rel;
+    let after = &text[idx + tag.len()..];
+    if after.starts_with(|c: char| c.is_ascii_alphanumeric()) {
+      search_from = idx + tag.len();
+      continue;
+    }
+    let rest = after.trim_start();
+    let value = rest
+      .split(|c: char| c.is_whitespace() || c == '*')
+      .find(|s| !s.is_empty() && *s != "*/")?;
```
crowlKats, line 203: **Tokenization bugs (all empirically verified by running this function).** Splitting on `'*'` means no token can ever equal `"*/"`, so that filter is dead and a value-less pragma captures the comment-close remnant: `/** @jsxImportSource */` -> `Some("/")` -> emitted `import ... from "//jsx-runtime"`. `'*'` in a real value is truncated: `npm:preact@*` -> `"npm:preact@"` -> literal `npm:preact@/jsx-runtime` ships. The first match wins, so a value-less pragma masks a later valid one. `// @jsxImportSource` at end of line steals the next code token (`"import"`). Matching is also case-sensitive while deno_graph's regex is `(?i)`.

api/src/npm/emit.rs
```diff
@@ -126,3 +129,139 @@ pub fn transpile_to_dts(
...
+
+  let header: String = source
+    .text()
+    .lines()
+    .take(50)
+    .collect::<Vec<_>>()
+    .join("\n");
+  let import_source = pragma_value(&header, "@jsxImportSource");
+  let runtime = pragma_value(&header, "@jsxRuntime");
+  let factory = pragma_value(&header, "@jsxFactory");
```
crowlKats, line 149: **`@jsxFactory`/`@jsxFragmentFactory` are not pragmas anywhere else — this creates a JSR-only dialect that diverges from Deno.** TypeScript/swc/Deno use `@jsx` and `@jsxFrag`; `jsxFactory`/`jsxFragmentFactory` are tsconfig/deno.json option names. With `/** @jsxFactory h */` (as enshrined in the new `transpile_tsx_classic_factory.txt` spec), the npm tarball emits `h("div", ...)` while Deno/tsc/esbuild emit `React.createElement(...)` for the identical source — the npm artifact and the Deno artifact of one published version behave differently. The snapshotted index.js also calls `h` with no import binding and `dependencies: {}` (unbound identifier at runtime). The standard `/** @jsx h */` works only by accident, via swc's own directive parsing.

api/src/npm/import_transform.rs
```diff
@@ -14,18 +14,35 @@ use deno_ast::swc::ecma_visit::VisitMutWith;
...
   pub specifier_rewriter: SpecifierRewriter<'a>,
   pub kind: RewriteKind,
 }
 
+impl ImportRewriteTransformer<'_> {
+  fn remap(&self, value: &str) -> Option<String> {
+    self
+      .specifier_rewriter
+      .rewrite(value, self.kind)
+      .or_else(|| rewrite_npm_and_jsr_specifier(value))
```
crowlKats, line 29: **The fallback only fires for specifiers deno_graph never validated — it converts loud leftovers into silently-broken bare imports.** For the normal injected jsx-runtime import, the graph already contains the exact key (`format!("{}/jsx-runtime", import_source)`), so `SpecifierRewriter::rewrite` handles it without this fallback. When the fallback does fire: (a) `@jsxRuntime react-jsxdev` makes swc inject `<src>/jsx-dev-runtime`, which is never a graph key -> a development-runtime import ships that no dependency declares; (b) a type-only or dynamic `jsr:@scope/foo/missing` whose resolution errored passes `graph.valid()` (CodeOnly walk, `follow_dynamic: false`) and gets rewritten to `@jsr/scope__foo/missing` — an uninstallable self-reference. It also bypasses the `NpmMissingConstraint` version check that graph-seen deps go through.

api/src/npm/emit.rs
```diff
@@ -35,27 +39,20 @@ pub fn transpile_to_js(
...
-      specifier_rewriter,
-      kind: RewriteKind::Source,
-    };
-    program.visit_mut_with(&mut import_rewrite_transformer);
-
     let transpile_options = TranspileOptions {
       decorators: DecoratorsTranspileOption::Ecma,
-
-      // TODO: JSX
+      jsx: resolve_jsx_runtime(source),
```
crowlKats, line 51: **`jsx: None` for `.ts`/`.mts` is not equivalent to the old default — it silently changes emitted JS for plain TypeScript registry-wide.** The old `..Default::default()` was `jsx: Some(Classic)`, and this code calls `fold_program` directly, bypassing `ParsedSource::transpile`'s per-media-type normalization, so every `.ts` module switches swc pipelines (`typescript::tsx`+react -> `typescript::typescript`). Concrete delta: `import React from "npm:react@18"` used only in type positions was preserved pre-PR (the tsx pass seeds `id_usage` with the pragma root) and is now stripped. The new behavior matches tsc/Deno and incidentally fixes a latent `.ts` directive-validation panic — but no spec pins it and the PR description doesn't mention it. Suggest adding a `.ts` spec with a type-only `React` import and noting the side effect.

api/src/npm/tarball.rs
```diff
@@ -146,19 +146,25 @@ pub async fn create_npm_tarball<'a>(
...
         }
       }
       deno_ast::MediaType::Dts | deno_ast::MediaType::Dmts => {
         // no extra work needed for these, as they can not have type dependencies
       }
-      deno_ast::MediaType::TypeScript | deno_ast::MediaType::Mts => {
+      deno_ast::MediaType::TypeScript
+      | deno_ast::MediaType::Mts
+      | deno_ast::MediaType::Tsx => {
```
crowlKats, line 158: **`.tsx` now collides with a same-stem `.ts`/`.jsx` — silent overwrite in the tarball.** Both `/foo.ts` and `/foo.tsx` rewrite to `/foo.js`, and `package_files.insert` discards the previous value, so graph iteration order decides which module's transpiled output survives (same for `/_dist/foo.d.ts`). Nothing in publish validation forbids the layout (the only duplicate check is case-insensitive whole-path equality), and imports of both files are rewritten to `./foo.js`, so consumers hit `does not provide an export named ...` at runtime. The collision class pre-existed for `.ts`+`.jsx`; this PR silently extends it to `.tsx`. A duplicate-key guard on `package_files` (like the `unreachable!` in api/src/tarball.rs:358) would at least make it loud.

api/src/npm/tarball.rs
```diff
@@ -146,19 +146,25 @@ pub async fn create_npm_tarball<'a>(
         {
           declaration_rewrites
             .insert(module.specifier(), resolved.specifier.clone());
+        } else if let Some(source_specifier) = source_specifier {
```
crowlKats, line 149: **This `else if` can chain `declaration_rewrites` through an unrelated module's types.** It inserts the derived `/foo.js` path as a map *value*, and `follow_specifier` walks the map transitively. If the package also ships a real hand-written `/foo.js` with its own types (e.g. `/// <reference types="./foo.d.ts" />`), loop 1 inserts `/foo.jsx -> /foo.js` (here) and `/foo.js -> /foo.d.ts` (JavaScript arm), so `create_npm_exports` and Declaration-kind rewriting resolve the JSX export to `./foo.d.ts` — a completely unrelated declaration file typing the JSX module. Pre-PR the `.jsx` had no `declaration_rewrites` entry, so the chain could not form. Same applies to the Tsx `else if` below.

api/src/npm/emit.rs
```diff
@@ -126,3 +129,139 @@ pub fn transpile_to_dts(
...
+    }
+    Some("automatic") | Some("react-jsx") | Some("react-jsxdev") => {
+      Some(JsxRuntime::Automatic(automatic))
+    }
+    Some("classic") => Some(classic_jsx_runtime(factory, fragment_factory)),
+    _ => {
+      if import_source.is_some() {
+        Some(JsxRuntime::Automatic(automatic))
+      } else {
+        Some(classic_jsx_runtime(factory, fragment_factory))
```
crowlKats, line 173: **Pragma-less `.tsx` silently emits `React.createElement` with no import and no `react` dependency.** A JSX file with no pragma (e.g. JSX configured via deno.json compilerOptions, which api/src never reads) produces a `/index.js` calling a free `React` identifier with `dependencies: {}` -> ReferenceError at first render. `@jsxRuntime automatic` without an import source similarly emits `import { jsx as _jsx } from "react/jsx-runtime"` (swc's default) with empty deps -> ERR_MODULE_NOT_FOUND. Pre-PR `.tsx` shipped raw (load-time extension error), so this is a sideways move to a quieter failure — and neither path has a spec. Erroring at publish when no import source is resolvable would be safer than defaulting to React.

api/src/npm/import_transform.rs
```diff
@@ -14,18 +14,35 @@ use deno_ast::swc::ecma_visit::VisitMutWith;
...
 use super::specifiers::SpecifierRewriter;
+use super::specifiers::rewrite_npm_and_jsr_specifier;
 
 pub struct ImportRewriteTransformer<'a> {
   pub specifier_rewriter: SpecifierRewriter<'a>,
   pub kind: RewriteKind,
 }
 
+impl ImportRewriteTransformer<'_> {
+  fn remap(&self, value: &str) -> Option<String> {
```
crowlKats, line 25: **Fallback asymmetry: only the AST path gets it; the text path does not.** `rewrite_specifiers` (used for `.js`/`.d.ts` files and the shipped `.ts`/`.tsx` source copies) still calls the bare `specifier_rewriter.rewrite`, so the same unresolved specifier is rewritten in one artifact and left as raw `npm:` text in its sibling. The PR's own spec shows it: `transpile_tsx_with_dts.txt` has `/_dist/index.d.ts` saying `from "preact"` while the shipped `/index.tsx` keeps `import type { JSX } from "npm:preact@^10.22.0"`. If the fallback stays, it belongs inside `SpecifierRewriter::rewrite` so both paths share one definition.

api/testdata/specs/npm_tarballs/transpile_tsx_with_dts.txt
```diff
@@ -0,0 +1,76 @@
...
+
+# npm:preact@^10.22.0
+<external>
+
+# npm:preact@^10.22.0/jsx-runtime
+<external>
+
+# output
+== /_dist/index.d.ts ==
+/** @jsxImportSource npm:preact@^10.22.0 */ import type { JSX } from "preact";
```
crowlKats, line 23: **Stale Deno-only pragma ships in the emitted artifacts (snapshotted here).** `remove_comments: false` plus no comment rewriting means `/** @jsxImportSource npm:preact@^10.22.0 */` survives verbatim in `/_dist/index.d.ts` and `/index.js` while the adjacent import was rewritten to `preact` — a self-contradictory, npm-unresolvable `npm:` specifier locked into three spec files. Inert for tsc reading the `.d.ts` (declaration files skip the synthetic jsx-runtime import), but honored by any tool that retranspiles the shipped `.js`/`.tsx` as JSX. Cheap fix: strip or rewrite `@jsx*` pragma comments during emit.

api/src/npm/emit.rs
```diff
@@ -126,3 +129,139 @@ pub fn transpile_to_dts(
 
   Ok((source, source_map.unwrap().into_bytes()))
 }
+
+fn resolve_jsx_runtime(source: &ParsedSource) -> Option<JsxRuntime> {
```
crowlKats, line 133: **Altitude: this whole resolver duplicates what the pipeline already does, and the copy is the buggy one.** deno_ast passes the file's comments into swc's react transform, whose `JsxDirectives` parser already honors `@jsxImportSource`/`@jsxRuntime`/`@jsx`/`@jsxFrag` from leading block comments and overrides the passed options (deno_ast's own `test_transpile_jsx_import_source_pragma` proves `/** @jsxImportSource x */` works with plain defaults). deno_graph's `module_info.jsx_import_source` is also already in hand at tarball.rs:251. Replacing this function's body with `Some(JsxRuntime::default())` for Jsx/Tsx media types (and `None` otherwise) keeps the PR's real fixes — the Tsx arms and running ImportRewriteTransformer after fold_program — while eliminating the panic paths, the false-positive text scanning, and the invented pragma dialect in one move.

api/src/npm/import_transform.rs
```diff
@@ -14,18 +14,35 @@ use deno_ast::swc::ecma_visit::VisitMutWith;
...
+    self
+      .specifier_rewriter
+      .rewrite(value, self.kind)
+      .or_else(|| rewrite_npm_and_jsr_specifier(value))
+  }
+
```
crowlKats, line 31: **Cleanup: `remap_declaration` duplicates `remap` modulo the kind, and `visit_mut_ts_import_type` is now dead in the `transpile_to_js` path.** After the reorder, `fold_program` has already erased TS type nodes before the visitor runs, so the TsImportType arm only fires from `transpile_to_dts`. A single `fn remap(&self, value: &str, kind: RewriteKind)` called with `self.kind` at the import/export sites and `RewriteKind::Declaration` in the TsImportType arm keeps the fallback chain defined once.

crowlKats (commented):
Automated deep review (10 finder angles, 8 verification passes; pragma_value defects verified by compiling and running the function). 15 findings posted inline, ranked roughly by severity: 3 panic paths reachable from published file contents (amplified registry-wide by the NPM_TARBALL_REVISION bump), pragma parsing divergences from deno_graph/swc/tsc that ship broken imports without package.json dependencies, a new same-stem output collision, a declaration_rewrites chaining hazard, a silent registry-wide .ts emit change, and layering/cleanup notes. The deep fix for most of these: drop resolve_jsx_runtime/pragma_value and rely on swc's directive parsing (see the comment on emit.rs:133).

PR #1506, opened 2026-08-24
fix: publishing tasks with oversized version meta wedge in pending forever

### LINKED ISSUE #1505: Publishing task never starts for packages with very large export maps (3.8k / 19.2k exports)
## Summary

Two packages in my scope have been permanently stuck at the "publishing task created but
never started" stage since 2026-08-21. The upload completes, the task is created, and the
status page shows **"Publishing has not started yet"** indefinitely. No error is ever
returned — the CLI simply polls forever, and the version never appears.

Both stuck packages have unusually large `exports` maps (3,824 and 19,201 entries). Every
other package in the same scope — including one published *between* the two failures on the
same day, from the same machine — publishes normally. All documented quotas and size limits
are satisfied with substantial headroom.

## Affected packages

| Package | Exports | Version attempted | `versionCount` |
|---|---:|---|---:|
| `@marianmeres/icons-fns` | 19,201 | 6.0.0, then 6.0.1 | 0 |
| `@marianmeres/emoji-fns` | 3,824 | 2.0.1 | 0 |

Publishing task status pages:

- `@marianmeres/icons-fns@6.0.0` — https://jsr.io/status/<-- TASK UUID -->
- `@marianmeres/emoji-fns@2.0.1` — https://jsr.io/status/<-- TASK UUID -->

## Timeline (same scope, same day, same machine, same Deno version)

| Time (UTC) | Package | Exports | Outcome |
|---|---|---:|---|
| 2026-08-21 10:37:32 | `icons-fns` | 19,201 | ❌ stuck, 0 versions |
| 2026-08-21 11:30:20 → 11:38:00 | `emoji-list` | 1 | ✅ **published fine** (now at v2.0.3) |
| 2026-08-21 12:25:53 | `emoji-fns` | 3,824 | ❌ stuck, 0 versions |
| 2026-08-22 14:47:17 → 14:49:07 | `llm-task` | few | ✅ **published fine** (now at v0.2.0) |

`emoji-list` succeeding at 11:38 — in between the two failures — and `llm-task` succeeding
the following day rule out a registry-wide incident, a scope-level problem, and anything
local to my machine or network. The scope contains 77 packages; the only two that have
*ever* failed to publish are the two with huge export maps.

## What this is not

I checked these before filing:

- **Not a size limit.** `icons-fns` is **4.6 MB gzipped** (limit: 20 MB). Largest individual
  file is `deno.json` at **1.7 MB** (limit: 20 MB). `emoji-fns` is **3.05 MB gzipped**.
- **Not a quota.** Scope is at 77/100 packages; well under the weekly package-creation and
  publish-attempt limits.
- **Not an outage.** See timeline above.
- **Not slow types.** `Checking for slow types in the public API...` passes; the CLI proceeds
  past it and uploads successfully.
- **Not a client-side hang.** Retrying with a patch bump (6.0.0 → 6.0.1) reproduced the exact
  same behaviour, and the web UI independently confirms the task exists but has not started.

## CLI behaviour

The upload succeeds, then the CLI enters its status-polling loop and stays there. Output:

```
$ deno publish
Checking for slow types in the public API...
'@marianmeres/icons-fns' doesn't exist yet. Visit https://jsr.io/new?... to create the package
Waiting...
Package @marianmeres/icons-fns created
Visit https://jsr.io/auth?code=XXXX-XXXX to authorize publishing of @marianmeres/icons-fns
Waiting...
Authorization successful. Authenticated as Marian Meres
Publishing @marianmeres/icons-fns@6.0.0 ...
```

With `--log-level=debug`, this repeats indefinitely — a bodyless GET against the status
endpoint, roughly every few seconds, with no state change:

```
DEBUG RS - hyper_util::client::legacy::pool:269 - reuse idle connection for ("https", jsr.io)
DEBUG RS - h2::codec::framed_write:213 - send frame=Headers { stream_id: StreamId(685), flags: (0x5: END_HEADERS | END_STREAM) }
DEBUG RS - h2::codec::framed_read:404 - received frame=Headers { stream_id: StreamId(685), flags: (0x4: END_HEADERS) }
DEBUG RS - h2::codec::framed_read:404 - received frame=Data { stream_id: StreamId(685) }
DEBUG RS - h2::codec::framed_read:404 - received frame=Data { stream_id: StreamId(685), flags: (0x1: END_STREAM) }
```

## Hypothesis (speculative)

I have not verified this, but the correlation with export count rather than payload size
suggests the publishing pipeline does per-entrypoint work — module-graph resolution, doc-page
generation, symbol indexing — that scales with the number of declared exports, and that
19,201 entrypoints exceeds some worker time or memory ceiling.

Notably, the *module graph* here is trivial: the root module deliberately exports no icons
(just shared types and a renderer), and each icon module is a handful of lines importing a
single shared helper. The package is small and shallow; it just has a very large number of
tiny entrypoints. That, plus the fact that the failure scales with export count and not with
bytes, is what points at per-entrypoint cost.

## Package shape, for context

`icons-fns` wraps eight icon sets (Bootstrap, Phosphor, Lucide, Font Awesome, Heroicons,
Boxicons, Feather, Bytesize) as functions returning SVG strings. Each icon is its own module
imported by its own subpath, so that consumers without a bundler pull only the icons they
actually reference:

```ts
import { iconBsAlarm } from "@marianmeres/icons-fns/bootstrap/iconBsAlarm";
```

That design is what produces 19,199 icon subpaths, plus `.` and `./search` — 19,201 exports
across 19,203 files in `src/`. Since JSR only permits importing paths declared in `exports`,
there's no way to preserve per-icon subpath imports with a smaller export map.

## Asks

1. **Please clear the stuck publishing tasks** for `@marianmeres/icons-fns@6.0.0`,
   `@marianmeres/icons-fns@6.0.1`, and `@marianmeres/emoji-fns@2.0.1`, so those version
   numbers become usable again. Right now a patch bump just produces another stuck task.
2. **Is there a practical ceiling on the number of exports?** Nothing in
   [Quotas and limits](https://jsr.io/docs/quotas-and-limits) mentions export or file counts —
   only size limits, which I'm well inside. If a ceiling exists in practice, documenting it
   would help; if it's raiseable per-scope like other quotas, I'd like to request that.
3. **Please fail fast instead of hanging.** A task that can never complete should return an
   error to the CLI rather than leaving it polling indefinitely. That alone would have saved
   most of the time I spent on this.

I'm happy to bisect the threshold with a synthetic package (N trivial modules, N exports,
varying N) and report where it breaks, if that would be useful.

## Environment

```
deno 2.9.5 (stable, release, aarch64-apple-darwin)
v8 15.0.245.2-rusty
typescript 6.0.3
macOS 15 (Darwin 25.6.0), arm64
```

crowlKats:
## The bug

Publishing a package whose version `meta` exceeds ~8KB — in practice, a package with a few thousand exports whose modules lack module-level doc comments — never completes. The upload succeeds, the task is created, and the status page shows the task pending forever while the CLI polls indefinitely (#1505).

Root cause: `idx_package_versions_latest` (added in #1372) declares `INCLUDE (meta)`. Postgres stores INCLUDE columns inline in the b-tree index tuple and never TOASTs them out-of-line, so an index row is hard-capped at ~8191 bytes (after compression). `meta.entrypoints_without_docs` holds one entry per undocumented export, so large-export packages blow the cap and the `INSERT INTO package_versions` fails:

```
error returned from database: index row requires 13464 bytes, maximum size is 8191
```

That error is classified as retryable, so `publish_task` resets the task to `pending` and returns 500; Cloud Tasks retries the identical deterministic failure 30 times, then gives up silently, stranding the task in `pending` with no error ever surfaced.

Verified by publishing a synthetic 4,000-export package through the full pipeline in the test harness (fails exactly as above), and the same package with a module doc on every entrypoint (publishes fine — the trigger is meta size, not export count).

## The fix

- **Migration**: rebuild `idx_package_versions_latest` without `INCLUDE (meta)`. The covered lookup is a `LIMIT 1` lateral probe, so this costs one heap fetch per package. (If the transactional `DROP`+`CREATE` lock is a concern on prod, the replacement index can be pre-created `CONCURRENTLY` by hand before deploying.)
- **Cap `entrypoints_without_docs` at 100 entries**. The unbounded list also bloated every API response and Algolia record carrying meta. `all_entrypoints_docs` stays accurate: a capped list is non-empty iff the uncapped one was.
- **Fail fast**: `publish_handler` reads `X-CloudTasks-TaskRetryCount`; a retryable error on the 10th attempt marks the task `Failure` with a user-facing `internalError` instead of handing it back to the queue. Clients stop polling, and the version number becomes usable again (`create_publishing_task` only blocks on non-failure tasks).
- **Headroom**: the `registry-api-tasks` Cloud Run service gets 2Gi memory (analyzing a ~20k-module package peaks near the old 512Mi default, risking OOM mid-publish) and a 600s request timeout (Cloud Tasks abandons buffered dispatches at 10 minutes, so 600s is the useful ceiling).

With these changes a 19,201-export package publishes end-to-end in ~25s in the test harness, so no exports-count limit is needed.

## Tests

- `db::tests::package_version_with_large_meta` — inserts a version whose meta lists 20,000 entrypoints; fails on the old schema, passes with the migration
- `publish::tests::publish_many_undocumented_entrypoints` — full-pipeline publish with 150 undocumented exports succeeds and stores the capped list
- `publish::tests::publish_fails_fast_after_max_attempts` — below the cap the task returns to `pending` and the error propagates; on the final attempt it is marked `Failure` with `internalError`
- `analysis::tests::entrypoints_missing_docs_caps_list`

## Follow-up (ops, after deploy)

Requeue the stuck tasks for `@marianmeres/icons-fns@6.0.0`/`6.0.1` and `@marianmeres/emoji-fns@2.0.1` via the admin endpoint — their tarballs are still in the publishing bucket and should publish cleanly, and terraform apply for the Cloud Run changes.

Fixes #1505

### DIFF
--- a/api/migrations/20260824000000_package_versions_latest_index_row_size.sql
+++ b/api/migrations/20260824000000_package_versions_latest_index_row_size.sql
@@ -1,0 +1,12 @@
+-- Drop INCLUDE (meta) from the latest-version index. INCLUDE columns are
+-- stored inline in the b-tree index tuple and can never be TOASTed
+-- out-of-line, so an index row is hard-capped at ~8191 bytes (after
+-- compression). A `meta` above that cap (e.g. `entrypoints_without_docs`
+-- listing thousands of undocumented entrypoints) made the INSERT INTO
+-- package_versions fail, permanently wedging the publishing task in
+-- `pending` (jsr-io/jsr#1505). The lateral latest-meta lookup this covered
+-- is a LIMIT 1 probe, so losing the index-only scan costs a single heap
+-- fetch per package.
+DROP INDEX IF EXISTS idx_package_versions_latest;
+CREATE INDEX idx_package_versions_latest ON package_versions (scope, name, version DESC)
+  WHERE is_yanked = false AND version NOT LIKE '%-%';
--- a/api/src/analysis.rs
+++ b/api/src/analysis.rs
@@ -348,6 +348,14 @@ fn generate_score(
   }
 }
 
+/// Cap on how many undocumented entrypoints are recorded in
+/// `PackageVersionMeta`. The full list is unbounded (one entry per export) and
+/// `meta` is stored in the database and returned in API responses, so a
+/// package with tens of thousands of undocumented entrypoints would bloat
+/// both (jsr-io/jsr#1505). `all_entrypoints_docs` stays accurate: a capped
+/// list is non-empty iff the uncapped list was.
+pub(crate) const MAX_ENTRYPOINTS_WITHOUT_DOCS: usize = 100;
+
 fn entrypoints_missing_module_doc(
   documents_by_url: &ParseOutput,
   main_entrypoint: Option<ModuleSpecifier>,
@@ -394,6 +402,7 @@ fn entrypoints_missing_module_doc(
     missing.push(name);
   }
 
+  missing.truncate(MAX_ENTRYPOINTS_WITHOUT_DOCS);
   missing
 }
 
@@ -1415,6 +1424,25 @@ mod tests {
     );
 
     assert_eq!(missing, vec!["./utils".to_string()]);
+  }
+
+  #[test]
+  fn entrypoints_missing_docs_caps_list() {
+    let mut output = indexmap::IndexMap::new();
+    let mut exports_map = indexmap::IndexMap::new();
+    for i in 0..(super::MAX_ENTRYPOINTS_WITHOUT_DOCS + 50) {
+      output.insert(
+        deno_ast::ModuleSpecifier::parse(&format!("file:///m{i}.ts")).unwrap(),
+        make_document(make_js_doc(None), vec![]),
+      );
+      exports_map.insert(format!("./m{i}"), format!("./m{i}.ts"));
+    }
+    let exports = crate::db::ExportsMap::new(exports_map);
+
+    let missing =
+      super::entrypoints_missing_module_doc(&output, None, false, &exports);
+
+    assert_eq!(missing.len(), super::MAX_ENTRYPOINTS_WITHOUT_DOCS);
   }
 
   #[test]
--- a/api/src/api/admin.rs
+++ b/api/src/api/admin.rs
@@ -315,6 +315,7 @@ pub async fn requeue_publishing_tasks(req: Request<Body>) -> ApiResult<()> {
     let span = Span::current();
     let fut = publish_task(
       publishing_task_id,
+      0,
       buckets,
       license_store,
       registry,
--- a/api/src/api/package.rs
+++ b/api/src/api/package.rs
@@ -1113,6 +1113,7 @@ pub async fn version_publish_handler(
     let span = Span::current();
     let fut = publish_task(
       publishing_task.id,
+      0,
       buckets,
       license_store,
       registry_url,
--- a/api/src/db/tests.rs
+++ b/api/src/db/tests.rs
@@ -695,6 +695,57 @@ async fn create_package_version_and_finalize_publishing_task() {
 }
 
 #[tokio::test]
+async fn package_version_with_large_meta() {
+  let db = EphemeralDatabase::create().await;
+
+  let user_id = uuid::Uuid::default();
+  let scope_name: ScopeName = "scope".try_into().unwrap();
+  let package_name: PackageName = "testpkg".try_into().unwrap();
+  let version: Version = "1.0.0".try_into().unwrap();
+
+  db.create_scope(
+    &user_id,
+    false,
+    &scope_name,
+    user_id,
+    &ScopeDescription::default(),
+  )
+  .await
+  .unwrap();
+  let CreatePackageResult::Ok(package) =
+    db.create_package(&scope_name, &package_name).await.unwrap()
+  else {
+    unreachable!()
+  };
+
+  // Regression test for jsr-io/jsr#1505: `meta` used to be an INCLUDE column
+  // of idx_package_versions_latest, so a meta larger than Postgres' ~8KB
+  // b-tree index row cap (e.g. thousands of undocumented entrypoints) made
+  // this insert fail, permanently wedging the publishing task. The version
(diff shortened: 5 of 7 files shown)


crowlbot (approved):
Solid, well-diagnosed fix — nice work tracing the ~8KB cap to `INCLUDE (meta)` on a b-tree index and connecting it to the exact error message, and the four-part fix (drop the INCLUDE, cap the unbounded list, fail-fast on the queue's final retry, give the task Cloud Run service headroom) all read as necessary rather than a shotgun fix.

A few notes, nothing blocking:

- **`entrypoints_missing_module_doc`**: `missing.truncate(MAX_ENTRYPOINTS_WITHOUT_DOCS)` after appending — confirmed `all_entrypoints_docs: entrypoints_without_docs.is_empty()` is computed from the already-truncated vec, so the "capped list is non-empty iff the uncapped one was" claim holds (truncating a non-empty vec to 100 can't make it empty).
- **`publish_fails_fast_after_max_attempts`**: confirmed `MAX_PUBLISH_ATTEMPTS = 10` lines up with the `X-CloudTasks-TaskRetryCount` semantics (0 on the first attempt) — `run(MAX_PUBLISH_ATTEMPTS - 1)` is really the 10th attempt, and the test checks both sides of the boundary (still `Pending` below it, `Failure` with `internalError` on it).
- **Migration**: `DROP INDEX` + `CREATE INDEX` (both non-concurrent) on `idx_package_versions_latest` — you already flagged the lock concern with a `CONCURRENTLY` pre-create workaround in the description. Given `package_versions` is presumably a hot table in prod, is pre-creating it concurrently by hand the actual plan for this deploy, or is a brief lock acceptable here? Worth confirming before merge so it isn't lost in the PR body.
- Column/predicate list in the new index (`(scope, name, version DESC) WHERE is_yanked = false AND version NOT LIKE '%-%'`) matches the pre-#1372 definition exactly — good, this is purely dropping the `INCLUDE`, not a behavior change.

CI green. Approved.

PR #1507, opened 2026-08-24
fix: don't count re-export references as undocumented symbols

### LINKED ISSUE #988: % of exported symbols is incorrect
https://jsr.io/@panva/jose/doc has every single exported symbol documented (which I think shouldn't be needed, see #975), and yet https://jsr.io/@panva/jose/score claims that only 61% of exported symbols are documented.

crowlKats:
Fixes #988

## Problem

The "% of exported symbols documented" score was far too low for packages that re-export symbols — e.g. `@panva/jose` (fully documented) showed 61%, and `@md/ensure-error` showed exactly 50%.

When a module re-exports a symbol (`export { x } from "./x.ts"`) and the target declaration is also present elsewhere in the doc output (typical for packages whose main entrypoint re-exports symbols that also have their own entrypoints), deno_doc represents the re-export as a `Reference` declaration with no JSDoc of its own — the documentation lives at the target declaration. `percentage_of_symbols_with_docs` counted those references as undocumented symbols, so every re-exported symbol was counted twice: once documented (at its definition) and once undocumented (the reference).

## Fix

Skip `Reference` declarations in `percentage_of_symbols_with_docs`. This never drops a symbol from the count: deno_doc's parser inlines or removes dangling references during `resolve_references_for_nodes`, so a surviving `Reference` always points at a declaration that exists elsewhere in the output and is counted there.

Includes a regression test (fails at 50% without the fix, passes at 100% with it).

Note: already-published versions have the wrong percentage stored in their version meta, so existing scores only correct on re-analysis or a new publish.

### DIFF
--- a/api/src/analysis.rs
+++ b/api/src/analysis.rs
@@ -427,6 +427,14 @@ fn percentage_of_symbols_with_docs(documents_by_url: &ParseOutput) -> f32 {
         .iter()
         .filter(|decl| {
           decl.declaration_kind != deno_doc::node::DeclarationKind::Private
+            // Skip re-export references: their docs live at the target
+            // declaration, which is counted where it is defined. Counting the
+            // (always doc-less) reference too would double-count every
+            // re-exported symbol as undocumented (jsr-io/jsr#988).
+            && !matches!(
+              &decl.def,
+              deno_doc::node::DeclarationDef::Reference(_)
+            )
         })
         .collect();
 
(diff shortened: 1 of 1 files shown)


crowlbot (approved):
Clean, well-scoped fix — the double-counting bug is exactly as described: a `Reference` declaration is a pointer to docs that live (and get counted) at the target, so counting it too as an undocumented symbol double-penalizes every re-exported symbol.

Checked the `has_overloads` interaction: `non_private_decls` now excludes `Reference` too, which only tightens the "all non-private decls are Function variants" check — a symbol mixing a `Reference` with a real `Function` decl no longer gets misclassified as *not* having overloads just because of the reference. No regression there.

Regression test constructs the exact re-export shape (mod.ts referencing hello from util.ts) and asserts 100% instead of 50% — good, it fails without the fix as claimed. CI green. Approved.

PR #1508, opened 2026-08-24
feat: align npm tarball paths with registry.npmjs.org layout

### LINKED ISSUE #405: Align JSR NPM package tarball paths with packages.npmjs.org
Some private registries require that upstreams have an identical tarball path layout as the upstream NPM registry. JSR does not currently do this, which means JSR is unable to be used as an upstream for these private registries.

This is known to affect Google Artifact Registry and JFrog Artifactory.

crowlKats:
Serves npm compatibility tarballs at the same path layout that registry.npmjs.org uses:

```
https://npm.jsr.io/@jsr/luca__cases/-/luca__cases-1.0.0.tgz
```

instead of the revisioned layout (`https://npm.jsr.io/~/11/@jsr/luca__cases/1.0.0.tgz`). npm proxies such as JFrog Artifactory and Google Artifact Registry construct tarball paths by npm convention instead of reading `dist.tarball`, so the old layout made JSR unusable as an upstream for them.

Fixes #405

## How it works

- The new layout applies to tarball revision 12 and later; `NPM_TARBALL_REVISION` is bumped to 12 so the existing `npm_tarball_enqueue` task can backfill all previously published versions at the new path.
- `generate_npm_version_manifest` advertises the npm-layout URL only for versions whose tarball is at revision >= 12, so a manifest never points at an object that doesn't exist yet. Until the backfill completes, older versions keep their legacy URLs.
- Legacy `~/{revision}/` objects are kept forever, so tarball URLs recorded in existing lockfiles keep resolving.
- Because the npm-layout path no longer encodes the revision, a rebuild of an existing version overwrites its tarball in place; the rebuild task now purges the tarball URL from the CDN cache along with the manifest.
- The lb download analytics track both layouts.

## Tradeoff: tarball URLs no longer encode the revision

The npm path convention has no room for a revision segment — proxies construct the path from just the package name and version. That means one URL per version, forever:

- A future revision bump that rebuilds **already-published** versions would change content behind URLs pinned in lockfiles, causing `EINTEGRITY` failures. After this ships, revision bumps should only apply to newly published versions (the same invariant registry.npmjs.org operates under).
- This PR's own 11→12 bump is safe: the rebuild pipeline is unchanged, and old versions keep serving their revision-11 URLs until rebuilt at 12.
- The docs' previous "tarball URLs are immutable" promise is rewritten accordingly in `npm-compatibility.md`.

## Deployment

After deploy, run `POST /tasks/npm_tarball_enqueue` (same runbook as previous revision bumps) to backfill all existing versions at revision 12. The registry is fully consistent before, during, and after the backfill — proxy compatibility for a given older version starts once it is rebuilt.

## Testing

- `cargo test` (206 passed), `cargo clippy`, `cargo fmt`
- `deno test` in `lb/` including new `analytics_test.ts`
- New unit test covering both path layouts in `s3_paths.rs`


crowlbot (approved):
Reviewed the mechanics behind the tradeoff called out in the PR description — verified they actually hold in the code, not just the description:

- `generate_npm_version_manifest` passes the per-version stored `version.npm_tarball_revision` into `npm_tarball_url`, so the manifest only advertises the npm-layout URL once *that version's* tarball is actually rebuilt at revision 12 — confirmed this isn't gated on the global `NPM_TARBALL_REVISION` constant, which would have been wrong (would advertise URLs for objects that don't exist yet during the backfill window).
- `npm_tarball_path`'s revision branch (`>= FIRST_NPM_LAYOUT_TARBALL_REVISION` keeps the legacy `~/{revision}/` layout, `<` doesn't) matches the "legacy objects kept forever" claim, and the new unit test covers both branches plus the URL helper.
- The npm-layout regex in `trackNPMDownload` uses a backreference (`\1__\2`) to require the directory segment and filename segment agree — the "ignores non-tarball paths" test includes a mismatched case (`luca__cases/-/other__pkg-...`) that would false-positive without the backreference, so that guard is actually exercised.
- `npm_tarball_build_handler`'s rebuild now purges the tarball URL from CDN alongside the manifest — necessary since revision 12+ writes are in-place overwrites rather than new revisioned objects, and this is the one behavior that changes for existing infra (revision bumps after this ship must not touch already-published versions, as documented).

The revision-not-in-path tradeoff is real but well-contained: legacy objects are permanent, the 11→12 migration itself is safe by construction (nothing already at revision 12 exists yet), and the future invariant is documented in both the PR and (per the diff) `npm-compatibility.md`. CI green. Approved.

PR #1509, opened 2026-08-24
fix: unbreak support email delivery, and sort out the ticket page's controls

crowlKats:
Two fixes to the support ticket system, both found after #1501 went live.

---

## 1. Queued email was never delivered

Inbound support email opens a ticket and records the automatic reply on it, but the reply itself is never sent. The ticket dashboard shows *"Automatic reply: we received your email and opened this ticket."* while the sender's inbox stays empty.

### Cause

Delivering queued email moved to `/tasks/send_email` in #1501, which runs on the `registry-api-tasks` service. Only `registry-api` was ever given `POSTMARK_TOKEN`, so on the tasks service `email_sender` is `None` and nothing can be sent. Both services already run as the same service account, so the secret's IAM binding covered this — only the env var was missing.

The missing credential on its own would have been loud. What made it silent was the handler's own guard:

```rust
let Some(email_sender) = email_sender else {
    error!("received an email delivery task but no email sender is configured");
    return Ok(...OK);   // 200 → Cloud Tasks considers it delivered
};
```

It was meant to avoid retrying forever in a deployment with no email configured. Instead it turned a misconfiguration into permanent, invisible mail loss: Cloud Tasks read the 200 as success, and the delivery row was left pending with nothing to mark it.

### Fix

- `POSTMARK_TOKEN` added to `registry_api_tasks`.
- A missing sender is now recorded against the delivery like any other failure. It retries if the configuration is fixed, and is abandoned after the usual number of attempts with a readable `last_error` instead of disappearing. The handler no longer special-cases it.
- Regression test asserts a delivery with no sender stays pending, records the error, and is still visible to the sweeper.

### After deploying

Deliveries stranded by this are still pending (`sent_at IS NULL`, `abandoned_at IS NULL`), so the sweeper will pick them up within five minutes and send them — **a burst of belated auto-replies** to everyone who wrote in during the window. Probably what you want, since they were never acknowledged. To see what is queued first:

```sql
SELECT id, to_address, subject, attempts, last_error, created_at
FROM email_deliveries
WHERE sent_at IS NULL AND abandoned_at IS NULL
ORDER BY created_at;
```

If any are old enough to be confusing, mark them abandoned before deploying:

```sql
UPDATE email_deliveries
SET abandoned_at = now(), last_error = 'stranded before the tasks service had a Postmark token'
WHERE sent_at IS NULL AND abandoned_at IS NULL AND created_at < '<cutoff>';
```

---

## 2. The ticket page's status controls and timeline

The status dropdown and the close button both changed the status, so there were two controls for one job. The dropdown also applied on `change` — worse than it sounds, because a native select fires `change` as the keyboard moves through its options, so arrowing past "spam" silently marked the ticket as spam.

Now there is one control per job:

- The **footer** keeps a single close/reopen button — the common action, flipping label with the state.
- The **full status control moves to the header**, replacing the read-only badge, where you would look to change it. Picking a value selects it and nothing more; a separate **Update** button applies it, disabled until the selection actually differs.

Also in this commit:

- **Timeline events** were bare paragraphs sitting after the message cards with no visual relationship to them. They now carry the coloured marker of the status they moved to, sized and muted as timeline entries. Audit rows written before the status enum recorded `{closed: true}` rather than a status, and those rows are still in the log, so they are read too instead of rendering as "a new value".
- **The automatic acknowledgement was labelled "Staff"** — the label came from the message direction, and the acknowledgement is outbound, so it implied a person had replied. It is now derived from the author, and a JSR-generated message is marked "Automatic".
- **"unverified sender" was styled as an alarm.** It sits on nearly every inbound message (SPF does not survive a forwarding hop, which is how mail reaches the ticket system), and a warning that is always on stops being read. Muted to "unverified", explanation still on hover.

> **The UI half is not visually verified.** Local dev 404s on every route including `/`, so I could not get a screenshot before pushing. It type-checks and lints; the layout wants a human eye.

---

## Testing

230 API tests pass. Clippy `-D warnings`, `cargo fmt`, `sqlx prepare --check`, `terraform fmt`, `deno task lint:frontend`, `deno fmt` and license headers all clean.

### DIFF
--- a/api/src/api/hooks.rs
+++ b/api/src/api/hooks.rs
@@ -888,6 +888,43 @@ mod integration {
   }
 
   #[tokio::test]
+  async fn a_delivery_is_not_lost_when_no_sender_is_configured() {
+    let t = TestSetup::new().await;
+    let db = t.ephemeral_database.database.as_ref().unwrap();
+
+    let id = db
+      .enqueue_email(crate::db::NewEmailDelivery {
+        to_address: "someone@example.com".to_owned(),
+        subject: "subject".to_owned(),
+        body_text: "text".to_owned(),
+        body_html: "<p>html</p>".to_owned(),
+        message_id: None,
+        in_reply_to: None,
+        reference_ids: vec![],
+      })
+      .await
+      .unwrap();
+
+    // The queue dispatches to a service that has no Postmark credential. This
+    // used to be swallowed with a 200, which marked the delivery done without
+    // ever sending it. It must be recorded as a failure instead, so the mail is
+    // still there to send once the configuration is fixed.
+    let terminal = crate::emails::deliver(db, None, id).await.unwrap();
+    assert!(!terminal, "a missing sender must not end the delivery");
+
+    let pending = db.get_pending_email_delivery(id).await.unwrap().unwrap();
+    assert_eq!(pending.attempts, 1);
+    assert!(pending.last_error.unwrap().contains("no email sender"));
+    assert!(pending.sent_at.is_none());
+
+    // And it is still visible to the sweeper, so it goes out after a fix.
+    assert_eq!(
+      db.list_stale_email_deliveries(0, 10).await.unwrap(),
+      vec![id]
+    );
+  }
+
+  #[tokio::test]
   async fn a_delivery_is_abandoned_after_too_many_failures() {
     let t = TestSetup::new().await;
     let db = t.ephemeral_database.database.as_ref().unwrap();
--- a/api/src/emails/mod.rs
+++ b/api/src/emails/mod.rs
@@ -366,7 +366,7 @@ pub async fn enqueue(
       }
     }
     None => {
-      if let Err(err) = deliver(db, email_sender, id).await {
+      if let Err(err) = deliver(db, Some(email_sender), id).await {
         tracing::error!(
           delivery_id = %id,
           "failed to send email inline: {:?}",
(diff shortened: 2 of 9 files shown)


crowlbot (approved):
Good catch and a clean fix. Verified the two claims in the description:

- `registry_api_tasks` and `registry_api` both run as `google_service_account.registry_api` (`terraform/cloud_run_api.tf:80,243`), and `postmark_token`'s only IAM binding is already scoped to that same account — so this genuinely is just the missing env var, no IAM change needed.
- `deliver`'s new `None` branch goes through `record_email_delivery_failure` like any other failure, so it's bounded by the same `MAX_EMAIL_ATTEMPTS` and shows up for the sweeper — no more silent 200-swallowing. The regression test checks all three: still pending, `last_error` populated, and visible to `list_stale_email_deliveries`.

The "After deploying" section is exactly right to call out — a live deploy here will trigger a burst of belated auto-replies for anything stranded, which is probably desired but is a real operational step, not just documentation. Good that it's flagged rather than left implicit.

CI green. Approved.

crowlbot (approved):
Reviewed the second commit (frontend ticket-page fixes) — the code matches every claim in the description:

- `TicketStatusControl` only applies on the explicit button click (`onChange` just updates local `selected` state), and the button stays `disabled` until `selected !== status` — confirms the arrow-key-through-a-native-select-fires-`change` bug is actually fixed, not just described.
- The old duplicate select+close-button pair in `TicketMessageInput` is gone; only the single close/reopen button remains, flipping both label and `button-primary`/`button-danger` class on `closed`.
- `AuthorRole` branches on `author.kind === "system"` rather than message direction, so the automatic acknowledgement is correctly labelled "Automatic" instead of implying a human replied.
- `StatusChange`'s fallback (`meta.status ?? (typeof meta.closed === "boolean" ? ... : undefined)`) does read the legacy pre-status-enum audit rows, with a graceful `"a new value"` fallback if neither is present.
- "unverified sender" → muted "unverified" styling change matches the stated reasoning (always-on alarms stop being read) and keeps the explanation on `title`/hover.

CI green. Approved.

PR #1510, opened 2026-08-25
fix(api): read the forwarder's SPF and DKIM result, not Postmark's

crowlKats:
Nearly every inbound support message is marked as coming from an unverified sender, and the badge saying so has stopped meaning anything.

## Cause

Support mail is forwarded from the Google Workspace inbox to Postmark, so by the time Postmark performs its SPF check the message is arriving from Workspace's servers rather than the sender's. SPF fails for almost everyone, including senders whose domain is configured perfectly. Verification therefore said "unverified" on mail that had in fact passed both checks — and a warning that is always on is one nobody reads.

## Fix

Workspace records the real result in `Authentication-Results` when it first receives the message, and that header survives the hop. Where the trusted upstream has left its results, they are now used in preference to Postmark's; without them nothing changes.

Which upstream to trust is configuration rather than a constant: `INBOUND_TRUSTED_AUTHSERV_ID`, set to `mx.google.com` in terraform. It is a trust anchor and belongs somewhere it can be seen and changed. Left unset, the header is never consulted and only Postmark's own check applies — so this is inert until deployed with the variable set.

## Trusting a header the sender can also write

That is the whole risk in this change, so:

- **Only the first `Authentication-Results` naming the trusted server is read.** Each hop prepends its own headers, so the trusted server's sit above anything the sender embedded — including a header forged to carry that server's name. Reading any but the first would let a sender award themselves a pass. There is a test that does exactly that and expects it to fail.
- **Parenthesised comments are stripped before parsing.** They are free text that routinely contains both `;` and strings like `spf=pass`, so parsing without removing them can be steered by the sender. Also tested.
- **Method names are matched whole**, so `dkim-atps` is not read as `dkim`.

This does depend on Postmark preserving header order in the webhook payload, which it does, and which the ordering argument above rests on. Worth knowing as an assumption.

## Testing

236 API tests pass, 6 of them new here: the forwarded-mail case, a genuine upstream failure overriding a Postmark pass, the forged header, results from an untrusted server, comment injection, whole-name matching, and an authserv-id carrying a version.

Clippy `-D warnings`, `cargo fmt`, `sqlx prepare --check`, `terraform fmt`, `deno task lint:frontend` and license headers all clean.

### DIFF
--- a/api/src/api/hooks.rs
+++ b/api/src/api/hooks.rs
@@ -42,6 +42,19 @@ use super::ApiError;
 /// The password Postmark authenticates with. `None` disables inbound handling.
 pub struct PostmarkWebhookPassword(pub Option<String>);
 
+/// The `authserv-id` of the mail server that receives support mail before it
+/// reaches Postmark — `mx.google.com` for a Google Workspace inbox.
+///
+/// Mail forwarded from that inbox arrives at Postmark from the forwarder's own
+/// servers, so the SPF check Postmark performs is against the wrong sender and
+/// fails for almost everything. The forwarder recorded the real result when it
+/// first received the message, and that record survives the hop, so it is what
+/// we read instead — but only for the server named here, since anyone can write
+/// an `Authentication-Results` header claiming whatever they like.
+///
+/// `None` means no upstream is trusted and only Postmark's own check is used.
+pub struct InboundTrustedAuthservId(pub Option<String>);
+
 /// Files above this are dropped rather than stored. Postmark caps an inbound
 /// message at 35 MB in total; this is well under that, and comfortably above
 /// anything a support conversation legitimately needs.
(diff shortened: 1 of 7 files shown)


crowlbot (approved):
Careful, well-threat-modeled fix — reviewed the header-trust logic specifically, since that's where a mistake here would matter:

- **Ordering claim checked**: `trusted_auth_results` uses `.iter().find(...)`, taking the *first* matching `Authentication-Results` in header order — correct, since each hop prepends, so the trusted forwarder's own header (added last, closest to the top of the raw message) sits above anything the original sender embedded, including a forged one naming the same `authserv-id`. The "forged header" test exercises exactly this (attacker's header first in the test array would be wrong; real header order has the trusted one first and it wins).
- **Comment-stripping is exercised meaningfully**: the test value embeds both a `;` and the literal string `spf=pass` inside a `(...)` comment attached to a real `spf=fail` result, and asserts the real (failing) result is what's read. Not a token test.
- **Method-name matching**: `dkim-atps=pass; dkim=fail` correctly resolves to `fail` for `dkim` — the whole-token match via `split_once('=')` on `dkim-atps` extracts `name="dkim-atps"`, which fails `eq_ignore_ascii_case("dkim")`, so it isn't confused with a bare `dkim=` result.
- **No missing-service-config repeat of #1509**: `hooks_router()` (and thus the inbound webhook) lives on `api_router()`, i.e. `registry-api`, and `INBOUND_TRUSTED_AUTHSERV_ID` is added to `local.api_envs`, which both Cloud Run services already consume — so it reaches the service that needs it. (It's also harmlessly present but unused on the tasks service.)
- **Fallback correctness**: no trusted-id configured, or a header present but from an untrusted server, both correctly fall through to the original `X-Spam-Test`-based check rather than silently passing.

The one thing I can't independently verify from here — that Postmark actually preserves header order in the inbound webhook payload — is exactly the assumption the PR calls out explicitly, which is the right way to handle it.

CI green. Approved.

PR #1511, opened 2026-08-25
feat: allow scope admins to delete recently published package versions

### LINKED ISSUE #899: ability to delete of published version under certain circumstances

crowlKats:
Closes #899

This turns the existing staff-only version deletion endpoint into a feature scope admins can use, under strict conditions that keep the registry's immutability guarantees meaningful. A scope admin may delete a version only if **all** of the following hold:

- the version was published **less than 24 hours ago**,
- the version has **fewer than 10 downloads** (summed over `version_download_counts_24h`, all kinds),
- **deleting it does not strand any dependent**: every distinct dependency constraint referencing the package (both `jsr` and npm-mapped names) is matched against the version with semver range semantics, and a matching constraint only blocks deletion if the version is the *sole remaining non-yanked version* satisfying it. Dependents that can resolve to a sibling version are not broken by the deletion, so a broad range (or a `*` wildcard) does not pin every version of a package. Unparseable and dist-tag constraints fail closed.

Staff with sudo bypass the age and download conditions, but the dependents check applies to everyone.

### API changes

- `DELETE /api/scopes/:scope/packages/:package/versions/:version` now requires scope admin access (or staff + sudo) instead of bare staff access, matching what `api.yml` already documented.
- New error codes `deleteVersionTooOld` and `deleteVersionTooManyDownloads`; deleting a nonexistent version now returns 404 instead of succeeding silently.
- Cleanup gaps in the existing delete path are fixed:
  - npm tarball objects in the npm bucket are now deleted (previously orphaned forever),
  - Algolia is refreshed after deletion so score/latest-version stay accurate,
  - the audit log records the actual `is_sudo` instead of a hardcoded `true`.
- The publishing task of a deleted version still survives, so the version number stays burned and cannot be re-published (unchanged behavior).

### Frontend

- The Delete button on the versions tab now shows for scope admins on versions younger than 24 hours (always for staff with sudo), with a confirmation dialog.
- A note under the version list explains the deletion conditions to admins. The UI gate is best-effort; the API enforces the real policy.

### Docs

`immutability.md`, `packages.md` (new "Deleting a version" section), `faq.md`, `npm-compatibility.md`, and `api.md` now describe the narrow exception instead of stating versions can never be deleted.

### Caveats

- Download counts are batch-scraped, so the threshold lags real-time; it is a heuristic, not airtight.
- Edge caches serve version files, `_meta.json`, and tarballs as `immutable`, so a deleted version's artifacts may be served until TTL expiry (pre-existing trait of the staff delete path).
- There is a pre-existing race between the dependents check and the delete (a dependent publishing in between); not closed here.

### Testing

The `delete_version` test was rewritten to cover: 404 on nonexistent versions, dependents blocking (for admins and staff+sudo), deletion when constraints don't match the version (with npm tarball bucket assertions), the download threshold in both directions, the age limit with staff bypass, authorization for members/non-members/staff-without-sudo, and that re-publishing a deleted version stays blocked.

### DIFF
--- a/api/.sqlx/query-1037aba623c2c6b3a8100adb4315159003c9effc17415fa70a0038b44551a5c6.json
+++ b/api/.sqlx/query-1037aba623c2c6b3a8100adb4315159003c9effc17415fa70a0038b44551a5c6.json
@@ -1,0 +1,17 @@
+{
+  "db_name": "PostgreSQL",
+  "query": "UPDATE package_versions SET created_at = $4 WHERE scope = $1 AND name = $2 AND version = $3",
+  "describe": {
+    "columns": [],
+    "parameters": {
+      "Left": [
+        "Text",
+        "Text",
+        "Text",
+        "Timestamptz"
+      ]
+    },
+    "nullable": []
+  },
+  "hash": "1037aba623c2c6b3a8100adb4315159003c9effc17415fa70a0038b44551a5c6"
+}
--- a/api/.sqlx/query-1ed3f4e457d4d53e3915e0190d168be79929b750cf5a34a76cda77b5603745d7.json
+++ b/api/.sqlx/query-1ed3f4e457d4d53e3915e0190d168be79929b750cf5a34a76cda77b5603745d7.json
@@ -1,0 +1,23 @@
+{
+  "db_name": "PostgreSQL",
+  "query": "SELECT DISTINCT dependency_constraint\n      FROM package_version_dependencies\n      WHERE (dependency_kind = 'jsr' AND dependency_name = $1)\n        OR (dependency_kind = 'npm' AND dependency_name = $2)",
+  "describe": {
+    "columns": [
+      {
+        "ordinal": 0,
+        "name": "dependency_constraint",
+        "type_info": "Text"
+      }
+    ],
+    "parameters": {
+      "Left": [
+        "Text",
+        "Text"
+      ]
+    },
+    "nullable": [
+      false
+    ]
+  },
+  "hash": "1ed3f4e457d4d53e3915e0190d168be79929b750cf5a34a76cda77b5603745d7"
+}
--- a/api/.sqlx/query-2a0256f81593416d76784e5de6ab8d1d5507965cbeffacbe774deae5ff3cc4e5.json
+++ b/api/.sqlx/query-2a0256f81593416d76784e5de6ab8d1d5507965cbeffacbe774deae5ff3cc4e5.json
@@ -1,0 +1,72 @@
+{
+  "db_name": "PostgreSQL",
+  "query": "SELECT scope as \"scope: ScopeName\", name as \"name: PackageName\", version as \"version: Version\", revision, sha1, sha512, size, updated_at, created_at FROM npm_tarballs\n      WHERE scope = $1 AND name = $2 AND version = $3",
+  "describe": {
+    "columns": [
+      {
+        "ordinal": 0,
+        "name": "scope: ScopeName",
+        "type_info": "Text"
+      },
+      {
+        "ordinal": 1,
+        "name": "name: PackageName",
+        "type_info": "Text"
+      },
+      {
+        "ordinal": 2,
+        "name": "version: Version",
+        "type_info": "Text"
+      },
+      {
+        "ordinal": 3,
+        "name": "revision",
+        "type_info": "Int4"
+      },
+      {
+        "ordinal": 4,
+        "name": "sha1",
+        "type_info": "Text"
+      },
+      {
+        "ordinal": 5,
+        "name": "sha512",
+        "type_info": "Text"
+      },
+      {
+        "ordinal": 6,
+        "name": "size",
+        "type_info": "Int4"
+      },
+      {
+        "ordinal": 7,
+        "name": "updated_at",
+        "type_info": "Timestamptz"
+      },
+      {
+        "ordinal": 8,
+        "name": "created_at",
+        "type_info": "Timestamptz"
+      }
+    ],
+    "parameters": {
+      "Left": [
+        "Text",
+        "Text",
+        "Text"
+      ]
+    },
+    "nullable": [
+      false,
+      false,
+      false,
+      false,
+      false,
+      false,
+      false,
+      false,
+      false
+    ]
+  },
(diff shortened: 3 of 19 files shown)


crowlbot (approved):
Careful design for a policy change that touches the registry's core immutability guarantee. Verified the parts that matter most:

- **The staff+sudo bypass gap is real and correctly patched.** `check_scope_admin_access` (pre-existing, unmodified) returns `sudo=false` whenever the caller *is* a scope admin member, even if they're staff with `sudo` on — it only reports `sudo=true` for staff bypassing the admin-membership check entirely. The `let sudo = sudo || (user.is_staff && iam_sudo);` line is what makes "staff with sudo bypasses age/downloads" actually true for a staff member who also happens to be a scope admin, not just for staff who aren't scope members. Without it this class of caller would incorrectly be held to the 24h/download limits.
- **The dependents check fails closed** on both an unparseable constraint and a dist-tag constraint (`range() == None`) — both treated as blocking rather than permissive, which is the right default for something checked before an irreversible delete.
- **Order of operations is correct**: constraints are checked before delete, npm tarball rows are read before the delete cascades them away, and `delete_package_version` now returns whether a row was actually affected (closing the "delete a nonexistent version silently succeeds" gap) rather than assuming success.
- **Download-count reasoning holds together**: `version_download_counts_24h` is a rolling 24h table, and since only versions ≤24h old are eligible for non-sudo deletion, summing that table for the version's lifetime download count is coherent (not truncating older activity that would matter for an older version).
- Test coverage matches every claim: 404 on nonexistent, dependents blocking for both admin and staff+sudo, admin/non-member/staff-without-sudo authorization split, the download threshold in both directions, age limit with staff bypass, npm tarball bucket deletion verified before/after, and that the burned version number still blocks republishing after delete.

The caveats section (download-count lag, edge-cache TTL, dependents-check race) is honest about what this doesn't close rather than overselling the guarantee — appropriate given this is explicitly a heuristic gate, not a hard guarantee.

CI green. Approved.

KnorpelSenf:
> * **no published package version depends on it**: the check is version-level, matching every distinct dependency constraint referencing the package (both `jsr` and npm-mapped names) against the version with semver range semantics. Unparseable and dist-tag constraints fail closed.

Just to make sure I understand this correctly:

- if another package depends on `pkg@^1.0.0` then this would prevent any `1.x` version of `pkg` from being deleted immediately after publish, but any `2.x` would be unaffected
- the easiest strategy to disable package deletion for another person's package would be to publish a new dummy package depending on `pkg@*`
- the feature could be broken by https://socket.dev/blog/when-everything-becomes-too-much#Unintended-Consequences

I genuinely mean that I'm trying to understand this, I'm not saying this is a bad idea (let alone that I have a better one). Deleting is needed and always a bit ugly so details matter, hence my interest.

crowlKats:
Good points, and your understanding of the initial implementation was correct on all three counts. This made me reconsider the semantics, and I've now changed the check (4f36300f) to address exactly this:

A dependent constraint now only blocks deletion if the version being deleted is the **sole remaining non-yanked version** satisfying that constraint. If dependents can resolve to a sibling version, deleting doesn't actually break them, so it's allowed.

For your scenarios:

- a dependent on `pkg@^1.0.0` no longer pins all of `1.x` — any `1.x` version can be deleted as long as at least one other (non-yanked) `1.x` version remains. Only the last one standing is protected.
- a dummy package depending on `pkg@*` blocks nothing, since `*` is satisfiable by any other version. Same for an `everything`-style package.

The remaining protections are: exact pins (`pkg@1.2.3` blocks deleting `1.2.3`), sole-match ranges (you can't delete the only version a range can resolve to), and lockfile pins to the deleted version, which are approximated by the "fewer than 10 downloads" condition rather than the dependents check.

crowlbot (approved):
This is exactly the fix the dependents-check discussion was pointing at, and it's a clean one: narrowing the block to "is this version the *only* one that still satisfies the constraint" rather than "does the constraint match this version at all" directly closes the griefing vector — a dependent with a `*`/broad-range constraint can no longer pin every version of a package hostage forever, only the last one actually needed to satisfy it.

Traced through the yank/unyank test: excluding yanked versions from `remaining_versions` is the right call (yanked versions aren't resolved by ranges under normal resolution), and the test's flip (yank 1.2.3 → 1.3.0 blocked, unyank → 1.3.0 deletable) exercises exactly that boundary. Dist-tag and unparseable constraints still fail closed unconditionally, so this narrows the *good* case without weakening the conservative fallback. Docs, error message, and the OpenAPI description are all updated to match the new semantics accurately.

CI's `test`/`check` jobs (the ones covering this Rust code) were still running as I reviewed — everything else passed. Approving on the logic; let me know if either of those comes back red and I'll take another look.

mannsion:
I like this change, but it really only protects for accidental publishes etc, and doesn't back date previous packages that already passed 24 hours.

It would be nice to expand this to something like

# A scope admin should be able to permanently delete a version when all of these are true:

- It has remained continuously yanked for a configurable period, perhaps 90 days.
- It has had zero downloads during that period.
- No published dependency would be stranded; the existing “another non-yanked version satisfies the constraint” logic applies.
- It is not referenced by an active dist-tag.
- Eligibility is rechecked atomically immediately before **deletion.**

KnorpelSenf:
I don't think 90 days is nearly enough. I often revisit a project after 1-2 years and it would be pretty bad if the deps are gone then. 

Besides, if you have to wait for months before you're allowed to delete a package, you'd have to find a better solution than deleting the version anyways. So which problem are you trying to solve exactly by letting developers delete their old versions?

mannsion:
I'm not sure what the correct thing to do is, but I do think the UI can be improved a lot.  If a yanked a package that was never released in a version 1 is not wanted by it's creator, it or it's scope should be able to be archived or abandoned, even if it's still available for people to resolve and download things for.

Make it so a user can archive a scope and or packages or something and then they're disabled and don't count towards your scope count.  Where they can be un-archived and resurrected etc if desired.

PR #1514, opened 2026-08-25
fix(frontend): return to the ticket after signing in to claim it

crowlKats:
The sign-in link on an unclaimed ticket went to a bare `/login`, so somebody following the claim link out of their auto-reply signed in and landed on the home page. The ticket they were trying to reach was gone — and so was the claim token that would have let them reach it again, since it lives only in that link and in the email it came from. Getting back meant digging the email out again.

The link now carries `?redirect=` back to this page, query string included, the same way the header's sign-in link does (`components/Header.tsx`). The reporter returns to the ticket already signed in, with the claim button waiting.

Two details:

- **The return path is passed in from the route** rather than read from `location`, which does not exist while the island is server-rendered.
- **The claim token rides along in the redirect**, which is what makes the flow work at all. `sanitize_redirect_url` keeps the query string and forces a same-origin relative path, so it survives intact and cannot be pointed elsewhere. It does mean the token is briefly stored in `oauth_states.redirect_url` — no worse than it already being in the browser's history for that page, and those rows are cleaned up daily by `clean_oauth_states`.

## Testing

`deno task lint:frontend` and `deno fmt` clean.

Not verified in a browser — local dev still 404s on every route for me, which is a separate problem. The change is small and follows the existing header pattern, but the flow itself is worth clicking through once after deploy: open a claim link signed out, sign in, and check you land back on the ticket with the claim button rather than on the home page.

### DIFF
--- a/frontend/islands/ClaimTicket.tsx
+++ b/frontend/islands/ClaimTicket.tsx
@@ -6,10 +6,14 @@ import { api, path } from "../utils/api.ts";
 /// rather than an account. Claiming binds it to the signed-in account, so it
 /// shows up alongside their other tickets instead of only in their inbox.
 export function ClaimTicket(
-  { ticketId, claimToken, signedIn }: {
+  { ticketId, claimToken, signedIn, returnTo }: {
     ticketId: string;
     claimToken: string;
     signedIn: boolean;
+    /// Where to come back to after signing in — this page, claim token and all.
+    /// Passed in rather than read from `location`, which does not exist when
+    /// this is rendered on the server.
+    returnTo: string;
   },
 ) {
   const [error, setError] = useState<string | null>(null);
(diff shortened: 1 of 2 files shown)


crowlbot (approved):
Traced this through end-to-end since you flagged not being able to click through it — the claims check out:

- `url` as a page prop and the `pathname + search` redirect-target pattern already exist elsewhere (`Header.tsx` does the same, just also including `url.hash`, which doesn't matter here since the claim token is a query param, not a fragment).
- `sanitize_redirect_url` (backend, pre-existing, tested) does keep the query string and forces same-origin — `sanitize_redirect_url("//evil.com//bar?foo=bar") == "/bar?foo=bar"` in its own test suite — so the claim token really does survive the round trip and can't be redirected off-site.
- `redirect_url` is stored in `oauth_states` and cleaned up by the existing `clean_oauth_states` task/scheduler, matching the "no worse than browser history, and it's swept daily" reasoning.

So the plumbing this relies on is all real and already proven by the existing login-redirect flow; this PR is just wiring the same pattern into `ClaimTicket`, not inventing a new one. Given that, and that CI (lint/fmt/type-check) is green, I'm comfortable approving — but agree with your own note that it's still worth an actual click-through after deploy, since neither of us has exercised the visual flow.

Approved.

PR #1515, opened 2026-08-25
fix(api): let staff reply to a ticket nobody has claimed

crowlKats:
**Replying to a ticket that arrived by email fails with a 500 and the reply is discarded.** That is the ordinary way support mail gets answered, so answering inbound email does not currently work at all.

Found while building something else; the new test was simply the first thing to exercise this path.

## Cause

`ticket_add_message` works out whether the author is the person who opened the ticket, to decide which way the message is going and who is owed a response. It asked the database:

```sql
SELECT creator = $1 as "is_creator!" FROM tickets WHERE id = $2
```

A ticket opened by email has no creator until somebody claims it, and `NULL = $1` is NULL rather than false — so the query returned NULL into a column the caller had asserted was non-null. Every such reply failed to decode and 500'd.

It went unnoticed because the flow it breaks is the one combination the tests did not cover: replies were only ever exercised on tickets opened through the web UI, which always have a creator.

## Fix

`COALESCE(creator = $1, false)` — a ticket with no creator has no author who is its creator.

The regression test replies to an unclaimed ticket as staff and checks the reply lands and moves the ticket to `waiting_on_user`. I confirmed it fails against the old query before fixing it, rather than assuming.

## Also here

Removed the em dashes from the ticket emails, which read badly in a plain-text mail client. Split into two sentences rather than swapped for a hyphen, which would only be a worse version of the same construction. Affects all three ticket email templates, text and HTML.

## Testing

237 API tests pass. Clippy `-D warnings`, `cargo fmt`, `sqlx prepare --check` clean.

### DIFF
--- a/api/.sqlx/query-9285cef0b61fce21343e353c512e140ef880c3be8ed33191c40e45e8ae6bc0ab.json
+++ b/api/.sqlx/query-9285cef0b61fce21343e353c512e140ef880c3be8ed33191c40e45e8ae6bc0ab.json
@@ -1,6 +1,6 @@
 {
   "db_name": "PostgreSQL",
-  "query": "SELECT creator = $2 as \"is_creator!\" FROM tickets WHERE id = $1",
+  "query": "SELECT COALESCE(creator = $2, false) as \"is_creator!\" FROM tickets WHERE id = $1",
   "describe": {
     "columns": [
       {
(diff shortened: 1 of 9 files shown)


crowlbot (approved):
Clean root-cause fix — `NULL = $2` evaluating to `NULL` rather than `false` in Postgres is exactly the kind of bug that's easy to miss because it only shows up on the one combination (email-opened, unclaimed ticket + a reply) that the existing tests never exercised. `COALESCE(creator = $2, false)` is the right minimal fix, and the regression test exercises the real path end-to-end (inbound webhook → unclaimed ticket → staff reply → status flips to `waiting_on_user`) rather than just unit-testing the query in isolation.

Em-dash cleanup in the templates is a reasonable, low-risk drive-by — sentence split reads fine in plain text.

CI green. Approved.

PR #1516, opened 2026-08-25
fix(api): version the usage snippets

### LINKED ISSUE #721: Show version in usage snippets for non-latest versions

### LINKED ISSUE #1226: Direct import code snippet should contain a version
According to the new Deno lint rules, direct imports with inline specifiers should contain a version. 
As it currently stands the code snippets on each package do not have versions. 

Image

crowlKats:
Two related fixes to the "Use with" snippets, both in `DocUsageComposer`:

- **#721**: when a pinned version is being viewed, the add commands (`deno add`, `pnpm i`, `yarn add`, `vlt install`, `npx jsr add`, `bunx jsr add`) now install exactly that version (`jsr:@scope/pkg@1.2.3`) instead of silently installing the latest.
- **#1226**: the "Import directly with a jsr specifier" snippet now always carries a version constraint, so copied snippets pass Deno's `no-unversioned-import` lint rule: a caret range on the latest version (`jsr:@scope/pkg@^1.2.3`, matching what `deno add` writes), and the exact version when a version is pinned.

The bare "Import symbol" specifier stays unversioned — it's the specifier that `deno add` / the package manager maps.

Added assertions to `test_package_docs` covering both the pinned and `latest` cases.

Note: since #1473 docs are only served for the latest version, so the pinned-version case currently only materializes on pinned-latest URLs — the fix is correct for both and future-proof if that restriction is ever lifted.

Fixes #721.
Fixes #1226.

## Screenshots

Production today (unversioned jsr specifier):

before: import from jsr:@luca/flag without a version

After — latest version page (caret constraint, unversioned add command) and pinned version page (exact version everywhere):

| Latest | Pinned |
|--|--|
| after: jsr specifier with caret range | after: exact version in add command and jsr specifier |

### DIFF
--- a/api/src/api/package.rs
+++ b/api/src/api/package.rs
@@ -4219,7 +4219,8 @@ ggHohNAjhbzDaY2iBW/m3NC5dehGUP4T2GBo/cwGhg==
     let task = process_tarball_setup(&t, create_mock_tarball("ok")).await;
     assert_eq!(task.status, PublishingTaskStatus::Success, "{:?}", task);
 
-    // index page
+    // index page (pinned version): usage snippets install exactly this
+    // version, and the jsr specifier is pinned to it
     let mut resp = t
       .http()
       .get("/api/scopes/scope/packages/foo/versions/1.2.3/docs")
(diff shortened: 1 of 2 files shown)


crowlbot (approved):
Verified the full `compose` function, not just the diff hunk — `scoped_name` (used identically by every package manager's add command: pnpm, Yarn, vlt, npm, Bun, plus Deno's own `deno add`) was already correctly gated on `version_is_latest` before this change; only the new `jsr_url` for the "direct jsr specifier" snippet needed the caret-vs-exact distinction, and that's exactly what's added. The bare `import` (Import symbol) snippet correctly stays on the unversioned `url`, matching the stated "that's the specifier `deno add` maps" reasoning.

Test coverage checks both branches concretely (asserts the caret form is present and the exact-pinned form is absent on the `latest` page, and vice versa on the pinned page) rather than just checking one snippet exists.

CI green. Approved.

PR #1517, opened 2026-08-25
fix: don't render doc section headers with no contents

### LINKED ISSUE #918: In docs, don't show a "Functions" header if it's not going to have any contents
https://jsr.io/@babel-test-6ae45912/runtime@0.0.1-8.0.0-alpha.15/doc/helpers/arrayWithHoles
Image

This is admittedly a bit of a weird package, as it has no type definitions at all

It also has no exports: it was meant to have exports, but I messed up my publishing script so it ended up publishing as CJS, which from an ESM point of view has no exports. This is however not an unreasonable case, as you might have an actual module that has no exports: it's very common for polyfills.

JSR should not show the "Functions" header, as there are no functions exported by the package.

crowlKats:
deno_doc creates section headers (e.g. "Functions") before filtering symbols by visibility, so a module whose only symbols are private ends up with a header-only section — both on the page and in the "Document Navigation" ToC.

Drop sections whose content is empty (and their ToC entries) when rendering module, index and all-symbols pages, and guard in the frontend `Section` component as well. Intentionally header-only `empty`-kind sections still render.

Fixes #918.

## Screenshots

A module whose only function is private (local dev registry):

| Before | After |
|--|--|
| before: empty Functions header and ToC entry | after: no empty section or ToC entry |

### DIFF
--- a/api/src/docs.rs
+++ b/api/src/docs.rs
@@ -793,6 +793,49 @@ pub fn get_generate_ctx(
   .unwrap()
 }
 
+/// Drops sections whose content is empty — for example a "Functions" section
+/// whose only entries were filtered out for being private, since deno_doc
+/// creates the section headers before filtering symbols by visibility.
+/// Returns the anchors of the dropped section headers so the matching ToC
+/// entries can be pruned.
+fn drop_empty_sections(
+  sections: &mut Vec<deno_doc::html::util::SectionCtx>,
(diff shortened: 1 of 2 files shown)


crowlKats:
Closing — this belongs upstream: deno_doc shouldn't create section headers for symbols it then filters out. Reworking it in deno_doc.

PR #1518, opened 2026-08-25
feat: show module JSDoc examples on the package overview page

### LINKED ISSUE #895: Include examples in generated package page if present in the JSDoc

crowlKats:
The overview page already rendered module-level `@example` tags when the module JSDoc was the overview's prose source, but dropped them entirely when the README was (readme source setting, or JSDoc fallback to README). Build the module doc context in both modes and only swap out the prose, so a module's examples (and its `@deprecated` banner) show up regardless of the readme source setting.

Fixes #895.

## Screenshots

A package with readme source set to README whose module JSDoc has an `@example` (local dev registry):

| Before | After |
|--|--|
| before: README only, examples dropped | after: README plus Examples section |

(The after shot also shows highlighted TOML/YAML fences from #1519 — same local stack.)

### DIFF
--- a/api/src/docs.rs
+++ b/api/src/docs.rs
@@ -824,21 +824,25 @@ pub fn render_docs_html(
 
       let render_ctx = RenderContext::new(ctx, doc_nodes, UrlResolveKind::Root);
 
-      let mut index_module_doc = match readme_source {
-        ReadmeSource::JSDoc => ctx
-          .main_entrypoint
-          .as_ref()
-          .map(|entrypoint| {
-            deno_doc::html::jsdoc::ModuleDocCtx::new(
-              &render_ctx,
-              entrypoint,
-              false,
-              false,
-            )
-          })
-          .unwrap_or_default(),
-        ReadmeSource::Readme => Default::default(),
-      };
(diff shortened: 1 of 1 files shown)


crowlbot (approved):
Traced the fix against the pre-change code: previously `ReadmeSource::Readme` produced `index_module_doc = Default::default()` wholesale, discarding `sections.examples`/`deprecated` along with the prose. The fix always builds `index_module_doc` from the entrypoint (so those sections are populated) and only clears `sections.docs` when the README is the configured source — letting the existing `if sections.docs.is_none()` fallback-injection block do its job for both the explicit-README and JSDoc-with-no-module-doc cases uniformly. That matches the stated "build in both modes, only swap the prose" description.

One gap: no test changes accompany this — CI is green only because existing coverage doesn't exercise the readme-source-with-JSDoc-examples combination this fixes. Worth an assertion in `test_package_docs` (or similar) checking `sections.examples`/`deprecated` survive with `readme_source: Readme`, so this doesn't regress silently later — not blocking given the fix is small and the screenshots demonstrate it, but flagging since every other PR in this batch added a regression test for its fix.

CI green. Approved.

Issue #593, opened 2024-06-04
Support more syntax highlighting languages in markdown


lowlighter:
I'm not sure what is the full list of supported languages, but it's definitely missing some.

IMHO all somewhat related "web languages" should be supported (css, html, js, ts, etc.)
I also think that any "data languages" should also be added (json, xml, toml, yaml, etc.) because it's not rare having cli/app that deals with these
Finally some "common languages" should also be added (md, sh, diff).

It's also a regression compare to deno.land/x, because these were supported, you can compare both render for instant:
- https://jsr.io/@libs/xml
- https://deno.land/x/xml

binyamin:
FWIW, I don't see a regression. In fact, there are 2 code snippets which JSR renders and deno.land/x does not.

But, for comparison, here are the languages JSR supports ([code](https://github.com/jsr-io/jsr/blob/68551a163fbfebda9e2c47015c71b11703765204/api/src/tree_sitter.rs)):
- JS
- TS
- JSX
- TSX
- JSON & JSONC
- CSS
- Markdown
- xml & dtd
- Rust
- HTML
- Bash & sh

The missing languages are:
- TOML
- YAML
- C
- ignore
- git
- docker
- batch
- diff
- editorconfig (I would go for "ini", and alias editorconfig to that)

crowlKats:
closing as the ones we have now added a bunch, and if there is any inteerest for more individual issues should be opened, as this is very open-ended.

PR #1519, opened 2026-08-25
feat: syntax highlighting for TOML, YAML, C, diff, INI, batch and Dockerfile

crowlKats:
Adds tree-sitter grammars for TOML (`toml`), YAML (`yaml`/`yml`) and C (`c`/`h`) to markdown code blocks and the file view, plus fence aliases for `mjs`/`cjs`/`mts`/`cts` (the file view already mapped those extensions, but markdown fences missed them). Grammar crates chosen for compatibility with the pinned tree-sitter 0.22: `tree-sitter-toml-ng@0.6.0`, `tree-sitter-yaml@0.6.1`, `tree-sitter-c@0.21.4`.

The rest of the languages requested in #593 (diff, ini/editorconfig, batch, dockerfile) need a tree-sitter 0.25 upgrade first — their current grammar crates only expose `LanguageFn` — so that's left for a follow-up; `gitignore`/`git` grammars aren't published on crates.io at all.

Towards #593 (the remaining languages land in a follow-up PR right after this merges — the merge queue locks this branch).

## Screenshots

A README with `toml`/`yaml`/`diff`/`ini`/`dockerfile` fences (local dev registry):

| Before | After |
|--|--|
| before: all fences plain | after: toml and yaml highlighted |

(diff/ini/dockerfile remain plain — they're in the follow-up tranche that needs the tree-sitter upgrade. The after shot also shows the Examples section from #1518 — same local stack.)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3528,6 +3528,7 @@ dependencies = [
  "tracing-opentelemetry",
  "tracing-subscriber",
  "tree-sitter-bash",
+ "tree-sitter-c",
  "tree-sitter-css",
  "tree-sitter-highlight",
  "tree-sitter-html",
@@ -3536,8 +3537,10 @@ dependencies = [
  "tree-sitter-md",
  "tree-sitter-regex",
  "tree-sitter-rust",
+ "tree-sitter-toml-ng",
  "tree-sitter-typescript",
  "tree-sitter-xml",
+ "tree-sitter-yaml",
  "url",
  "urlencoding",
  "uuid",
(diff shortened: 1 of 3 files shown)


crowlbot (approved):
Straightforward, follows the exact existing pattern for every other language (same `HighlightConfiguration::new` + `OnceLock` shape as bash/rust/etc.), and the crate versions in `Cargo.toml` match what's described. Good test: it doesn't just check `tree_sitter_language_cb` returns `Some` for each new alias, it runs the highlighter and asserts real `<span>` output — which would have caught a grammar/query mismatch that the `Some`-check alone wouldn't. CI green (including the build itself, which is the real verification of tree-sitter 0.22 compatibility). Approved.

crowlKats:
Second commit completes the issue: upgraded tree-sitter to 0.25 (single `tree-sitter` in the graph, HTML output unchanged — the renderer callback API changed shape but emits identical class spans), then added the remaining languages: `diff`/`patch`, `ini`/`editorconfig`/`gitconfig`, `bat`/`batch`/`cmd`, and `dockerfile`/`docker`/`containerfile` (via `tree-sitter-containerfile`, the only current LanguageFn-based dockerfile grammar). The file view also falls back to well-known extensionless file names (`Dockerfile`, `Containerfile`, `.editorconfig`). `gitignore`/`git` grammars have no published crates, so those stay unsupported.

This now closes #593 when merged.

PR #1521, opened 2026-08-25
fix(api): show npm-ecosystem usage instructions for browser/workerd-only packages

### LINKED ISSUE #799: "Use with" instructions should also be shown for browser-only libraries
People developing for the browser still use a package manager to install their dependencies, so instructions on how to install something thrown `npm`/`yarn`/etc are also valuable for them.

https://jsr.io/@nic/local-highlight-registry is a library that only works in browsers. However, I had to mark Node.js compatibility as "unknown" rather than "no" to get the instructions to show up.

crowlKats:
A package that explicitly marks only browsers (or Cloudflare Workers) as supported — with `node: false` — got no pnpm/Yarn/vlt/npm tabs at all, even though browser-only libraries are still installed through npm-ecosystem package managers via a bundler. The reporter had to mark Node compat as "unknown" to get install instructions to show.

The npm-family tabs now also show when `browser` or `workerd` is explicitly `true`, regardless of `node`.

Also folds `runtime_compat` into the `GenerateCtxCache` key: the usage instructions baked into a cached `GenerateCtx` depend on it, and it is editable in the package settings, so editing runtime compat could serve stale tabs until cache eviction.

Fixes #799.

## Screenshots

A browser-only package (`runtimeCompat: { browser: true, deno: false, node: false, workerd: false, bun: false }`) on a local dev registry.

| Before | After |
|--|--|
| before: no usage instructions at all | after: pnpm/Yarn/vlt/npm tabs shown |

### DIFF
--- a/api/src/docs.rs
+++ b/api/src/docs.rs
@@ -202,8 +202,9 @@ pub async fn download_doc_nodes(
 }
 
 /// Cache for fully-built GenerateCtx. Keyed by
-/// `scope/package/version/is_latest/has_readme` so concurrent requests
-/// for the same doc page share a single GenerateCtx without rebuilding.
+/// `scope/package/version/is_latest/has_readme/runtime_compat` so concurrent
+/// requests for the same doc page share a single GenerateCtx without
+/// rebuilding.
 #[derive(Clone)]
 pub struct GenerateCtxCache {
   cache: moka::future::Cache<String, Arc<GenerateCtx>>,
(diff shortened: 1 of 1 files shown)


crowlbot (approved):
Both changes check out:

- **Cache key fix**: `runtime_compat` was already a parameter to `GenerateCtxCache::get` (used elsewhere in the function) but wasn't part of the cache key — so this is a genuine "editing runtime compat in settings could serve stale usage-instruction tabs until eviction" bug, not a hypothetical. Including it in the key via `{:?}` is the right minimal fix (compiles fine, so `RuntimeCompat: Debug` holds).
- **Tab-visibility condition**: purely additive — the original `!node.is_some_and(|c| !c)` clause (show unless node is explicitly `false`) is unchanged, just OR'd with `browser == Some(true)` / `workerd == Some(true)`. So packages with node/browser/workerd all unspecified behave exactly as before; only the explicit-node-false-but-browser/workerd-true case newly shows the npm-family tabs, matching the bug report precisely.

Same note as #1518: no test changes here, so this relies on the screenshot rather than an assertion in `test_package_docs` (e.g. checking pnpm/Yarn tabs appear for `node: false, browser: true`). Not blocking, but worth adding given how easy this condition would be to silently regress.

CI green. Approved.

PR #1522, opened 2026-08-25
feat(tickets): staff-only notes and canned replies

crowlKats:
Two things support asked for.

> **Stacked on #1515.** That branch carries a production fix (staff replies to email tickets 500), and this work sits on top of it. Merge #1515 first; the diff here collapses to just this feature once it lands.

## Staff-only notes

Staff can write a note on a ticket that the person who opened it never sees. It shows in the timeline on a muted yellow card, marked "Internal note".

Modelled as a flag on the existing messages rather than a table of its own, so a note keeps its place in the conversation — the point of writing one is that it sits next to the message it is about. A note is never emailed, and cannot move the ticket's status: nobody outside the team has seen it, so it cannot change who is owed a response. It is also excluded from the unread badges for the same reason.

### Keeping them from leaking

This is the part worth reviewing. Rather than remembering to filter in each place, the blanket `From<FullTicket> for ApiTicket` is gone, replaced by:

```rust
ApiTicket::for_viewer(full, viewer_is_staff)
```

A conversion reachable by `.into()` is one that will eventually be reached without anybody deciding whether the reader may see notes. Making the viewer an argument means the compiler asks. It pointed at five call sites, **two of which would have leaked**: the response to claiming a ticket, and the account ticket list.

Tests cover the reporter reading with a claim token, reading after claiming, the account ticket list, and a non-staff attempt to write a note.

## Canned replies

The answers from the runbook, grouped by situation (scope quota, package quota, package size, reserved scopes, deletions), in a picker next to the reply box for staff.

- **Inserting appends rather than replaces**, so a draft already being written is not thrown away.
- **The parts a person has to fill in are in brackets**, as in the runbook, and anything left unreplaced is listed under the box before sending. Sending "I've increased your scope quota to `[new_limit]`" verbatim is the obvious way to misuse a template.
- **The warning does not block.** Square brackets are ordinary prose, and a draft that cannot be sent is worse than one that is flagged.

### An assumption to confirm

**The templates live in the frontend, not the database.** They are house policy, they change about as often as the policy pages they link to, and keeping them in the repo means a wording change gets reviewed like any other user-facing copy — at the cost of a deploy to change one. If support would rather edit them in the admin UI, they can move; say so and I'll do it.

Two of the runbook's entries said only "explain the denial reason" rather than giving wording, so I wrote a short opener with a `[reason]` placeholder for those (package size denied, reserved scope denied). Worth a look — they're mine, not yours.

## Testing

241 API tests pass, 4 of them new here. Clippy `-D warnings`, `cargo fmt`, `sqlx prepare --check`, `deno task lint:frontend`, `deno fmt` and license headers clean.

Not verified in a browser — local dev still 404s on every route for me. The API behaviour is covered by tests; the visual side (the yellow note card, the template picker) wants an eye.


crowlbot (commented):
Reviewed the part you flagged as worth reviewing — the leak-prevention design holds up. Traced all 5 call sites the `for_viewer` refactor touched:

- `tickets.rs::claim_handler` and `self_user.rs::list_tickets` are indeed the two that would have leaked under the old blanket `From` — confirmed both now explicitly pass `false` (and `list_tickets` correctly passes `false` even for a staff member reading their own tickets, since that's the account view, not the admin one — a detail that's easy to get backwards and you got right).
- `tickets.rs::get_handler` filters `internal` before events are even built, `admin.rs`'s two endpoints correctly pass `true` (both behind `check_admin_access`), and `post_message_handler` blocks non-staff from setting `internal: true` and skips notification/status-change for a note — all matching the description.
- The DB-level `CHECK` constraint (`internal = false OR (author IS NOT NULL AND direction = 'outbound' AND email_message_id IS NULL)`) is good defense in depth beyond the application-level checks.
- Test coverage directly targets the leak scenarios by name (`a_staff_note_is_never_shown_to_the_reporter`, `a_staff_note_does_not_survive_claiming_the_ticket`, `a_note_cannot_be_written_by_anyone_but_staff`) rather than just exercising the happy path.

On the open question: keeping templates in the repo rather than the database seems right to me for the reason given (reviewed like other user-facing copy) — no objection there.

**One blocker, unrelated to the content**: this PR currently shows `mergeable: CONFLICTING` / `mergeStateStatus: DIRTY` against `main` — looks like it branched before #1515 landed and needs a rebase (the diff here still carries #1515's `COALESCE` change, which is now already on `main`). CI ran green, but GitHub won't let this merge until that's resolved. Holding off on approval for that reason alone; will re-review once it's rebased — the substance above already reads as solid.

crowlKats:
@crowlbot re-review

crowlbot (changes requested):
Re-reviewed after the rebase — the merge-conflict issue from before is gone, and the `for_viewer`/`.into()` redesign is a good way to force every read site to make an explicit call about staff notes. The five call sites you mention (`admin::list_tickets`, `admin::patch_ticket`, `self_user::list_tickets`, `tickets::post_handler`, `tickets::claim_handler`) all look correctly decided, and the four new tests cover the read-side leak paths well.

One bug I think blocks merge, in `Database::ticket_add_message` (`api/src/db/database.rs:4780`):

```rust
let direction = if author_is_creator {
  TicketMessageDirection::Inbound
} else {
  TicketMessageDirection::Outbound
};
```

This doesn't account for `message.internal`. The new CHECK constraint requires `direction = 'outbound'` whenever `internal = true`:

```sql
ALTER TABLE ticket_messages ADD CONSTRAINT ticket_messages_internal_is_unsent_and_authored CHECK (
    internal = false OR (author IS NOT NULL AND direction = 'outbound' AND email_message_id IS NULL)
);
```

`post_message_handler` only lets `TicketAccess::User { staff: true }` write an internal note, but a staff account can also be the `creator` of a ticket — anyone can open one via `POST /api/tickets`, staff included (e.g. a staff member filing a ticket about their own scope/quota, or just testing). If that same staff account then adds an internal note to a ticket they themselves opened, `author_is_creator` is `true`, `direction` computes to `Inbound`, and the `INSERT` trips the CHECK constraint — the request 500s instead of the note being saved.

None of the four new tests hit this: all of them use an email-opened ticket where the staff author is never the creator. Repro would be: staff user opens their own ticket via `POST /api/tickets`, then that same user posts `{"message": "...", "internal": true}` to it.

Fix is presumably to force outbound whenever the message is internal, independent of `author_is_creator`:

```rust
let direction = if message.internal || !author_is_creator {
  TicketMessageDirection::Outbound
} else {
  TicketMessageDirection::Inbound
};
```

Everything else here (the migration, the sql_fragments unread-badge exclusion, the frontend note styling, the template picker with the append-not-replace behavior and the non-blocking placeholder warning) looks right to me. On the open question in the description — templates living in the frontend repo vs. the DB — I think the repo is the right call for the reasons you gave; I wouldn't move it to the admin UI unless support actually asks.

crowlKats:
Good catch, and confirmed — that 500s exactly as you described.

I reproduced it before fixing: staff open a ticket via `POST /api/tickets`, then post `{"internal": true}` to it, and the insert trips `ticket_messages_internal_is_unsent_and_authored`. Pushed as 1123d6d, essentially your suggested fix:

```rust
// A note is always from the JSR side, even when the staff member writing it
// is the person who opened the ticket — which happens whenever staff file
// one themselves. Treating it as inbound there would both misattribute it
// and violate the check constraint that ties internal to outbound.
let direction = if author_is_creator && !message.internal {
  TicketMessageDirection::Inbound
} else {
  TicketMessageDirection::Outbound
};
```

I wrote it as `author_is_creator && !message.internal` rather than `message.internal || !author_is_creator` — same truth table, but it reads as "inbound only when the creator is speaking for real", which is the actual rule.

The new test opens a ticket as staff and notes on it. I checked it fails against the old logic before applying the fix rather than assuming it reproduced.

You put your finger on the gap in my tests, too: all four note tests used an email-opened ticket, where staff can never be the creator, so the author-is-creator branch was never exercised with `internal = true` at all. That combination was invisible to the whole suite.

Noted on the templates staying in the repo — leaving them there.

crowlKats:
@crowlbot rereview

crowlbot (approved):
Confirmed — `direction = if author_is_creator && !message.internal { Inbound } else { Outbound }` closes the gap cleanly, and the new `a_staff_member_can_note_on_a_ticket_they_opened_themselves` test covers exactly the self-authored-ticket case the CHECK constraint would have 500'd on. CI green. LGTM.

PR #1523, opened 2026-08-25
feat: render markdown files in the source view

### LINKED ISSUE #709: Support rendering markdown files besides README.md
It's common for README.md files to link to other markdown files. On GitHub, when you click one of those links it will take you to a page with the markdown rendered instead of the source code for the markdown file. On JSR, those links take you to the source code for the markdown files.

You can see an example of this by clicking any of the links near the bottom of the overview for my package.

https://jsr.io/@udibo/react-app

crowlKats:
Markdown files in the Files view now render like on GitHub, with a **Preview / Code** toggle in the file navigation bar (preview is the default; `?code` shows the highlighted source). The API returns rendered HTML alongside the highlighted source for `.md`/`.markdown` files, using the same comrak pipeline, ammonia sanitizer and relative-link resolution as README rendering — links resolve against the file's directory, and packages with a linked GitHub repository keep their existing link rewriting.

Fixes #709.

## Screenshots

Production today (markdown files show raw source only):

before: raw markdown source on jsr.io

After (local dev registry) — preview by default, with a toggle:

| Preview | Code |
|--|--|
| after: rendered markdown with Preview/Code toggle | after: highlighted source via the Code toggle |

### DIFF
--- a/api/src/api.yml
+++ b/api/src/api.yml
@@ -3556,6 +3556,10 @@ components:
               type: string
               nullable: true
               description: HTML-rendered source code view. Null for binary files.
+            rendered:
+              type: string
+              nullable: true
+              description: Rendered HTML, present for markdown files.
           required:
             - kind
             - size
--- a/api/src/api/package.rs
+++ b/api/src/api/package.rs
@@ -1738,7 +1738,7 @@ pub async fn get_source_handler(
 
   let db = req.data::<Database>().unwrap();
   let buckets = req.data::<Buckets>().unwrap();
-  let _ = db
+  let (_, repo, _) = db
     .get_package(&scope, &package)
     .await?
     .ok_or(ApiError::PackageNotFound)?;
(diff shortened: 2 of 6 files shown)


crowlbot (approved):
Checked the part that matters most here — this renders arbitrary package-supplied markdown to HTML, so sanitization has to actually be shared with the README path, not just described as such:

- `render_markdown_file` calls `deno_doc::html::comrak::create_renderer(...)` with `AMMONIA.clean` as the sanitizer callback — `AMMONIA` is the same pre-existing static `lazy_static! { static ref AMMONIA: ammonia::Builder ... }` that `get_generate_ctx` already uses for README rendering (only call sites of `create_renderer` in the file, both wired to the identical sanitizer). Not a parallel, differently-configured pipeline.
- `get_url_rewriter` is likewise the existing shared helper (used for the repo-link-rewriting claim), not reimplemented.
- The relative-link test is concrete, not just smoke-tested: a file at `/docs/guide.md` linking `./other.md` resolves to `/@scope/pkg/1.0.0/docs/./other.md` (correct directory-relative resolution, browser-normalizes the `./`), and a fenced code block is confirmed to go through the tree-sitter highlighter (`<span class=` in the output) rather than rendering as plain text.

CI green. Approved.

PR #1524, opened 2026-08-25
fix(api): render markdown footnotes

### LINKED ISSUE #291: request: support markdown footnotes in READMEs
A widespread Markdown extension is footnotes. See https://www.markdownguide.org/extended-syntax/#footnotes. It'd be great if READMEs supported a footnote extension.

Compare e.g. https://jsr.io/@olets/css-fluid-length, where footnote-y syntax becomes in-place links, and https://codeberg.org/olets/css-fluid-length, where the same renders with footnotes.

crowlKats:
jsr half of #291. deno_doc 0.205.0 (denoland/deno_doc#849) turns comrak's `footnotes` extension on, so this bumps to it and widens the ammonia allowlist to let the markup it emits through — the allowlist stripped the `<section>` wrapper and the `footnote-ref`/`footnote-backref` classes, which would have mangled every rendered footnote. The `#fn-*`/`#fnref-*` anchors already survive: `id` is an allowed generic attribute and fragment hrefs pass the URL rewriter untouched.

- `api`: `=0.204.0` → `=0.205.0`, plus `section` + the three footnote classes on the sanitizer. Everything that renders markdown through `AMMONIA` gets footnotes — JSDoc, READMEs, and the markdown source view from #1523.
- Test renders `claim[^1]` through `render_markdown_file` (the real render path, sanitizer included) rather than a hand-written fixture, so it tracks whatever markup the deno_doc of the day emits, and asserts both the classes and the anchor pair that makes the jump there and back work.
- `frontend`: the pin moves with the api's (they're kept in sync), which also lets `ModuleDoc` drop its `symbols_trimmed` cast — 0.205.0's html-types declare the field (denoland/deno_doc#845).

Footnote styling ships in deno_doc's `comrak.gen.css`, which the frontend already serves as `comrakCss`, so no css change here.

Fixes #291.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1056,9 +1056,9 @@ dependencies = [
 
 [[package]]
 name = "deno_doc"
-version = "0.204.0"
+version = "0.205.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "05c121a45c4001f3b89cc7af2e3cab71514db6807ded0c9a0317b118d3d3fe87"
+checksum = "4d80bf35d7150d36768fc1f4e6bb71444f39924263b71a343e1a7b4efa64b9d8"
 dependencies = [
  "anyhow",
  "cfg-if",
--- a/api/Cargo.toml
+++ b/api/Cargo.toml
@@ -98,7 +98,7 @@ async-compression = { version = "0.4", features = ["futures-io", "gzip"] }
 deno_graph = "=0.110.2"
 deno_ast = { version = "0.53.0", features = ["view"] }
 # sync with frontend/deno.json
-deno_doc = { version = "=0.204.0", features = ["comrak"] }
+deno_doc = { version = "=0.205.0", features = ["comrak"] }
 deno_error = "0.7.0"
 comrak = { version = "0.29.0", default-features = false }
 ammonia = "4.0.0"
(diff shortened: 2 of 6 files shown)


crowlbot (approved):
Tightly scoped allowlist change — checked for accidental over-broadening since this touches the sanitizer directly:

- Only one new bare tag (`section`), and it gets exactly one allowed class (`footnotes`) — no new attributes, no `href`/`src`/`data-*` widening.
- `sup`/`a` were already allowed tags; this only adds one class each (`footnote-ref`, `footnote-backref`).
- The `data-footnote-ref`/`data-footnote-backref`/`data-footnotes` attributes comrak actually emits are *not* allowlisted and get stripped — correctly left out since they're inert markers, not required for the footnote to render or link correctly, and the test doesn't assert their presence.
- The test uses the literal HTML shape comrak's footnotes extension emits (not a simplified stand-in) and checks the specific things that need to survive: the `section`/`sup`/backref classes, and the `#fn-1`/`id="fn-1"` anchor pair that makes the footnote jump work — not just "sanitizer doesn't crash."

Good call flagging it's inert until the deno_doc bump lands and that merge order doesn't matter.

CI green. Approved.

crowlbot (approved):
The follow-up commit looks good — bumping to 0.205.0 turns the extension on as planned, the test now goes through the real comrak renderer (and checks the `fnref` anchor pair too, which the old fixture-based test didn't), and dropping the `symbols_trimmed` cast now that the published types carry it is a nice cleanup. CI green, nothing further from me.

olets:
Thank you!

PR #1526, opened 2026-08-25
feat(api): show a `deno create` command for packages that support it

### LINKED ISSUE #1358: Command code block for "deno create"
Could we add another command code block for "deno create" under "Add package" in the top-right conrer of jsr.io, provided the package supports it?

Image

crowlKats:
Closes #1358.

`deno create @scope/package` rewrites the argument to `jsr:@scope/package/create` and runs it, so a package supports the command exactly when it exports `./create`. That export is already persisted with every version, so JSR has the signal it needs — no new config or heuristic.

The Deno tab of the usage panel now leads with a "Scaffold a project" block for those packages, and is unchanged for every other one, so the command is never advertised where it would fail.

### `@sweep/util`, which does not

When a non-latest version is being viewed, the command pins it the same way the `deno add` line does (`deno create @scope/package@1.2.3`).

Both cases are covered by unit tests, including that a `create.ts` file exported under some *other* key does not count.

### DIFF
--- a/api/src/analysis.rs
+++ b/api/src/analysis.rs
@@ -273,6 +273,9 @@ async fn analyze_package_inner(
       workerd: None,
       bun: None,
     },
+    // only the search index is taken from this ctx, so usage instructions
+    // don't matter here
+    false,
     registry_url.to_string(),
     None,
   );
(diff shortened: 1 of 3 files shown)


crowlbot (approved):
Clean. The two other `get_generate_ctx` call sites (analysis.rs's search-index ctx, and the diff handler) are both updated with `false` and a comment explaining why usage instructions don't matter there — didn't find a third call site that was missed. The scaffold/add ordering and the export-key-vs-file-name distinction (`./create` vs a file named `create.ts` under some other export) are both directly tested. Nice.

PR #1527, opened 2026-08-25
feat(api): link import/export specifiers in the file view

### LINKED ISSUE #17: Make import/export specifier clickable links in file view
These should be clickable:

image

crowlKats:
Closes #17.

Import and export specifiers in the file view are now links.

No re-parsing is involved: every published version's `moduleGraph2` metadata already records the source range of every specifier the module mentions — static and dynamic imports, `export … from`, `@ts-types` overrides, triple-slash references, the JSX import source pragma, and JSDoc type imports — so the view maps those ranges onto byte offsets and hands them to the highlighter.

Where each kind points:

| specifier | link |
| --- | --- |
| `./util.ts`, `../options.ts` | that file in this version's file view |
| `jsr:@scope/pkg@1.2.3` | the package page, pinned when the specifier pins a version |
| `npm:chalk@5` | the package on npmjs.com |
| `node:path` | the module on nodejs.org |
| `https://…` | the URL itself; registry URLs resolve back into JSR |

A relative specifier only becomes a link when it names a file that is actually in the version, so the view never offers a link to a 404.

### `/mod.ts` of a demo package

The links keep the syntax colour of the string they sit in, and announce themselves with a dotted underline that goes solid on hover (line 9 is hovered here):

### Rendering

`tree_sitter_highlight::HtmlRenderer` can't emit anchors, so the highlighting is rendered here instead. With no links to place it produces byte-identical output — a test pins that against the upstream renderer across several languages, including CRLF input, non-ASCII characters, and a file with no trailing newline.

An anchor never spans a highlight boundary or a line break: a link covering several highlight spans becomes several anchors with the same href, which keeps the markup well-formed (also covered by a test) and every part of it clickable.

### Cost

The file view now fetches the version metadata alongside the file, and only for paths that could be modules (`.ts`, `.tsx`, `.js`, `.mjs`, …), so READMEs, JSON and other assets are unaffected. Anything missing or unreadable there just yields a file view without links.

### DIFF
--- a/api/src/api/package.rs
+++ b/api/src/api/package.rs
@@ -1787,18 +1787,29 @@ pub async fn get_source_handler(
     None
   };
 
-  let path_buf = std::path::PathBuf::from(path);
+  let path_buf = std::path::PathBuf::from(&path);
 
   let source = if let Some(file) = file {
     let size = file.len();
 
-    let highlighter = deno_doc::html::comrak::ComrakHighlightWrapperAdapter(
-      Some(Arc::new(crate::tree_sitter::ComrakAdapter {
-        show_line_numbers: true,
-      })),
-    );
-
     let view = if let Ok(file) = String::from_utf8(file.to_vec()) {
+      let links = source_specifier_links(
+        &scope,
+        &package,
+        &version.version,
+        &path,
+        &file,
+        buckets,
+        req.data::<RegistryUrl>().unwrap().0.as_str(),
+      )
+      .await;
+
+      let mut adapter = crate::tree_sitter::ComrakAdapter::new(true);
+      adapter.links = links;
+      let highlighter = deno_doc::html::comrak::ComrakHighlightWrapperAdapter(
+        Some(Arc::new(adapter)),
+      );
+
       let mut out = vec![];
       highlighter.write_pre_tag(&mut out, Default::default())?;
       highlighter.write_code_tag(&mut out, Default::default())?;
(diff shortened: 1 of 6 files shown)


crowlbot (approved):
Thorough. Went through `source_links.rs` and the new `render_lines` in `tree_sitter.rs` in detail:

- `resolve_relative_href`'s containment check (`ctx.files.contains(path)`) is a good way to sidestep path-traversal concerns from `..`-heavy specifiers entirely — no link is offered unless the resolved path is literally one of the version's own files, so there's no need to reason about how many `..`s the URL crate's join allows.
- The href is escaped byte-wise into the attribute (`push_escaped` covers `"`, so no attribute breakout), and the chunking in `render_lines` correctly treats a link boundary as another reason to end a chunk alongside highlight-span and line boundaries — verified the "anchor never spans a highlight boundary" property by hand against the depth-tracking test, and it holds.
- `matches_upstream_renderer` pinning the hand-rolled path against `HtmlRenderer` byte-for-byte (CRLF, no-trailing-newline, multi-byte chars, several languages) is exactly the risk this kind of rewrite carries, and it's covered well.
- `dedup_by`'s "later vs earlier" comparator matches slice::dedup_by's actual (a, b) = (current, last-retained) argument order, so overlap removal is correct including chains of 3+ overlapping ranges, not just adjacent pairs.

Nothing blocking. CI green.

PR #1528, opened 2026-08-25
feat: show the number of documented symbols on the Docs tab

### LINKED ISSUE #1206: Display total symbols count in main `<PackageNav>`
Hello

In order to have a quick glance to "the surface API area" of a given package, it would be great to feature the total number of symbols more prominently.

Here's a quick mock-up of what it would look like in the main tabs labels: 

Image

Thanks

crowlKats:
Closes #1206.

The Docs tab now carries a chip with the number of documented symbols, next to the existing version / dependency / dependent counts, so the size of a package's API surface is visible at a glance.

### Where the count comes from

Doc nodes are generated at publish time and stored per version, so counting per page load would mean loading and flattening all of them on every request. Instead the count is computed during publish-time analysis, from the same render context the search index is already built from, and stored on the version's `meta` — which is JSONB, so there is no migration.

It counts what the docs actually list: exported symbols per entrypoint plus namespace members, minus `@internal` and non-exported types, deduplicated per `(file, qualified name)` the same way the search index is. Class and interface members aren't counted separately — they belong to their parent symbol rather than being their own listing entries.

### Versions published before this

There is no backfill, so their count is `null` rather than `0`, and the nav renders no chip for them instead of claiming the package has no symbols. Versions get a count as they are published; a backfill could be added later by re-running analysis, but that's a much heavier operation than this change warrants.

The count is exposed on both `Package` (from the latest version) and `PackageVersion`, and the nav prefers the version being viewed so a pinned old version doesn't show the latest version's number.

### DIFF
--- a/api/src/analysis.rs
+++ b/api/src/analysis.rs
@@ -234,7 +234,7 @@ async fn analyze_package_inner(
   .await
   .map_err(PublishError::NpmTarballError)?;
 
-  let (meta, readme_path) = {
+  let (mut meta, readme_path) = {
     let readme = files
       .iter()
       .find(|file| file.0.case_insensitive().is_readme());
@@ -276,6 +276,7 @@ async fn analyze_package_inner(
     registry_url.to_string(),
     None,
   );
+  meta.symbol_count = Some(count_symbols(&ctx));
   let search_index = deno_doc::html::generate_search_index(&ctx);
   let doc_search_json = if let serde_json::Value::Object(mut obj) = search_index
   {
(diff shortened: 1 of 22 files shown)


crowlbot (approved):
Mechanical wiring checks out — verified `PackageVersionMeta` has `#[serde(default)]` at the container level (with `Default` derived), so existing stored metas without `symbol_count` deserialize to `None` rather than erroring, which matters a lot given this is read from JSONB on every existing row. Also checked `NavItem`'s chip renders on `!== undefined`, not truthiness, so a package with a genuine zero-symbol count still shows a "0" chip rather than silently hiding it like the pre-count `null` case does — the `??` in `symbolCount={... ?? undefined}` preserves that distinction correctly.

One thing I'd note: `count_symbols` has no direct unit test (unlike the search-index code it mirrors) — worth a small one given it's the kind of function where an off-by-one in the dedup key would be easy to miss and hard to notice visually. Not blocking, CI's green and the route wiring is straightforward enough to trust.

PR #1529, opened 2026-08-25
feat(api): serve pages for non-exported symbols the public API refers to

### LINKED ISSUE #165: Documenting transitive types in API/symbols?
I was browsing the `@badrap/valita` symbols and the generated API docs and I see a lot of symbols are not linked/documented.

For example:

https://jsr.io/@badrap/valita/doc/~/ArrayType#method_func_0

```
func(arr: unknown, flags: number): RawResult<ArrayOutput<Head, Rest>>
```

Here neither `RawResult` nor `ArrayOutput` are linked or documented. These two type definitions are not exported by `src/index.ts` so it is understandable that they are not linked. But, to make things useful the reader of the API docs would want to see what these types are even if they cannot be imported directly.

crowlKats:
Closes #165. The other half is denoland/deno_doc#851, released in **deno_doc 0.206.0**, which this bumps to.

A type the public API refers to but doesn't export is already in the stored doc nodes, documented, with `DeclarationKind::Private`. The symbol page handler filtered exactly those out, so the links deno_doc now emits for them would 404.

Two changes to that filter:

- non-exported symbols get a page — that page is what the reference in the signature points at;
- `@internal` symbols now get none. The author asked for those to be hidden, and they were previously reachable by URL.

The `@internal` check uses `DocNodeWithContext::is_hidden()` from the release, rather than reimplementing the tag lookup here.

### The two cases, side by side

`parse()` returns a non-exported `RawResult` — linked, and the page behind it documents the type:

`probe()` returns a type that is *also* `@internal` — the author opted out, so it stays plain text with no page, exactly as before:

### What's reachable where

Verified against a package published to a local instance on this branch:

| symbol | exported | `@internal` | page | linked from signature | listings / search |
| --- | --- | --- | --- | --- | --- |
| `parse`, `Options` | ✅ | — | ✅ | — | ✅ |
| `RawResult`, `Origin` | ❌ | — | ✅ | ✅ | ❌ |
| `HiddenHelper` | ✅ | ✅ | ❌ (404) | — | ❌ |
| `OptedOut` | ❌ | ✅ | ❌ (404) | ❌ | ❌ |

The search index for that package contains exactly `Options`, `Options.separator` and `parse` — no occurrence of the non-exported or hidden names. So these pages are reachable only by following a reference from a signature that already mentions the type; they're never advertised as API.

### deno_doc bump

0.206.0 over 0.205.0 is exactly denoland/deno_doc#851 plus the version commit — no template or ctx-struct changes, so there is nothing to mirror into the JSX.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1056,9 +1056,9 @@ dependencies = [
 
 [[package]]
 name = "deno_doc"
-version = "0.205.0"
+version = "0.206.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "4d80bf35d7150d36768fc1f4e6bb71444f39924263b71a343e1a7b4efa64b9d8"
+checksum = "0a2af46876eb43b3334539f548c8c106f9ad86d330ca5661c04fcdb27fd8e463"
 dependencies = [
  "anyhow",
  "cfg-if",
--- a/api/Cargo.toml
+++ b/api/Cargo.toml
@@ -98,7 +98,7 @@ async-compression = { version = "0.4", features = ["futures-io", "gzip"] }
 deno_graph = "=0.110.2"
 deno_ast = { version = "0.53.0", features = ["view"] }
 # sync with frontend/deno.json
-deno_doc = { version = "=0.205.0", features = ["comrak"] }
+deno_doc = { version = "=0.206.0", features = ["comrak"] }
 deno_error = "0.7.0"
 comrak = { version = "0.29.0", default-features = false }
 ammonia = "4.0.0"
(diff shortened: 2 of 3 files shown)


crowlbot (approved):
Makes sense, and matches the description: swapping the `all(Private)` exclusion for an `@internal`-tag check is exactly what's needed to let a referenced-but-unexported type get a page while still honoring an explicit hide. Since it's gated on deno_doc#851 and nothing links here yet, the blast radius if the semantics are slightly off is just hand-typed URLs, so I'm fine with it as-is — though a small unit test on `is_hidden_symbol` (mirroring the two cases in the description: unexported-but-not-internal shows, `@internal` hides) would be nice to have before or when the upstream release lands and real links start pointing here, since that's when a regression would actually bite.

PR #1531, opened 2026-08-25
fix(api): keep markdown heading permalink icons intact through sanitization

### LINKED ISSUE #1530: Docs heading permalink icons mangled
After the `deno_doc` 0.204.0 update in #1502, permalink icons added to Markdown headings render incorrectly on jsr.io.

In dark mode, headings show a white square where the chain-link icon should be. In light mode, the icon is effectively invisible.

Example: https://jsr.io/@kjanat/ansispeck

### Light mode

<a href="https://github.com/user-attachments/assets/3a8049c1-568b-4ce5-ab2b-d58fa8788570">
  JSR package README in light mode with missing heading permalink icons
</a>

### Dark mode

<a href="https://github.com/user-attachments/assets/618129da-309b-4f51-95a9-4f93b5864110">
  JSR package README in dark mode with white squares instead of heading permalink icons
</a>

<details open><summary>gpt deep dive</summary>

## What happens

`deno_doc` 0.204.0 generates heading permalinks using an `anchorable` heading, an `anchor` link, and its bundled `link.svg`.

For a Markdown heading such as:

```markdown
## Install
```

`deno_doc` generates markup with the following structure:

```html
<h2 id="install" class="anchorable">
  <a href="#install" class="anchor" aria-label="Anchor" tabindex="-1">
    <svg
      width="16"
      height="16"
      viewBox="0 0 14 14"
      fill="none"
      xmlns="http://www.w3.org/2000/svg"
    >
      <g clip-path="url(#clip0_1081_1237)">
        <path ... fill="currentColor" />
      </g>
      <defs>
        <clipPath id="clip0_1081_1237">
          <rect width="14" height="14" fill="white" />
        </clipPath>
      </defs>
    </svg>
  </a>
  Install
</h2>
```

The SVG is bundled by `deno_doc` itself:

https://github.com/denoland/deno_doc/blob/80822afd7b6e2dc397e106cd89713bb5c7b7f575/src/html/templates/icons/link.svg

After JSR renders and sanitizes the README, the same heading is served as:

```html
<h2 id="install">
  <a
    href="https://jsr.io/@kjanat/ansispeck@0.4.2#install"
    rel="nofollow"
  >
    <svg
      width="16"
      height="16"
      viewBox="0 0 14 14"
      fill="none"
      xmlns="http://www.w3.org/2000/svg"
    >
      <path ... fill="currentColor"></path>
      <rect width="14" height="14" fill="white"></rect>
    </svg>
  </a>
  Install
</h2>
```

The rendered output differs in several relevant ways:

- the `anchorable` class is removed from the heading;
- the `anchor` class, `aria-label`, and `tabindex` are removed from the permalink;
- `<g>`, `<defs>`, and `<clipPath>` are removed from the SVG;
- their `<path>` and `<rect>` descendants remain.

The `<rect>` originally exists only inside the clipping definition. Once its containing `<defs>/<clipPath>` elements are removed, it becomes an ordinary painted SVG element.

Because the rectangle is painted after the chain-link `<path>`, it covers the icon completely.

This explains the visible result:

- **dark mode:** the white rectangle appears as a white square;
- **light mode:** the rectangle blends into the page background, making the permalink icon appear missing.

## The sanitizer also strips the anchor classes

The same sanitization also removes the classes required by JSR's permalink styling.

`deno_doc` emits:

```html
<h2 class="anchorable">
  <a class="anchor" ...>
```

JSR added CSS in #1502 specifically for this markup:

```css
.anchor {
  @apply hidden float-left leading-none text-stone-600 ml-[-24px] p-1 pr-1
    pt-1 top-0 bottom-0 dark:text-stone-400;
}

.anchorable {
  @apply relative scroll-mt-16;

  &:hover .anchor {
    @apply block;
  }
}
```

However, the rendered JSR output contains neither `class="anchorable"` nor `class="anchor"`.

As a result, the styling added for the generated permalink anchors cannot apply to the rendered markup.

## Regression

The regression appears to have been introduced by:

https://github.com/jsr-io/jsr/commit/2f1c0f883ca121a1f20686714bfea43ea6fd6d9c

#1502 updated `deno_doc` to 0.204.0 and added JSR-side styling for the newly generated Markdown heading anchors.

The heading-anchor feature itself was introduced upstream in `deno_doc` by:

https://github.com/denoland/deno_doc/commit/80822afd7b6e2dc397e106cd89713bb5c7b7f575

The upstream `link.svg` is valid. The broken structure appears in the rendered output after JSR's sanitization step.

</details>

crowlKats:
Fixes #1530.

## What was wrong

The ammonia allowlist in `api/src/docs.rs` permits `svg`, `path` and `rect`, but not `g`, `defs` or `clipPath`. Ammonia *unwraps* a disallowed tag rather than dropping its subtree, so the clip rect in the `link.svg` that `deno_doc` bundles escaped its `<clipPath>` and became an ordinary painted 14x14 white rectangle, drawn after the icon path and covering it — a white square in dark mode, an invisible icon in light mode.

Reproduced through the real render path (`## Install` in, sanitized HTML out):

```html
<h2 id="install"><a href="#install" class="" rel="nofollow"><svg …>
  <path … fill="currentColor"></path>
  <rect width="14" height="14" fill="white"></rect>
</svg></a>Install</h2>
```

That output also shows the second half of the bug: the `anchorable` and `anchor` classes are gone (`h*` had no allowed classes, `a` allowed only `footnote-backref`), along with `aria-label` and `tabindex`. The CSS added in #1502 to reveal the permalink on hover therefore had nothing to match.

And a third one, which only becomes visible once the classes survive: `.markdown a:not(.no_color)` (specificity 0-2-1) out-specifies `.anchor` (0-1-0), so the icon would render cyan and underlined.

## The fix

- Allow `g`, `defs` and `clipPath`, plus `clip-path` on `g`.
- Allow `anchor` on `a`, `anchorable` on `h1`–`h6`, and `aria-label`/`tabindex` on `a`.
- Exclude `.anchor` from the markdown link styling.

Allowing `<defs>` is deliberately preferred over dropping its content: anything that survives the allowlist inside a definition is inert by construction, so this closes the whole "definition content leaks out and paints" class of bug rather than just this one icon.

## Verification

Output now matches what `deno_doc` intended — `<h2 id="install" class="anchorable">`, `class="anchor"`, `<g clip-path=…>`, and the rect back inside `</clipPath>`. Covered by a new `renders_heading_permalinks_through_the_sanitizer` test that goes through the real render path, so it keeps up with whatever markup the `deno_doc` of the day emits. `cargo test` 241/241, `cargo fmt --check` and `cargo clippy --all-targets` clean.

### DIFF
--- a/api/src/docs.rs
+++ b/api/src/docs.rs
@@ -287,8 +287,12 @@ lazy_static::lazy_static! {
     let mut ammonia_builder = ammonia::Builder::default();
 
     ammonia_builder
-      .add_tags(["video", "button", "svg", "path", "rect", "section"])
+      .add_tags([
+        "video", "button", "svg", "path", "rect", "section", "g", "defs",
+        "clipPath",
+      ])
       .add_generic_attributes(["id", "align"])
+      .add_tag_attributes("g", ["clip-path", "fill", "opacity", "transform"])
       .add_tag_attributes("button", ["data-copy"])
       .add_tag_attributes(
         "svg",
@@ -325,7 +329,15 @@ lazy_static::lazy_static! {
       // comrak footnote output
       .add_allowed_classes("section", ["footnotes"])
       .add_allowed_classes("sup", ["footnote-ref"])
-      .add_allowed_classes("a", ["footnote-backref"])
+      .add_allowed_classes("a", ["footnote-backref", "anchor"])
+      // deno_doc heading permalinks
+      .add_tag_attributes("a", ["aria-label", "tabindex"])
+      .add_allowed_classes("h1", ["anchorable"])
+      .add_allowed_classes("h2", ["anchorable"])
+      .add_allowed_classes("h3", ["anchorable"])
+      .add_allowed_classes("h4", ["anchorable"])
+      .add_allowed_classes("h5", ["anchorable"])
+      .add_allowed_classes("h6", ["anchorable"])
       .add_allowed_classes(
         "div",
         [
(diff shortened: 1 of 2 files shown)


crowlbot (approved):
Good catch chasing this down to three separate but related sanitizer gaps. Confirmed the fix addresses all three: `g`/`defs`/`clipPath` are allowlisted so the clip rect stays nested (verified via the test's split-on-`</clipPath>` check, which is a nice way to actually assert containment rather than just presence), the `anchorable`/`anchor` classes plus `aria-label`/`tabindex` survive for the CSS to hook into, and excluding `.anchor` from the markdown link selector fixes the specificity fight. Allowing `<defs>` rather than dropping its contents is the right general fix too — closes the whole class of bug rather than just this instance. No new attack surface: none of the added tags/attributes accept a URL-bearing value (no `xlink:href`/`href`/`use`), so this stays presentation-only.

kjanat:
Fast one!

Issue #768, opened 2024-10-18
Relative URLs in README starting with /


dbushell:
If I have a relative URL starting with a forward slash `/` in a README for example:

```markdown
[Documentation](/docs/getting-started.md)
```

JSR will link to `jsr.io` and 404, e.g. `https://jsr.io/docs/getting-started.md`

I would expect this to link back to Github. Both GitHub and NPM allow `/` linking this way.

If the URL is not prefixed with `/` e.g. `docs/getting-started.md` JSR links back to GitHub correctly.

Muhammad-Owais-Warsi:
Hey @dbushell, can you please provide me with the link of the page, where the issue arises.

dbushell:
@Muhammad-Owais-Warsi [see this package](https://jsr.io/@dbushell/http-tunnel) for example — the "MIT License" link at the end is `/LICENSE`

lucacasonato:
The bug is somewhere in this function, presumably the early return that checks for leading `/` is wrong at the top of this file: https://github.com/jsr-io/jsr/blob/main/api/src/docs.rs#L131.

Muhammad-Owais-Warsi:
> s somewhere in this function, presumably the early return that checks for leading `/` is wrong at the top of this file:

Thanks @dbushell

Muhammad-Owais-Warsi:
> The bug is somewhere in this function, presumably the early return that checks for leading `/` is wrong at the top of this file: https://github.com/jsr-io/jsr/blob/main/api/src/docs.rs#L131.

Hey @marvinhagemeister , are you talking about this code snippet ?
```    if let Some(entrypoint) = entrypoint {
      if key.strip_prefix('/').unwrap_or(key) == entrypoint {
        entrypoint_url = Some(specifier.clone());
      }
    
   ```

inverted-capital:
> If the URL is not prefixed with `/` e.g. `docs/getting-started.md` JSR links back to GitHub correctly.

Not if the package was published out of a workspace monorepo, in which case it will link relative to the workspace root and 404 https://github.com/jsr-io/jsr/issues/709

mattvr:
Shouldn't this ideally link to `jsr.io/@namespace/repo/version/file` instead of `github.com/repo/file`? There aren't necessarily GH repos for every package, this could properly pin the version, and it might be a more expected & beneficial behavior for relative links to stay on site.

PR #1532, opened 2026-08-25
fix(api): don't rewrite registry links in JSDoc to the repository

### LINKED ISSUE #1175: JSDoc 404s when re-re-exported namespace `export * as is` produces `is.isAny` links
Summary
When re-re-exporting a namespace that itself re-exports many items with renamed exports, JSR's generated docs produce double-qualified names like is.isAny. Those links 404 on jsr.io and break symbol resolution.

Context
- Library: @core/unknownutil
- Version: 4.4.0-pre.0
- Docs page showing the issue: https://jsr.io/@core/unknownutil@4.4.0-pre.0/doc
- Pattern:
  - is/is.ts enumerates many re-exports such as:
    `export { isAny as Any } from "./any.ts";`
  - is/mod.ts re-re-exports the namespace:
    `export * as is from "./is.ts";`
  - mod.ts re-re-re-exports the namespace:
    `export { is } from "mod.ts";`
- On jsr.io docs, symbols appear as is.isAny (and similar), and clicking them yields 404 Not Found. The symbol browser and JSDoc linking seem inconsistent with the actual export graph and/or symbol names.

Minimal Reproduction
1) is.ts
   ````
   export { isAny as Any } from "./any.ts";
   export { isString as String } from "./string.ts";
   // ... many similar lines
   ```

2) mod.ts
   ```
   export * as is from "./is.ts";
   ```

Observed Behavior
- The docs render symbols under a nested path like `is.isAny`.
- Links for those nested symbol names point to pages that do not exist, resulting in 404 Not Found.
- JSDoc for those symbols is either missing or linked incorrectly due to the double qualification.

Expected Behavior
- Re-re-exported namespace members should resolve to stable, correct doc URLs.
- Links for symbols should not 404, and symbol names should reflect the actual exported names (e.g., `is.Any` rather than `is.isAny` if the public name is Any).
- The doc generator should correctly handle:
  - `export { X as Y } …` combined with
  - `export * as ns from "…"`

Notes and Hypothesis
- The doc generator may be deriving link targets from internal/original names (e.g., `isAny`) instead of public aliases (`e.g., Any`) when nested under a re-exported namespace.
- Alternatively, the namespace nesting logic might be duplicating the namespace qualifier (`is.isAny`) instead of presenting the aliased member name (`is.Any`).
- Either way, the computed slug or path for the symbol likely does not match the router's expected URL, causing 404s.

What Would Help
- Clarify which name (original vs. alias) JSR intends to display and link for re-re-exported members.
- Ensure consistent symbol canonicalization so that:
  - Displayed names match the actual exported surface (including aliases).
  - URL slugs/anchors match those displayed names.
  - Re-re-exported namespaces don't introduce duplicated qualifiers.

Environment
- Package: @core/unknownutil
- Version: 4.4.0-pre.0
- Affected docs: the jsr.io generated documentation pages for the package

Request
- Please confirm whether this is a known issue in the doc/link generation pipeline.
- If helpful, I can provide a minimal standalone repo replicating the re-re-export pattern and its resulting broken links.

crowlKats:
Closes #1175.

### What's actually broken

The issue reports two things. Checking the reported package today, only one of them is still real:

**Already fixed — the `is.isAny` names and their 404s.** On `@core/unknownutil@4.4.0-pre.0` the namespace members now render under their aliased names (`is.Any`, not `is.isAny`) and resolve. I walked every symbol link on that page: **83 links, 82 return 200.**

**Still broken — the one remaining link, and it's a different cause.** The 83rd is:

```
https://github.com/jsr-core/unknownutil/blob/HEAD/@core/unknownutil/doc/is/unknown/~/isUnknown
```

which comes from this JSDoc in `is/any.ts`:

```ts
/** Use {@linkcode [is/unknown].isUnknown|isUnknown} to assume that a value is `unknown`. */
```

deno_doc resolves `{@link [module].symbol}` into a registry URL — `/@core/unknownutil/doc/is/unknown/~/isUnknown` — before the markdown renderer runs. Then `get_url_rewriter` sees a root-relative URL and sends it to the repository, because of the rule added in #768 to make README links like `/LICENSE` resolve against the repo. A working link becomes a 404.

So it's a jsr-side bug, not a deno_doc one, and it has nothing to do with the re-re-export pattern in the title — any `{@link [module].symbol}` in a package with a linked repository hits it.

### Fix

Leave `/@scope/...` paths alone. That prefix is unambiguously a registry URL on jsr, so it's right whether deno_doc emitted it or the author wrote it, and it covers cross-package links (`/@std/assert@1.0.2/doc/~/assertEquals`) too. Repository-relative links keep their #768 behaviour.

### Verified end-to-end

Published a package reproducing the pattern (aliased namespace + `{@linkcode [is/unknown].isUnknown|isUnknown}`) to a local instance **with a GitHub repository linked**, which is the case that breaks, and rendered the same page on each side:

| | link emitted |
| --- | --- |
| before | `https://github.com/sweep-core/link175/blob/HEAD/@sweep/link175@1.0.0/doc/is/unknown/~/isUnknown` → 404 |
| after | `/@sweep/link175@1.0.0/doc/is/unknown/~/isUnknown` → correct page |

Following it lands where it should:

A regression test covers both the linked-repository and no-repository cases, and asserts `/LICENSE` is still rewritten.

### One caveat

The rule keys off the `/@` prefix, so a README link to a genuine repository path that starts with `@` (say a top-level `@types/` directory) would no longer reach the repo. That seems clearly the better trade: `/@scope/name` reads as a registry link to almost every author, and cross-package doc links are far more common than repo paths beginning with `@`.

### DIFF
--- a/api/src/docs.rs
+++ b/api/src/docs.rs
@@ -619,6 +619,15 @@ fn get_url_rewriter(
       return url.to_string();
     }
 
+    // So are registry URLs. `{@link [module].symbol}` in a JSDoc comment is
+    // resolved by deno_doc into `/@scope/package/doc/...` before the rewriter
+    // sees it, and rewriting that against the repository turns a working link
+    // into a 404 (jsr-io/jsr#1175). A `/@scope/...` path is unambiguously a
+    // registry URL, so leave it be whether deno_doc or the author wrote it.
+    if url.starts_with("/@") {
+      return url.to_string();
+    }
+
     let base = if let Some(github_repository) = &github_repository {
       if url.rsplit_once('.').is_some_and(|(_path, extension)| {
         matches!(
@@ -1965,5 +1974,42 @@ mod tests {
     );
     // Protocol-relative URLs are left untouched.
     assert_eq!(rewriter(None, "//example.com/x"), "//example.com/x");
+  }
+
+  /// Regression for #1175: `{@link [module].symbol}` in a JSDoc comment is
+  /// resolved by deno_doc into a registry URL before the rewriter sees it.
+  /// Those must not be sent to the repository, which is what turned them into
+  /// 404s.
+  #[test]
+  fn url_rewriter_leaves_registry_links_alone() {
+    for rewriter in [
+      get_url_rewriter(
+        String::from("/@foo/bar/1.2.3"),
+        Some(GithubRepository {
+          id: 0,
+          owner: "foo".to_string(),
+          name: "bar".to_string(),
+          updated_at: Default::default(),
+          created_at: Default::default(),
+        }),
+        true,
+      ),
+      // and with no repository linked, where the package base is the fallback
+      get_url_rewriter(String::from("/@foo/bar/1.2.3"), None, true),
+    ] {
+      // a link into this package's own docs
+      assert_eq!(
+        rewriter(None, "/@foo/bar/doc/sub/~/someSymbol"),
+        "/@foo/bar/doc/sub/~/someSymbol"
+      );
+      // and into another package's
+      assert_eq!(
+        rewriter(None, "/@std/assert@1.0.2/doc/~/assertEquals"),
+        "/@std/assert@1.0.2/doc/~/assertEquals"
+      );
+
+      // a genuine repository path is still rewritten
+      assert_ne!(rewriter(None, "/LICENSE"), "/LICENSE");
+    }
   }
 }


crowlbot (approved):
Root cause tracing is solid, and placing the `/@` bypass right before the existing root-relative (`#768`) rule is the correct spot — it short-circuits before that rule would otherwise catch the same path and send it to the repo. The regression test covers exactly the right matrix (linked-repo and no-repo, same-package and cross-package registry links, plus confirming `/LICENSE` still routes to the repo so #768 doesn't regress). The `/@types/`-style caveat is called out and the trade-off reasoning holds — cross-package doc links are far more common than a repo path starting with `@`. LGTM.

PR #1533, opened 2026-08-26
fix(api): resolve fast-check remapped specifiers in generated npm d.ts

### LINKED ISSUE #1018: `_dist` in npm tarball is missing entries when there are `.js` + `.d.ts` files
~~See here: https://jsr.io/@iroha/core/0.2.1/crypto \
There is `wasm.js` and `wasm` directory. `wasm.js` has `// @ts-self-types="./wasm/iroha_crypto_wasm.d.ts"`.~~ See my [comment below](https://github.com/jsr-io/jsr/issues/1018#issuecomment-2719498263). Refer to https://jsr.io/@iroha/core/0.0.13/crypto

When installing this package in an NPM project and trying to typecheck with `tsc`:

```
node_modules/.pnpm/@jsr+iroha__core@0.0.13/node_modules/@jsr/iroha__core/_dist/crypto/mod.d.ts:56:23 - error TS2307: Cannot find module './wasm/iroha_crypto_wasm.d.ts' or its corresponding type declarations.

56 import * as wasm from "./wasm/iroha_crypto_wasm.d.ts";
                         ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

node_modules/.pnpm/@jsr+iroha__core@0.0.13/node_modules/@jsr/iroha__core/_dist/crypto/util.d.ts:1:41 - error TS2307: Cannot find module './wasm/iroha_crypto_wasm.d.ts' or its corresponding type declarations.

1 import type { Bytes as WasmBytes } from "./wasm/iroha_crypto_wasm.d.ts";
                                          ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

And indeed, when I look into the `_dist` directory, I don't see the `wasm/iroha_crypto_wasm.d.ts` file:

```
node_modules/@iroha/core/_dist/crypto
├── mod.d.ts
├── mod.d.ts.map
├── util.d.ts
└── util.d.ts.map
```

For reference, here is the non-`.d.ts` layer of the tarball:

```
node_modules/@iroha/core/crypto
├── mod.js
├── mod.js.map
├── mod.ts
├── util.js
├── util.js.map
├── util.ts
└── wasm
    ├── iroha_crypto_wasm.d.ts
    ├── iroha_crypto_wasm.internal.js
    └── iroha_crypto_wasm.js
```

crowlKats:
Fixes #1018

deno_graph's fast check transform rewrites relative import specifiers in the generated .d.ts to the resolved, types-preferred dependency: an import of `./wasm/lib.js` where the JS file has an adjacent hand-written declaration file (via `@ts-self-types` or a types reference) becomes `./wasm/lib.d.ts` inside the dts program.

`SpecifierRewriter::rewrite` looked specifiers up by their source-text key in the module's dependency map. The remapped specifier (`./wasm/lib.d.ts`) is not a key there (the key is `./wasm/lib.js`), so the specifier was emitted verbatim into `_dist/**/*.d.ts` — where, relative to `_dist/`, the file doesn't exist, causing `TS2307: Cannot find module` when type checking with tsc (reproduced with `@iroha/core@0.0.13`).

This resolves unmatched relative specifiers in declaration rewrites against the original module specifier and continues through the existing rewrite pipeline (including the `.d.ts` → `.js` extension mapping so TypeScript probes for the hand-written declaration file). The generated `_dist/crypto/mod.d.ts` now emits a correct relative path pointing at the file that exists in the tarball.

The committed expectation in `type_imports.txt` had the same latent breakage (`_dist/main.d.ts` importing a nonexistent `_dist/foo.d.ts`) and is corrected by this change as well.

Verified end-to-end with tsc (`moduleResolution: nodenext`): the pre-fix layout fails with exactly the reported TS2307; the post-fix layout typechecks clean.

### DIFF
--- a/api/src/npm/specifiers.rs
+++ b/api/src/npm/specifiers.rs
@@ -22,6 +22,9 @@ pub enum RewriteKind {
 #[derive(Clone, Copy)]
 pub struct SpecifierRewriter<'a> {
   pub base_specifier: &'a ModuleSpecifier,
+  /// The specifier of the module in the module graph whose specifiers are
+  /// being rewritten. Relative specifiers are resolved against this.
+  pub module_specifier: &'a ModuleSpecifier,
   pub source_rewrites: &'a HashMap<&'a ModuleSpecifier, ModuleSpecifier>,
   pub declaration_rewrites: &'a HashMap<&'a ModuleSpecifier, ModuleSpecifier>,
   pub dependencies: &'a IndexMap<String, Dependency>,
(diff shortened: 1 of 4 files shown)


crowlbot (approved):
Root cause and fix line up well. Traced through `SpecifierRewriter::rewrite`: when the source-text lookup misses, falling back to resolving the (now deno_graph-remapped) relative specifier against `module_specifier` — the original module, not `base_specifier` (the output location) — is exactly right, since those two can differ (the declaration-target branches pass a different `base_specifier` while the module being read from is unchanged). The resolved absolute specifier then flows through the same `follow_specifier`/extension-swap pipeline as any other lookup, so the existing `.d.ts`→`.js` TS2846 workaround still applies — which the new fixture's expected output (`../wasm/lib.js`, not `.d.ts`) confirms. Good catch also fixing `type_imports.txt`'s expectation, which was pinning the same latent bug. All 6 call sites in tarball.rs got `module_specifier` wired through. CI green.

PR #1534, opened 2026-08-26
fix(api): let staff recompute a version's score meta to heal stale slow-types flags

### LINKED ISSUE #698: False negative "slow types"
Package: https://jsr.io/@whiteand/cbor/score

I see low score, because of the 'Slow type'

But, I don't have slow types. My package has "isolatedDeclarations": true. And the `jsr publish` is successful.

crowlKats:
Fixes #698

The "No slow types" score component (`allFastCheck`) is computed once at publish time and stored in `package_versions.meta` forever. Investigating the three packages reported on the issue showed the "false positives" are stale stored data, not a bug in current code:

- **@li/regexp-escape-polyfill@0.3.4** (published 2024-09): still stored as `allFastCheck: false` today. Re-running the *current* analysis on the exact published files yields `true` — its `.mjs` entrypoints are typed via `/// <reference types>` declaration files, a layout the analysis rejected until #992 (2025-02) and subsequent deno_graph upgrades. The stored meta was simply never recomputed.
- **@whiteand/cbor** (the original 2024 report): same class of problem at the time; its current version is healthy (`allFastCheck: true`, score 100), and current deno and the current server agree on the old layout too.
- **@rough/rough-elements@3.0.12** (2026-08 comment): *not* a false positive — both the server analysis and `deno publish --dry-run` on its published files report ~79 real slow-types problems (mixin super-class expressions etc.), and it was published with `--allow-slow-types` (`npx jsr publish --allow-slow-types` in its npm scripts). `deno lint --rules-tags=jsr` not flagging mixin super-class expressions is a deno CLI lint gap. Their 3.0.15 fixed the types and scores 100.

Since the analysis itself is already correct, this adds `POST /api/admin/packages/:scope/:package/:version/recompute_meta`: staff-only, re-downloads the version's module files from the modules bucket, re-runs `analyze_package`, and stores only the freshly computed score meta (audit-logged; `hasProvenance` is carried over since it's recorded post-publish and not derivable from the files; docs and the npm tarball are untouched). The admin panel's packages tab gets a "recompute" action that prompts for a version and calls the endpoint, showing the recomputed score.

After deploy, running it against `@li/regexp-escape-polyfill@0.3.4` (and any other pre-2025 version with this layout) heals the score.

Includes an integration test that publishes a #698-layout package, corrupts the stored meta to simulate the old pipeline, verifies non-staff get 403, and confirms the recompute heals `allFastCheck` while preserving provenance.

### DIFF
--- a/api/.sqlx/query-fa5333f441b4533c8f1a2a68ffd68eca5a65f0bf664f22ef844219bf8d058631.json
+++ b/api/.sqlx/query-fa5333f441b4533c8f1a2a68ffd68eca5a65f0bf664f22ef844219bf8d058631.json
@@ -1,0 +1,17 @@
+{
+  "db_name": "PostgreSQL",
+  "query": "UPDATE package_versions SET meta = $1\n      WHERE scope = $2 AND name = $3 AND version = $4",
+  "describe": {
+    "columns": [],
+    "parameters": {
+      "Left": [
+        "Jsonb",
+        "Text",
+        "Text",
+        "Text"
+      ]
+    },
+    "nullable": []
+  },
+  "hash": "fa5333f441b4533c8f1a2a68ffd68eca5a65f0bf664f22ef844219bf8d058631"
+}
--- a/api/src/api/admin.rs
+++ b/api/src/api/admin.rs
@@ -14,8 +14,13 @@ use tracing::Span;
 use tracing::field;
 use tracing::instrument;
 
+use std::collections::HashMap;
+
+use crate::analysis::PackageAnalysisData;
+use crate::analysis::analyze_package;
 use crate::db::*;
 use crate::iam::ReqIamExt;
+use crate::ids::PackagePath;
 use crate::ids::ScopeDescription;
 use crate::publish::publish_task;
 use crate::util;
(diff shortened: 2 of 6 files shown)


crowlbot (approved):
Appreciate the investigation splitting the issue's three reports into "stale stored data" vs. "actually using --allow-slow-types" — that's exactly the kind of due diligence that keeps this from becoming a blanket re-score. On the endpoint itself: confirmed `analyze_package`/`analyze_package_inner` has no bucket/storage side effects (that's confined to the separate `rebuild_npm_tarball_inner`, which this doesn't call), so the "docs and npm tarball are untouched" claim in the description holds — this really only touches the `meta` column. Provenance carryover, the staff-only gate, and the audit log are all there and covered by the test, including the FORBIDDEN check for non-staff. One very minor thing, not blocking: the config-file fallback to `/jsr.json` when none of the four candidates exist would produce a slightly wrong path in an error message for an edge case that (per the comment) can't really happen in practice, so I don't think it's worth guarding further. LGTM.

crowlbot (approved):
Backend logic is unchanged from what I reviewed (comments were trimmed, no functional diff). The new `RecomputeVersionMeta` admin-panel modal is a sensible, low-risk addition — it's gated by the same staff-only `check_admin_access` on the backend regardless of frontend state, posts with no body (matching the handler, which doesn't decode one), and the result string is rendered as plain JSX text rather than `dangerouslySetInnerHTML`, so a server error message can't inject markup. Consistent with other admin-only routes in not needing an `api.yml` entry. LGTM.

PR #1535, opened 2026-08-26
chore: upgrade deno_graph to 0.111.0 and deno_doc to 0.207.0

### LINKED ISSUE #155: Slow types: Better detection/fallback for complex constant declarations
I often use a pattern in my code where some object has a lot of keys, and I want TypeScript to automatically detect the keys and make them part of the type.

For example, [in `quantity-math-js`](https://github.com/bradenmacdonald/quantity-math-js/blob/92c34af8d8145cfaf74089e8347cf05a3abd52ae/units.ts#L4-L39):

```ts
/** SI prefixes */
export const prefixes = Object.freeze(
    {
        ...
        k: 1e+3,  // kilo
        M: 1e+6,  // Mega
        G: 1e+9,  // Giga
        ..
    } as const satisfies Record<string, number>,
);
```

I want for example to use that to define a related type using all the keys:
Screenshot 2024-03-03 at 2 13 43 PM

Here are two other examples of this pattern with even longer type definitions (more keys):

| A [second example: units.ts](https://github.com/bradenmacdonald/quantity-math-js/blob/92c34af8d8145cfaf74089e8347cf05a3abd52ae/units.ts#L65-L405) | A [third example: Icon.tsx](https://github.com/bradenmacdonald/neolace/blob/1c6ac132b6091d72d169a3faa7bee7c0bb54b7f8/frontend/components/widgets/Icon.tsx#L12-L110) |
|-|-|
| Screenshot 2024-03-03 at 2 21 41 PM | Screenshot 2024-03-03 at 2 19 14 PM |

Now, not surprisingly, `deno lint` and `deno publish` complain that these are "slow types". But given how complex the types are (see examples), I don't want to duplicate them in the code. And given how useful it is to have the keys available in my IDE with auto-completion, I don't want to simplify the types either.

------------------

I think a nice compromise solution would be to degrade these types to whatever type is specified using the `satisfies` operator, when necessary. So if I slightly re-write my declaration as follows:

```ts
export const prefixes = Object.freeze(
    {
        ...
        k: 1e+3,  // kilo
        M: 1e+6,  // Mega
        G: 1e+9,  // Giga
        ..
    } as const,
) satisfies Record<string, number>;
```

Then ideally, `deno` and `typescript` and other first-class users of the code will have the full, detailed type with all the exact keys available, but for JSR's purposes of documenting this and transpiling it for Node.js or whatever other things need to avoid slow types, it should just simplify the type of this to `Record<string, number>` which is OK to do because of the explicit`satisfies` declaration

Would that be possible?

### LINKED ISSUE #615: function in object is inferred as `any` with Node.js project
## Reproduction repository
https://jsr.io/@ryoppippi/jsr-bug-reproduction
https://github.com/ryoppippi/jsr-bug-reproduction

## description
Here is the description of the bug.

So, suppose you define object like this and upload it to `JSR`.

```ts
export function fAdd(a: number, b: number): number {
  return a + b;
}

export const f = {
  add: fAdd,
};
```

Then, you create `Node.js` or `Bun` project and import your library like this:

```ts
import * as mod from "@ryoppippi/jsr-bug-reproduction";

mod.f.add satisfies string; // OK because `mod.f.add` inferred as any.....
```

and you can see that the function is inferred as `any`...

## Type definition from `JSR`	

we can see `node_modules/@ryoppippi/jsr-bug-reproduction/_dist/mod.d.ts`, which is a type definition generated by `JSR`:
```ts
export declare function fAdd(a: number, b: number): number;
export declare const f: {
  readonly add;
};
//# sourceMappingURL=mod.d.ts.map
```
As you can see, no types are added to the `add` method.

## Related
https://github.com/jsr-core/unknownutil/issues/86

### LINKED ISSUE #886: Generated d.ts file from `npm.jsr.io` doesn't have `Symbol.iterator` declaration
Hi, I recently published a module on JSR: https://jsr.io/@melkam/circular-buffer. I tried using it with other package managers like pnpm and bun. It works fine overall, but I noticed a problem: my CircularBuffer class doesn't include the `[Symbol.iterator]()` method in its generated .d.ts file, even though it's defined in the source code.

Here’s my source code:

```ts
export class CircularBuffer<T> implements Iterable<T, undefined> {
  // ... other class methods
  [Symbol.iterator](): Iterator<T, undefined> {
    return { ... };
  }
}
```
But the generated .d.ts file looks like this:

```ts
export declare class CircularBuffer<T> implements Iterable<T, undefined> {
  // ... other class methods
  // [Symbol.iterator] is missing :(
}
```
For some reason, [Symbol.iterator] just disappears in the declaration file. To check it out, you can install my package with a package manager other than Deno and inspect `/node_modules/@melkam/circular-buffer/_dist/mod.d.ts`.

I also tried creating a sandbox example to demo this issue, but it seems JSR modules can’t be installed in StackBlitz right now.

### LINKED ISSUE #1179: Missing .d.ts output for `[Symbol.dispose]`
Running `jsr install <pkg>` generates declaration files under `node_modules/<scope>/<pkg>/_dist/mod.d.ts`.

In `main.ts`, the `<pkg>` defines:
```ts
class A {
  ...
  [Symbol.dispose]() { ... }
}
```

However, the generated file `node_modules/<scope>/<pkg>/_dist/mod.d.ts` does not include a declaration for `A[Symbol.dispose]`.

```
The initializer of a 'using' declaration must be either an object with a '[Symbol.dispose]()' method, or be 'null' or 'undefined'.
  Property '[Symbol.dispose]' is missing in type 'A' but required in type 'Disposable'. ts(2850)
```

### LINKED ISSUE #1258: bug: wrong types for constants
If I define a constant like this: https://jsr.io/@grammyjs/grammy/2.0.0-beta.2/src/convenience/constants.ts#L62

Then the type is `any[]` in the docs: https://jsr.io/@grammyjs/grammy@2.0.0-beta.2/doc/constants/~/ALL_UPDATE_TYPES

And if you inspect the npm tarball, it is just this:

```ts
export declare const ALL_UPDATE_TYPES: readonly ["chat_member", "message_reaction", "message_reaction_count"];
```

### LINKED ISSUE #1478: Publishing a package with a stable `with { type: "text" }` import fails server-side: registry analysis builds its graph with `unstable_text_imports: false`
## Summary

Publishing any package that uses a **runtime-stable** text import (`with { type: "text" }`, stabilized in Deno 2.8 via denoland/deno#34238) fails during registry-side processing with:

```
failed to build module graph: The import attribute type of "text" is unsupported.
  Specifier: file:///README.md
    at file:///cli.ts:18:27
```

`deno publish --dry-run` accepts the exact same package — the rejection only surfaces on the real authenticated publish, which makes it invisible to every local/CI preflight.

Originally filed against Deno as denoland/deno#35546. After tracing both codebases, the Deno CLI is **not** at fault: since denoland/deno#34692 the client-side publish path (graph build, diagnostics, tarball) permits stable text imports, and dry-run vs. real publish share an identical client pipeline (dry-run simply stops before upload). The error string originates in `deno_graph`'s asset-load gate, raised from **this repository's publishing-task analyzer**, which builds its own module graph with text imports disabled.

## Root cause (in this repo)

Pipeline: `POST /scopes/.../versions/...` → `process_publishing_task` (`api/src/publish.rs:188`) → `process_tarball` (`api/src/tarball.rs:102`) → `analyze_package` (`api/src/tarball.rs:340` → `api/src/analysis.rs:79`).

`analyze_package_inner` builds a fresh `deno_graph::ModuleGraph` over the uploaded tarball with:

```rust
unstable_bytes_imports: false,
unstable_text_imports: false,   // api/src/analysis.rs:167  ← rejects stable text imports
unstable_css_imports: false,
```

`deno_graph` then classifies the import as an asset load and `graph.valid()` surfaces `UnsupportedImportAttributeType` (`deno_graph src/graph.rs:3040` / `:5545`), which `api/src/tarball.rs:643` wraps into the publish error.

A second identical omission exists in the npm-tarball rebuild path (`rebuild_npm_tarball_inner`, `api/src/analysis.rs:566-609`, flags at `:602-605`).

## Reproduction

Any minimal package with `import text from "./asset.txt" with { type: "text" }`: `deno publish --dry-run` → success; real publish → the error above. Hit in production publishing `@netscript/mcp@0.0.1-beta.10` on 2026-07-17 (Deno 2.9.0); internal tracking: rickylabs/netscript#810.

## Notes

- Bytes imports remain unstable in Deno, so keeping `unstable_bytes_imports: false` is presumably still correct — this report is about **text** imports only.
- One policy consideration for maintainers: JSR's npm compatibility layer does not rewrite `with { type: "text" }` for Node consumers, so you may want to sequence npm-tarball handling (e.g. inlining) with the flag flip. A proposed minimal fix follows in a comment.

Cross-references: denoland/deno#35546 (original report, can be closed/transferred in favor of this one), denoland/deno#34238 (stabilization), denoland/deno#34692 (client-side publish fix), rickylabs/netscript#810 (production impact).

crowlKats:
Fixes #155
Fixes #615
Fixes #886
Fixes #1179
Fixes #1258
Fixes #1478

Upgrades deno_graph 0.110.2 → 0.111.0 and deno_doc 0.206.0 → 0.207.0, picking up the slow-types / d.ts generation fixes from denoland/deno_graph#667 and the docs-side type inference fix from denoland/deno_doc#852:

- **#615**: object properties referencing functions get `typeof fn` in the generated d.ts instead of an untyped (implicit `any`) property.
- **#886 / #1179**: `[Symbol.iterator]`, `[Symbol.dispose]` and other well-known symbol class members are no longer dropped from the generated d.ts.
- **#1258**: spread elements in `as const` arrays are no longer silently dropped from the emitted tuple type (d.ts side, deno_graph) and no longer render as `any[]` on doc pages (docs side, deno_doc). Negative number literals in tuples are also fixed.
- **#155**: `Object.freeze(... as const satisfies ...)` and `new`-expressions of non-generic module-level classes no longer error with missing-explicit-type.

Note these fixes apply at publish time — already-published versions keep the d.ts/docs generated when they were published.

Also: deno_graph 0.111.0 removed the `unstable_text_imports` build option (text imports are now always enabled in graph building), so the three `BuildOptions` sites that passed `false` just drop the field.

deno_doc 0.206.0 → 0.207.0 touches only type inference (`src/ts_type.rs`) — no `.hbs` template or ctx struct changes, so no frontend JSX mirroring is needed. `cargo test npm::` (tarball specs) passes unchanged; clippy clean.


crowlbot (approved):
Mechanical bump, all three `BuildOptions` sites consistently drop the removed `unstable_text_imports` field, `Cargo.lock`/`Cargo.toml` pins match, and CI's green. The issue-to-fix mapping in the description is clear and each one traces to a real upstream change. Nothing to add.

AugustinMauroy (approved):

PR #1538, opened 2026-08-27
fix: render the package page for non-latest versions instead of redirecting

### LINKED ISSUE #1537: Unable to access specific version from the "Versions" tab on jsr.io
When trying to go to a specific version on the [Versions](https://jsr.io/@serial/serial/versions) tab for example accessing [@serial/serial@0.1.14](https://jsr.io/@serial/serial@0.1.14) I always get redirected to the latest version [@serial/serial/doc/all_symbols](https://jsr.io/@serial/serial/doc/all_symbols) instead of the selected version.

I also get this error in the console:
```
Uncaught (in promise) TypeError: b.subitems is not iterable
    at fresh-island__LocalSymbolSearch-BjWx6tzW.js:6:549
(anonymous) @ fresh-island__LocalSymbolSearch-BjWx6tzW.js:6
await in (anonymous)
(anonymous) @ fresh-island__LocalSymbolSearch-BjWx6tzW.js:6
N @ hooks.module-Duq_hKdj.js:1
D @ hooks.module-Duq_hKdj.js:1
setTimeout
n @ hooks.module-Duq_hKdj.js:1
requestAnimationFrame
S @ hooks.module-Duq_hKdj.js:1
u.diffed @ hooks.module-Duq_hKdj.js:1
(anonymous) @ signals.module-C8X8lUOU.js:1
R.diffed @ fresh-island__LocalSymbolSearch-BjWx6tzW.js:1
J @ preact.module-eXpSlL-j.js:1
le @ preact.module-eXpSlL-j.js:1
J @ preact.module-eXpSlL-j.js:1
be @ preact.module-eXpSlL-j.js:1
r @ client-entry-BNwJGB_h.js:1
postTask
Y @ client-entry-BNwJGB_h.js:1
$e @ client-entry-BNwJGB_h.js:1
(anonymous) @ all_symbols:231
```

crowlKats:
Since #1473, docs are only served for the latest version of a package. The package overview route also requests docs, so visiting `/@scope/pkg@version` for any non-latest version got `docsOnlyForLatestVersion` from the API and 302'd to `/@scope/pkg/doc/all_symbols` — making version pages unreachable from the Versions tab.

Now only the actual doc routes (entrypoint/symbol/all_symbols) redirect to the latest version's docs; the package overview page fetches the version's metadata and renders the header and nav for that version, with a note that documentation is only available for the latest version.

This also fixes the `TypeError: b.subitems is not iterable` crash in the local symbol search island reported in the issue: since #1518, module docs can contain an `example` section, but the island cast every section to `namespace_section` and iterated `symbol.subitems`. Sections other than `namespace_section` are now skipped when building the search index and when filtering results.

Fixes #1537

### DIFF
--- a/frontend/routes/package/(_islands)/LocalSymbolSearch.tsx
+++ b/frontend/routes/package/(_islands)/LocalSymbolSearch.tsx
@@ -14,11 +14,7 @@ import {
 import { Highlight, type Position } from "@orama/highlight";
 import { api, path } from "../../../utils/api.ts";
 import { useMacLike } from "../../../utils/os.ts";
-import type {
-  AllSymbolsCtx,
-  AllSymbolsItemCtx,
-  SectionContentNamespaceSectionCtx,
-} from "@deno/doc/html-types";
+import type { AllSymbolsCtx, AllSymbolsItemCtx } from "@deno/doc/html-types";
 import { renderToString } from "preact-render-to-string";
 import { AllSymbols } from "../../../components/doc/AllSymbols.tsx";
 
@@ -113,11 +109,8 @@ export function LocalSymbolSearch(
 
       for (const entrypoint of searchContent.value!.entrypoints) {
         for (const kindGroup of entrypoint.module_doc.sections.sections) {
-          for (
-            const symbol
-              of (kindGroup.content as SectionContentNamespaceSectionCtx)
-                .content
-          ) {
+          if (kindGroup.content.kind !== "namespace_section") continue;
+          for (const symbol of kindGroup.content.content) {
             searchItems.push({
               name: symbol.name,
               symbolName: symbol.name,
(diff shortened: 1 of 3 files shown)


crowlbot (approved):
Both fixes check out.

**Redirect scoping**: narrowing the `docsOnlyForLatestVersion` redirect to only the doc-specific routes (`"entrypoint" in docs || "all_symbols" in docs`) is the right discriminator — the `Symbol(entrypoint, symbol)` case still carries `entrypoint` so it still redirects correctly, and only the bare overview request (neither key present) takes the new no-redirect path. The new branch's error handling mirrors the existing `scopeNotFound`/`packageNotFound` pattern already used just above it, plus the new `packageVersionNotFound` case. And the frontend's three-way `data.docs ? … : data.selectedVersion ? … : …` correctly distinguishes "version exists but has no docs" from "package never published" for the two different messages.

**LocalSymbolSearch crash**: discriminating on `content.kind !== "namespace_section"` before touching `.content`/`.subitems` is the correct fix for the union type — applied consistently in both the index-building loop and the results-filtering map, and skipping non-namespace sections from the *symbol* search index is the right behavior (an `example` section isn't a symbol listing anyway). CI green.

PR #1541, opened 2026-08-27
fix: make the lb's bucket cache purgeable on publish

### LINKED ISSUE #1455: @epdoc/datetime — general meta.json stale after publish, versions 4.1.8 and 4.1.9 invisible to resolver
## Description
After publishing @epdoc/datetime@4.1.8 (and subsequently 4.1.9), the general meta.json at https://jsr.io/@epdoc/datetime/meta.json does not include these versions. The version-specific _meta.json files exist and return 200:
- https://jsr.io/@epdoc/datetime/4.1.8_meta.json ✅ returns 200
- https://jsr.io/@epdoc/datetime/4.1.9_meta.json ✅ returns 200
But meta.json has been serving stale content for 22+ hours:
cf-cache-status: HIT
age: ~80000s
cache-control: public, max-age=14400, s-maxage=86400, stale-while-revalidate=86400
Cache-busting with ?timestamp proves the origin has the correct data (latest: 4.1.9, includes both 4.1.8 and 4.1.9). The issue is the CDN edge cache not being purged.
Cache control mismatch
Source code at api/src/s3.rs defines:
CACHE_CONTROL_MANIFEST = "public, max-age=60, s-maxage=60, stale-while-revalidate=86400"
But the actual response uses max-age=14400, s-maxage=86400. This suggests either:
1. The production deployment has different cache-control values
2. Or the Cloudflare purge API call failed silently during these publishes
Impact
deno update --latest fails with:
error: Could not find version of '@epdoc/datetime' that matches specified version constraint '^4.1.9'
Since Deno's resolver reads from meta.json, new versions are invisible to all consumers until the CDN cache refreshes.

crowlKats:
Fixes #1455.

### The bug

The lb keys R2-backed responses under a namespace of their own, because `/@scope/...` is served as *either* a module file (bucket) or an HTML page (frontend) depending on request headers, and the two must not cross-serve. That namespace was a synthetic host: `https://bucket-cache.jsr.internal/<host><path>`.

Cloudflare's purge-by-URL API only accepts URLs inside the zone, so every publish-time purge of `https://jsr.io/@scope/pkg/meta.json` missed the key the manifest was actually cached under. The result is exactly what was reported: R2 had the regenerated `meta.json`, a cache-busting query string returned the correct `latest`, but the edge kept serving the previous copy until its TTL lapsed. `_meta.json` and the version page were live the whole time, while Deno's resolver — which reads `meta.json` — could not see the new version, and further publishes did not dislodge it. The npm version manifest on `npm.jsr.io` had the same gap.

(#1456 capped how *long* that lasts by dropping SWR and shortening `s-maxage`; this makes the purge itself land, so a publish is visible immediately rather than after the TTL.)

### The fix

Namespace bucket entries under a reserved path on the **public** origin — `https://jsr.io/__bucket-cache/@scope/pkg/meta.json` — instead of a synthetic host. The host is still part of the key, so the module and npm buckets stay distinct, and the purge now has a URL inside the zone that it can target.

`route` rejects inbound requests for the reserved prefix before any backend, cache read, or cache write, so the namespace cannot be reached — or poisoned — from outside.

The API mirrors the prefix (`s3_paths::BUCKET_CACHE_PREFIX`, kept in sync with `lb/proxy.ts`) and purges both the public and the namespaced form of each mutable manifest. The public URL still matters there: a browser navigation to the same URL is served, and cached, by the frontend.

Changing the key also orphans the current entries, so deploying this clears any manifest that is stuck right now.

### Tests

- `s3_paths`: both purge helpers return the public URL *and* the namespaced key.
- `lb/proxy_test.ts`: bucket cache keys live on the public origin under the prefix, and the module vs. npm buckets stay keyed apart.
- `lb/main_test.ts`: `route` 404s the reserved namespace on all three hosts without touching a backend, while a merely similar-looking path routes normally.

`cargo clippy --all-targets` clean; `deno test` in `lb/` passes (43 tests).

### DIFF
--- a/api/src/api/package.rs
+++ b/api/src/api/package.rs
@@ -709,11 +709,11 @@ async fn update_description(
     .await?;
 
   cache_purge
-    .purge(vec![crate::s3_paths::npm_version_manifest_url(
+    .purge(crate::s3_paths::npm_version_manifest_purge_urls(
       npm_url,
       scope,
       &package.name,
-    )])
+    ))
     .await;
 
   Ok(package)
@@ -824,12 +824,12 @@ async fn upload_package_manifests(
     )
     .await?;
 
-  cache_purge
-    .purge(vec![
-      crate::s3_paths::package_metadata_url(registry_url, scope, package),
-      crate::s3_paths::npm_version_manifest_url(npm_url, scope, package),
-    ])
-    .await;
+  let mut purge_urls =
+    crate::s3_paths::package_metadata_purge_urls(registry_url, scope, package);
+  purge_urls.extend(crate::s3_paths::npm_version_manifest_purge_urls(
+    npm_url, scope, package,
+  ));
+  cache_purge.purge(purge_urls).await;
 
   Ok(())
 }
--- a/api/src/external/cloudflare.rs
+++ b/api/src/external/cloudflare.rs
@@ -140,8 +140,13 @@ pub struct CachePurge(pub Option<CachePurgeClient>);
 impl CachePurge {
   /// Purge `urls` if a client is configured. Errors are logged inside
   /// `purge_urls` and converted into `Ok(())` here, since callers want
-  /// best-effort behaviour (the manifests have `stale-while-revalidate`
-  /// as their durability net).
+  /// best-effort behaviour (the manifests' short `s-maxage` is their
+  /// durability net — see `CACHE_CONTROL_MANIFEST`).
+  ///
+  /// Note that a bucket-backed object is not cached under its public URL:
+  /// build its URLs with `s3_paths::package_metadata_purge_urls` /
+  /// `npm_version_manifest_purge_urls`, which also cover the lb's namespaced
+  /// cache key. Purging the public URL alone is a no-op for those.
   pub async fn purge(&self, urls: Vec<String>) {
     let Some(client) = &self.0 else {
       return;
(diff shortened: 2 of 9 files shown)


crowlbot (approved):
Solid root-cause tracing — moving the bucket cache key off a synthetic host onto the public origin is the right fix for the purge-by-URL constraint, and I paid close attention to the security implication of that move since it puts the cache namespace back on a routable host.

Confirmed `isBucketCachePath` is checked as the very first thing in `route()`, before any hostname branch, CORS handling, or backend/cache access — so a crafted request to `/__bucket-cache/...` on any of the three hosts can neither read nor plant an entry there, which the `main_test.ts` addition verifies nicely (mocking the bucket/frontend to throw if touched at all, rather than just asserting the response). The host staying part of the cache key (via `u.origin`) keeps the module and npm buckets apart, covered by the new per-host test. On the Rust side, `BUCKET_CACHE_PREFIX` matches the TS constant exactly and both are exercised by exact-string-match tests, and all four purge call sites were updated consistently to the pluralized helpers.

One only-worth-mentioning-once thing: the prefix has to stay manually in sync across the Rust and TS constants (no shared source), which the comments call out but nothing enforces at compile time — not asking for anything here, just flagging it as the one place a future refactor could quietly drift. CI green.

PR #1550, opened 2026-09-03
feat(tickets): let staff open a ticket addressed to a user

crowlKats:
Staff could only ever answer tickets; there was no way to start a conversation with a user. This reuses the ticket system for that rather than adding a separate messaging feature.

## What changes

- **`POST /api/admin/users/:user_id/tickets`** (staff only) opens a ticket of the new kind `staff_outreach`, owned by the target user, with the staff member's message as its first, outbound message. Takes `subject`, `message` and an optional `meta` object. Writes an audit log entry.
- The ticket starts out **`waiting_on_user`** so it stays out of the open queue until the user replies. From then on it behaves like any ticket the user opened themselves: it lists in their account, their replies are inbound and flip the status to waiting on support, the unread badge counts it, and staff notes and canned replies work on it.
- A new **`support_ticket_outreach` email** introduces the conversation ("The JSR team has opened a support ticket to get in touch with you about …") instead of announcing a reply. Subject is `[TICKET-…] <subject>`. It is sent under a fresh Message-ID so a reply from the user's inbox threads back onto the ticket via the existing inbound webhook. Users with no email on file get the web ticket alone.
- **Admin UI**: a "contact" button on each row of the admin users table opens a subject/message form and lands on the new ticket. `EditModal` gains optional `method`, `label` and `redirect` props to support a create form; existing uses are unchanged.
- Migration adds the `staff_outreach` enum value only, so it is safe to run ahead of the API deploy.

## Not included

Contacting by bare email address (the reporter-email/claim-token ticket shape) was considered and deliberately left out; this targets accounts only.

## Testing

New test `api::admin::tests::create_outreach_ticket` covers: non-staff rejected, empty message and unknown user refused, resulting ticket kind/status/subject/meta/reporter/author, the contacted user replying (inbound, status flips), the ticket appearing in their list, and another user getting a 404. Full API suite passes; sqlx offline cache regenerated with `--all-targets`.

### DIFF
--- a/api/.sqlx/query-12a9310c1545b8bf4c9656df981403e6a69b8d4af7b8d08fd90227a3e438b35b.json
+++ b/api/.sqlx/query-12a9310c1545b8bf4c9656df981403e6a69b8d4af7b8d08fd90227a3e438b35b.json
@@ -20,7 +20,8 @@
                 "scope_quota_increase",
                 "scope_claim",
                 "package_report",
-                "other"
+                "other",
+                "staff_outreach"
               ]
             }
           }
--- a/api/.sqlx/query-2d7b84a6ce0ac4b0d1b9912463155a59524620f5504355232490a66d412d190e.json
+++ b/api/.sqlx/query-2d7b84a6ce0ac4b0d1b9912463155a59524620f5504355232490a66d412d190e.json
@@ -1,0 +1,24 @@
+{
+  "db_name": "PostgreSQL",
+  "query": "INSERT INTO tickets (kind, creator, subject, meta, status)\n          VALUES ('staff_outreach', $1, $2, $3, 'waiting_on_user')\n          RETURNING id",
+  "describe": {
+    "columns": [
+      {
+        "ordinal": 0,
+        "name": "id",
+        "type_info": "Uuid"
+      }
+    ],
+    "parameters": {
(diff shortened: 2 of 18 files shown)


crowlbot (approved):
Reviewed the API, DB, email, and frontend changes.

- `create_outreach_ticket`: correctly gated on `check_admin_access`, validates subject/message/meta, 404s on unknown user before touching anything sensitive.
- `create_staff_outreach_ticket`: single transaction (insert ticket + insert message + audit log), parameterized queries throughout, ticket load happens inside the tx before commit — no partial-state window.
- Email threading: `Message-ID` is generated before the insert and stored on the row, so a reply threads back correctly; failure to enqueue is logged rather than failing the request (right call — the ticket already exists). Templates use `{{ }}` (auto-escaped), so staff-authored subject/content can't inject HTML into the outbound email.
- `EditModal`'s new `post`/`redirect` props are backward compatible (existing `patch` callers unchanged), and the redirect URL is built from the API's own response id, not client input.
- Migration only adds an enum value, safe to ship ahead of the API per the PR description.
- Test coverage looks thorough (non-staff rejection, empty message, unknown user, full round-trip including the user's inbound reply flipping status).

No blocking issues. LGTM.

PR #1551, opened 2026-09-03
perf(frontend): fetch the symbol search index lazily on first interaction

crowlKats:
Every package docs page visit fetched `docs/search_structured` and built the Orama index on mount, even when the visitor never touched the symbol search. This defers that work until the search input is first focused or typed into.

- The mount effect is replaced by a memoized `ensureLoaded()` that runs the fetch and index build at most once.
- `onFocus` and `onInput` both trigger it. Clicking, tabbing, and the ⌘/ shortcut all focus the input, so any of them starts the load.
- The input is no longer `disabled` while the index is absent, since a disabled input cannot receive focus. If the user types before the index is ready, the load's completion re-runs the search with the current input value so early keystrokes are not lost.
- The search body moved into a `runSearch(term)` helper shared by the input handler and the post-load catch-up.

For packages the API refuses to list, typing now silently does nothing instead of the input being greyed out, which matches the previous end state after the failed fetch.

### DIFF
--- a/frontend/routes/package/(_islands)/LocalSymbolSearch.tsx
+++ b/frontend/routes/package/(_islands)/LocalSymbolSearch.tsx
@@ -85,8 +85,15 @@ export function LocalSymbolSearch(
   const searchCounter = useSignal(0);
   const searchContent = useSignal<AllSymbolsCtx | null>(null);
 
-  useEffect(() => {
-    (async () => {
+  const inputRef = useRef<HTMLInputElement | null>(null);
+  const loadPromise = useRef<Promise<void> | null>(null);
+
+  // The symbol listing is only fetched and indexed once the user interacts
+  // with the search input, so plain page visits never pay for it. The first
+  // focus or input event kicks it off; subsequent ones reuse the same promise.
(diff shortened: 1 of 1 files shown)


crowlbot (approved):
Reviewed the lazy-load logic in `LocalSymbolSearch.tsx`.

- `ensureLoaded()` is memoized via the `loadPromise` ref, so repeated focus events after the first are free.
- The ⌘/ shortcut calls `.focus()` on the input, which triggers `onFocus`/`ensureLoaded`, so that path still works.
- Race handling looks correct: while loading, `onInput` calls `ensureLoaded()` (no-op, already in flight) and returns early since `db.value` is still unset — no exception, no dropped-but-stale search. The catch-up read of `inputRef.current?.value` after load happens *after* the early return on a failed fetch, so a package the API refuses to list doesn't spuriously run a search against a nonexistent index.
- Removing `disabled={!db.value}` is required for focus-triggered loading to work at all (a disabled input can't be focused) and the PR description correctly notes the failed-fetch end state (typing silently does nothing) is unchanged from before.

CI green. LGTM.

PR #1552, opened 2026-09-04
fix(frontend): stop probing the docs API for non-latest versions

crowlKats:
## Problem

Since #1473 the API only serves docs for the latest unyanked version and rejects every other pinned version with `docsOnlyForLatestVersion` (HTTP 404). The frontend still issued that docs request for every versioned doc URL (`/@scope/pkg@1.2.3/doc/...`) and only then redirected to the versionless page.

Crawlers walk huge numbers of old versioned symbol URLs, so this guaranteed-to-fail probe is now the dominant API traffic. In a 50-minute production sample:

| | count |
|---|---|
| all API requests | 22,457 |
| docs-route requests | 17,961 |
| docs-route 404s | ~7,450 |
| of which pinned non-latest version | ~6,000 |

About 70% of the 404s come from two header-rotating scraper pools and 18% from declared bots. Each hit cost the API two DB queries, an uncached 302 from the frontend, and usually a full docs render once the redirect was followed.

## Changes

- **Skip the docs request for non-latest versions.** `Package.latestVersion` is computed by the same query the docs endpoint uses to pick "latest", so when it is set and differs from the pinned version the frontend redirects (doc pages) or renders the overview without docs, exactly as it did after the API 404, but without the round trip. When `latestVersion` is null (only prereleases published) the API stays the authority and the previous fallback path is unchanged.
- **Cache the redirect at the edge** with the versioned doc pages' existing policy (`s-maxage=3600`). The target depends only on the URL. An old version only becomes latest again if every newer version is yanked, and then the versionless URL serves it anyway.
- **`robots.txt`: `Disallow: /@*/*@*/doc`** so well-behaved crawlers stop requesting versioned doc pages that can only redirect. Versionless docs stay crawlable.

## Behaviour

- Versioned doc page, version is not latest: 302 to the versionless URL, now cacheable, no docs API call.
- Versioned doc page, version is latest: unchanged (docs API is called and serves it).
- Versioned package overview, version is not latest: rendered without docs, unchanged, one fewer API call.
- Unversioned URLs: unchanged.

`deno lint` passes and the touched files type-check; there is no test coverage for the frontend data loaders.

### DIFF
--- a/frontend/static/robots.txt
+++ b/frontend/static/robots.txt
@@ -1,1 +1,6 @@
+User-agent: *
+# Versioned doc pages (/@scope/pkg@1.2.3/doc/...) only redirect to the
+# versionless docs of the latest version, so crawling them is pure waste.
+Disallow: /@*/*@*/doc
+
 Sitemap: https://jsr.io/sitemap.xml
--- a/frontend/utils/data.ts
+++ b/frontend/utils/data.ts
@@ -105,16 +105,42 @@ export async function packageDataWithDocs(
   version: string | undefined,
   docs: DocsRequest,
 ): Promise<PackageVersionDocsRedirect | DocsData | Response | null> {
-  let [data, pkgDocsResp] = await Promise.all([
-    packageData(state, scope, pkg),
-    state.api.get<PackageVersionDocs>(
-      path`/scopes/${scope}/packages/${pkg}/versions/${
-        version || "latest"
-      }/docs`,
+  let data: PackageData | null;
+  let pkgDocsResp: APIResponse<PackageVersionDocs> | null;
+  if (version) {
+    // The API only serves docs for the latest unyanked version and rejects
+    // every other pinned version with `docsOnlyForLatestVersion`. Crawlers
+    // walk huge numbers of old versioned symbol URLs, so decide locally from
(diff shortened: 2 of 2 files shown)


crowlbot (approved):
Reviewed the redirect-logic refactor and the robots.txt change.

- The version-branch short-circuit is correct: `packageData` is fetched first, and only when `latestVersion` disagrees with the pinned `version` does it skip straight to `nonLatestVersionDocs` — the docs API call happens only when `version === latestVersion` or `latestVersion` is null (prerelease-only case, where the API remains authoritative via the existing `docsOnlyForLatestVersion` fallback). That preserves correctness while cutting the guaranteed-404 round trip.
- `robots.txt`: `Disallow: /@*/*@*/doc` requires an `@` in the second path segment, so it only matches versioned doc URLs (`/@scope/pkg@1.2.3/doc`) and correctly leaves versionless doc pages (`/@scope/pkg/doc`) crawlable.
- The new `Cache-Control` on the redirect matches the existing policy used by `source.tsx`, `doc/[file].tsx`, `doc/[symbol].tsx`, etc. — consistent with the rest of the codebase.
- Behavior table in the description matches what the code does for all four cases.

CI green. Good use of production traffic data to justify the change. LGTM.

PR #1553, opened 2026-09-14
fix: update dependencies to resolve dependabot security alerts

piscisaureus:
Updates both lockfiles and bumps the crates whose fixes need manifest
changes, closing all open dependabot alerts (#2 through #66).

* Lockfile-only: `tokio`, `time`, `bytes`, `rand`, `tar`, `quinn-proto`,
  `crossbeam-channel`, `ammonia`, `tracing-subscriber`, `rsa`, and in
  `workers-rs` `tokio-postgres` / `postgres-protocol`.
* `tree-sitter-javascript` 0.21 pinned `cc ~1.0`, which held back
  `aws-lc-sys`, `rustls`, `rustls-webpki` 0.103 and `ring` 0.17. The
  grammar crates move to their current releases and
  `tree-sitter-highlight` to 0.26 (0.27 needs rustc 1.90;
  `tree-sitter-language` is pinned to 0.1.7 for the same reason).
* `reqwest` 0.11 → 0.13, dropping `rustls` 0.21 and `rustls-webpki`
  0.101. `oauth2` 4 → 5 (typestate client; its built-in reqwest client
  is disabled and the flows go through `util::oauth2_http_request`,
  which does not follow redirects) and `postmark` 0.10 → 2. rust-s3
  keeps its own reqwest 0.12 without `gzip`/`brotli`: with those
  features unified in, reqwest inflates objects stored with
  `Content-Encoding: gzip` on download, which broke every publish test.
  The `Cargo.toml` comment records that constraint.
* `jsonwebtoken` 8 → 11 with the `aws_lc_rs` backend, replacing
  `jsonwebkey` with jsonwebtoken's own JWK types. The `aud` claim stays
  unchecked as before; jsonwebtoken 9+ would otherwise reject every
  token that carries one.
* `opentelemetry*` 0.27 → 0.32 and `tracing-opentelemetry` 0.28 → 0.33:
  the batch processors no longer take a runtime and run on their own
  thread, so the OTLP exporter uses the blocking reqwest client;
  `Resource::builder()`; trace ids are read through `get_otel_context`
  since `OtelData` is no longer public.
* `x509-parser` 0.15 → 0.18 (ring 0.17) and `async-tar` 0.4 → 0.6.

The full API test suite passes locally, and the `workers-rs` wasm crate
builds and passes clippy with its updated lockfile.


piscisaureus:
@avocet-bot please review.

piscisaureus:
Review with `codex review --base main` (codex-cli 0.154.0):

> The dependency migrations and corresponding API adaptations are internally consistent. The workspace compiles cleanly, clippy passes, and the relevant non-integration tests pass; integration failures observed were caused by sandbox restrictions on local services.

No findings.

